PR #344, opened 2014-04-11
JCLOUDS-533: Amazon S3 server-side encryption

gaul:
This commit mimics the implementation of storage classes, hooking into
AWSS3PutObjectOptions.

### DIFF
--- a/apis/s3/src/main/java/org/jclouds/s3/domain/MutableObjectMetadata.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/domain/MutableObjectMetadata.java
@@ -57,6 +57,9 @@ public interface MutableObjectMetadata extends ObjectMetadata {
     */
    void setOwner(CanonicalUser owner);
 
+   /** Currently defaults to 'NONE'. */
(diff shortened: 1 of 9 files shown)


squiddle:
@andrewgaul server side encryption interacts not at all with user-provided encryption keys. The user provided encryption keys feature is only implemented in the AWS client SDKs. It could even be combined with each other.
User provided encryption keys could even be a blobstore-generic feature which works for all blobstores (which treat data as opaque).

In short it works by encrypting it locally and adding one piece of user metadata (the checksum of the key) to identify the key later.

udy:
Has this fix made into any 2.0.x? or is it planned to go into 2.1?

andrewgaul:
This patch did not merge since we did not look at the interactions with SSE-C.

deyanTG:
So, as I understand it is not possible to use aws s3 server side encryption with jcloud api? If that is not the case how one is supposed to trigger server side encryption?

gaul:
jclouds lacks support for server-side encryption.  We will gladly accept pull requests to complete this work, including building on top of this one, but someone needs to look at the interactions with SSE-C so that we do not preclude implementing it in the future.

PR #406, opened 2014-06-16
JCLOUDS-599 ability to execute function locally before initScript runs.

ashley-taylor:
modified the templateOptions to except a predicate that will execute just before the initScript runs passing the node information into it.


everett-toews:

PR #486, opened 2014-08-14
JCLOUDS-670 Option force no password when launching images on providers

wernerb:
Adds the option to override whatever password is given from the provider for the image.

See [jira](https://issues.apache.org/jira/browse/JCLOUDS-670)


everett-toews:

PR #511, opened 2014-09-06
JCLOUDS-651: Add copyBlob to portable abstraction and add provider optimized variants

andrewgaul:


everett-toews:

blobstore/src/main/java/org/jclouds/blobstore/BlobStore.java
```diff
@@ -224,6 +227,16 @@
    String putBlob(String container, Blob blob, PutOptions options);
 
    /**
+    * Copy blob from one container to another.  Some providers implement this
+    * more efficiently than corresponding getBlob and putBlob operations.
```
andrewgaul: Done.

blobstore/src/main/java/org/jclouds/blobstore/AsyncBlobStore.java
```diff
@@ -138,6 +140,12 @@
    ListenableFuture<String> putBlob(String container, Blob blob, PutOptions options);
 
    /**
+    * @see BlobStore#copyBlob(String,String,String,String)
+    */
+   ListenableFuture<String> copyBlob(String fromContainer, String fromName, String toContainer, String toName,
+         Optional<CopyOptions> options);
```
andrewgaul: Done.

andrewgaul:
I added partial support for openstack-swift and reported [JCLOUDS-872](https://issues.apache.org/jira/browse/JCLOUDS-872) for the remaining features.  Unfortunately Swift itself does not have the conditional copy options that other providers offer.  We may want to implement this with two Swift RPCs, e.g., `blobMetadata`, check lastModified/ETag in jclouds, then call `copyBlob`, to gain cross-provider support.

andrewgaul:
The current implementation only allows overriding user metadata, not system metadata like Content-Type.  Debating whether to push as-is or wait for some other provider support like [JCLOUDS-872](https://issues.apache.org/jira/browse/JCLOUDS-872) and [JCLOUDS-875](https://issues.apache.org/jira/browse/JCLOUDS-875) to land first.

andrewgaul:
I will address system metadata in subsequent commits.

PR #515, opened 2014-09-08
Throw sane error if one of security groups is null

chemikadze:
"vararg" methods for both `NovaTemplateOptions` and `TemplateOptions` does not check values for nullness, resulting in `NullPointerException: null` from `ImmutableSet.copyOf`.


everett-toews:

PR #623, opened 2014-12-04
JCLOUDS-578: replaced request headers with swift object corresponding on...

n3llino:
Modifications in order to allow to set custom headers to swift object


andrewgaul:
Closing this as it applies to the removed swift provider.  Please test against the new openstack-swift provider.

PR #624, opened 2014-12-07
JCLOUDS-792: Configure chef-client security attributes

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-792

Adds the SSL configuration options to the BootstrapConfig object. It also changes the internal behaviour of some methods that return the config DataBag or the raw json config to return instead the BootstrapConfig object, which is much more user friendly and useful.

It also deprecates methods that no longer make sense and aligns the Enterprise Chef provider with the jclouds formatting guidelines.


nacx:
Pushed to master as [6b350fb](https://git-wip-us.apache.org/repos/asf?p=jclouds.git;a=commit;h=6b350fb6b75abbe93a8077d52b50eaa93f2d1165).

PR #629, opened 2014-12-29
JCLOUDS-529: fix aws-ec2 cleanupIncidentalResources

aledsage:
- Some users get a `DependencyVioloation`, rather than `InvalidGroup.InUse`, when attempting to delete the security group. This caused `cleanupIncidentalResources` to propagate an exception.
- Fixes it by converting this to an `IllegalStateException` (in same way as is done for `InUse`)
- Adds tests (using `MockWebServer`) for happy-path and for failing to delete the security group with each of `InUse` and `DependencyViolation` responses.


providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/compute/AWSEC2ComputeServiceApiMockTest.java
```diff
@@ -234,4 +243,99 @@ public void listNodesWhereImageDoesntExist() throws Exception {
...
+      runDeleteIncidentalResourcesGivingErrForSecurityGroup("DependencyViolation");
+   }
+   
+   public void deleteIncidentalResourcesGivingInUseForSecurityGroup() throws Exception {
+      runDeleteIncidentalResourcesGivingErrForSecurityGroup("InvalidGroup.InUse");
+   }
+   
+   protected void runDeleteIncidentalResourcesGivingErrForSecurityGroup(String errCode) throws Exception {
+      // Complicated dispatcher is needed because cleanUpIncidentalResources will retry an unpredictable 
+      // number of times (because it is time-based, for 3 seconds - not configurable).
```
nacx: If the hardcoded retry policy is an issue here, and it is demonstrated that might not be convenient when dealing with eventual consistency, should it be better be refactored to read those values from properties (or make the entire predicate injectable, or whatever) so users can customize it and we can add a better test?

aledsage: @nacx agreed that making this configurable is sensible. I'd prefer that work to be done in multiple stages if possible - with this PR being merged first, as it fixes the issue of delete-VM always failing for a sub-set of AWS account holders.

nacx: I understand what you mean, but given that you've had to struggle with this test to make it work with the current implementation (because it is not a good one), I think it is reasonable to propose to fix and improve the code as soon as we find ourselves having to workaround it.

I know this might not be directly related to the change in the error parser, but if we just *fix stuff* without trying to make the related code better, we'll just be growing the codebase in a way that will make it harder to maintain and evolve in the future. In this case, there is just a test that calls code that is not test friendly. Instead of making the test complex, it is better to fix that code, as it should be pretty straightforward.

aledsage:
@nacx @andreaturli I've added support for reconfiguring the retry timeout for cleaning up incidental resources. Could you take another look please?

nacx:
Changes lgtm. Thanks @aledsage! Mind squashing and rebasing so I can cleanly merge it?

aledsage:
I've squashed that down to two commits. I've left it as two because the changes in the retry-predicate are logically separate from the cleanupIncidentalResources changes. It could have been its own pull request, but that felt unnecessary given it is being reviewed in the context of this PR.

@nacx @andreaturli will one of you merge this please, and then I'll submit PR against master with same changes?

andreaturli:
Thanks @aledsage 

Merged at https://git1-us-west.apache.org/repos/asf?p=jclouds.git;a=commit;h=ee42fb1a687d3f8de0db9ff870cb7e9ad00d53aa
and
https://git1-us-west.apache.org/repos/asf?p=jclouds.git;a=commit;h=7866612a2eca6e3ee32bc7928e692ec325ac7d30

nacx:
Thx!

PR #630, opened 2014-12-31
Fix Java 7 build issues

andrewgaul:
We should always build the filesystem blobstore.  Also configure animal-sniffer for OkHttp.

References [JCLOUDS-797](https://issues.apache.org/jira/browse/JCLOUDS-797).


nacx:
The animal sniffer config must be removed from the okhttp driver. We need to ensure it can run on android and others. Enforcing that there are only java 6 signatures in use is a must.

andrewgaul:
@nacx I do no understand your comment; OkHttp requires Java 7 Path and thus using animal-sniffer should only enforce that the module does not use features from Java 8 and onwards?

nacx:
OkHttp should be compatibla with Java 6. They also use this plugin to make sure they don't use signatures from java > 6:
https://github.com/square/okhttp/blob/master/pom.xml#L191-210

It should be OK (and we must try) to enforce java6 with animal sniffer everywhere except in the filesystem provider.

andrewgaul:
Abandoned OkHttp commit and pushed filesystem commit to master as 38ae98e409802ebb2318f58a6395b8b9c21fedd0.

nacx:
Thanks @andrewgaul!

PR #636, opened 2015-01-09
Handle listing with marker after the last key

andrewgaul:
Also handle zero maxResults.  Found with Ceph s3-tests and s3proxy.

### DIFF
--- a/blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
+++ b/blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
@@ -20,9 +20,10 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.base.Throwables.getCausalChain;
 import static com.google.common.base.Throwables.propagate;
-import static com.google.common.collect.Iterables.find;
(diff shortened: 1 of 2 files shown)


andrewgaul:
@kahing Could you review this?

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
```diff
@@ -262,7 +269,7 @@ public boolean apply(StorageMetadata o) {
          if (!contents.isEmpty()) {
             StorageMetadata lastElement = contents.last();
             contents = newTreeSet(Iterables.limit(contents, maxResults));
-            if (!contents.contains(lastElement)) {
+            if (maxResults != 0 && !contents.contains(lastElement)) {
```
kahing: can `maxResults` be negative?

kahing:
+1

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
andrewgaul: `ListOptions.maxResults` includes this check:

```java
checkArgument(maxKeys >= 0, "maxKeys must be >= 0");
```

PR #637, opened 2015-01-09
support directory blobs

kahing:
make the filesystem blob store distinguish between a/ and a. a/
must be a directory blob with no content and only metadata

on listing, only directories with md5 attribute is considered a
blob and returned

### DIFF
--- a/apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
+++ b/apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
@@ -43,6 +43,7 @@ import org.jclouds.blobstore.LocalStorageStrategy;
 import org.jclouds.blobstore.domain.Blob;
 import org.jclouds.blobstore.domain.BlobBuilder;
 import org.jclouds.blobstore.options.ListContainerOptions;
+import org.jclouds.blobstore.reference.BlobStoreConstants;
 import org.jclouds.domain.Location;
 import org.jclouds.filesystem.predicates.validators.FilesystemBlobKeyValidator;
 import org.jclouds.filesystem.predicates.validators.FilesystemContainerNameValidator;
@@ -67,6 +68,13 @@ import com.google.common.io.ByteSource;
 import com.google.common.io.Files;
 import com.google.common.primitives.Longs;
 
+/**
+ * FilesystemStorageStrategyImpl implements a blob store that stores objects
+ * on the file system. Content metadata and user attributes are stored in
+ * extended attributes if the file system supports them. Directory blobs
+ * (blobs that end with a /) cannot have content, but otherwise appear in
+ * LIST like normal blobs.
+ */
 public class FilesystemStorageStrategyImpl implements LocalStorageStrategy {
 
    private static final String XATTR_CONTENT_DISPOSITION = "user.content-disposition";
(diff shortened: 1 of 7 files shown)


apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -328,7 +376,20 @@ public void removeBlob(final String container, final String blobKey) {
       logger.debug("Deleting blob %s", fileName);
       File fileToBeDeleted = new File(fileName);
       if (!fileToBeDeleted.delete()) {
-         logger.debug("Could not delete %s", fileToBeDeleted);
+         if (fileToBeDeleted.isDirectory()) {
+            try {
+               UserDefinedFileAttributeView view = getUserDefinedFileAttributeView(fileToBeDeleted.toPath());
+               if (view != null) {
+                  for (String s : view.list()) {
+                     view.delete(s);
```
andrewgaul: Does this handle recursive deletes correctly?

andrewgaul: Actually I might misunderstand, we have to delete all the attributes before deleting a directory?

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -328,7 +376,20 @@ public void removeBlob(final String container, final String blobKey) {
       logger.debug("Deleting blob %s", fileName);
       File fileToBeDeleted = new File(fileName);
       if (!fileToBeDeleted.delete()) {
-         logger.debug("Could not delete %s", fileToBeDeleted);
+         if (fileToBeDeleted.isDirectory()) {
```
andrewgaul: Should this be inverted, e.g.,

```java
if (file.isDirectory()) {
    // delete attributes
}
if (!file.delete()) {
    // log failure
}
```

andrewgaul:
Thanks for addressing this @kahing, the jclouds directory support generally is a little wonky due to differences between filesystem-based providers like Atmos and fake path-based providers like S3.

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
kahing: No, delete() would fail if it's not empty, in that case we need to delete all the attributes, which will cause us to no longer treat it as a blob

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
kahing: the proposed change modifies the intended semantic

andrewgaul:
Spurious test failures.  Pushed to master as b7ab8b18a090f7ff03c7a25382f8558b0e6d9751.

jdaggett:
@gaul @kahing Hi! I wanted to give you a head's up that on Mac OS X, I am consistently seeing these failures on master:
```
Failed tests:
  FilesystemBlobStoreTest.testGetDirectoryBlob:488 expected [true] but found [false]
  FilesystemBlobStoreTest.testListDirectoryBlobs:505 expected [true] but found [false]
  FilesystemBlobStoreTest.testPutDirectoryBlobs:471 expected [true] but found [false]
  FilesystemStorageStrategyImplTest.testDeleteIntermediateDirectoryBlob:399 expected [true] but found [false]
  FilesystemStorageStrategyImplTest.testGetDirectoryBlob:360 expected [true] but found [false]
  FilesystemStorageStrategyImplTest.testWriteDirectoryBlob:352 expected [true] but found [false]
```
Looking into a fix...

kahing:
I know what the problem is and working on a fix.

PR #638, opened 2015-01-09
Custom Chef version in the Omnibus installer

nacx:
The Chef version to be installed can be configured when installing the Chef gem, but not when using the Omnibus installer. This change propagates the same `chef.version` property to the Omnibus installer so it takes it into account too.


nacx:
Pushed to master as [8f9acd1](https://git-wip-us.apache.org/repos/asf?p=jclouds.git;a=commit;h=8f9acd1cadf4a4e54af96441510813be2e48c16a).

PR #640, opened 2015-01-12
disable directory blob tests in OS X

kahing:
On OS X, java doesn't support xattrs, which is required by
directory blobs. Disable those tests on OS X

should fix @jdaggett 's problem with #637

### DIFF
--- a/apis/filesystem/src/main/java/org/jclouds/filesystem/predicates/validators/internal/FilesystemBlobKeyValidatorImpl.java
+++ b/apis/filesystem/src/main/java/org/jclouds/filesystem/predicates/validators/internal/FilesystemBlobKeyValidatorImpl.java
@@ -41,11 +41,6 @@ public class FilesystemBlobKeyValidatorImpl extends FilesystemBlobKeyValidator {
         if (name.startsWith(File.separator))
             throw new IllegalArgumentException(String.format(
                     "Blob key '%s' cannot start with character %s", name, File.separator));
-
-        //blobkey cannot end with / (or \ in Windows) character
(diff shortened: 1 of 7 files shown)


jdaggett:
@kahing Thanks for the fix! This is exactly how I worked around the issue earlier this morning. As soon as the build succeeds, I am +1 for this PR. cc/ @andrewgaul

andrewgaul:
Pushed to master as 5cc3b243fc878a44091b78b14d8e5cb00086a75a.

This commit addresses the test failures, although we might need a more general solution.  Unfortunately lack of extended attributes on Mac OS X is a [JDK issue](https://bugs.openjdk.java.net/browse/JDK-8030048).

jdaggett:
@kahing @andrewgaul After pulling this commit and running the tests again, I am getting two new failures :confused: :
```
Failed tests:
  FilesystemBlobIntegrationTest>BaseBlobIntegrationTest.testPutByteSource:531->BaseBlobIntegrationTest.testPut:584
Expecting:
 <{}>
to be equal to:
 <{"key1"="value1", "key2"="value2"}>
but was not.
  FilesystemBlobIntegrationTest>BaseBlobIntegrationTest.testPutInputStream:539->BaseBlobIntegrationTest.testPut:584
Expecting:
 <{}>
to be equal to:
 <{"key1"="value1", "key2"="value2"}>
but was not.
```

Any insight is appreciated, thx!

jdaggett:
We should skip those additional two tests as well. I will submit a PR shortly with the changes, thx!

PR #641, opened 2015-01-12
Update okhttp to version 2.2.0

jdaggett:
This PR updates the `okhttp` dependency to version **2.2.0**.

There are quite a number of [fixes](https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-220) in this release. For APIs that require this dependency, the new [interceptors](https://github.com/square/okhttp/wiki/Interceptors) could prove to be a very useful feature.

### DIFF
--- a/project/pom.xml
+++ b/project/pom.xml
@@ -206,7 +206,7 @@
     <maven.site.url.base>gitsite:git@github.com/jclouds/jclouds-maven-site.git</maven.site.url.base>
     <clojure.version>1.3.0</clojure.version>
     <guava.version>16.0.1</guava.version>
-    <okhttp.version>2.1.0</okhttp.version>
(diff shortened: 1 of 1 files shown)


nacx:
>the new interceptors could prove to be a very useful feature.

Indeed! Change lgtm. Thanks! (IIRC all downstream repos should inherit this value, right?)

jdaggett:
@nacx Yes, the downstream repos will inherit this value. Do we need to update the 1.8.x branch with this commit?

nacx:
Nope. 1.8.x still uses OkHttp 1.6.0 and we'd better not upgrade libraries in bug fix versions.

jdaggett:
@nacx Thanks for the clarification! Pushed to **master** e09b615

PR #642, opened 2015-01-13
make xattr work in docker volume

kahing:
when you bind a host volume into docker, java does not correctly
detect that and checks the xattr support on the root fs instead
of the host fs. The root fs often does not support xattr, so
the check would fail even if the target really does support xattr.

the fix is just try setting the xattrs anyway, and let them fail
if there really isn't xattrs support


andrewgaul:
Pushed to master as 5edd5de38c01f3a1ea962dd9f0a4bd6d060f32de.

PR #644, opened 2015-01-14
Fixes the exponential backoff for a value of 0

zack-shoylev:


nacx:
There is one checkstyle violation to fix and... Where are the tests that shows that this change works as expected? :) Add the corresponding ones to the [BackoffLimitedRetryHandlerTest](https://github.com/jclouds/jclouds/blob/master/core/src/test/java/org/jclouds/http/handlers/BackoffLimitedRetryHandlerTest.java).

zack-shoylev:
Well I really wanted to make sure there are no issues with the idea behind this PR first - I think it makes sense, but also setting this value to 0 eliminates the retry wait for this backoff.
If we are ok with allowing 0, then I can just finish this PR. The alternative would be to disallow it, but I think in that case adding a way to override the backoff handler would be needed.

nacx:
Oh, got it. In that case, the fix should be enforcing a 0 value is never used. I'd change this to validate the value with a `checkArgument`, and fail if it is 0 (and add the corresponding test).

There are other mechanisms to disable the `BackoffLimitedRetryHandler`. Apis can just override the [HttpApiModule#bindRetryHandlers](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/rest/config/HttpApiModule.java#L73) method as follows:
```java
bind(HttpErrorHandler.class).annotatedWith(ServerError.class).toInstance(HttpRetryHandler.NEVER_RETRY);
```
Or just use it by default and let users to provide a custom module when creating the context with that same binding, if they want to disable it. No need to say this same mechanism can be used to inject a custom implementation of the retry handler.

zack-shoylev:
Based on my testing and spending a lot more time with this, I think we should allow values of 0. I also discovered that there was a bug with the algorithm with small values (<10) and fixed that as well.
Tests added.

zack-shoylev:
Extra note: the new tests did fail with the old code.

zack-shoylev:
@nacx How does this PR look to you now?

jdaggett:
@nacx I just did a line by line review with @zack-shoylev on this PR and it LGTM +1

zack-shoylev:
About to squash and merge

zack-shoylev:
merged

nacx:
LGTM. Thx!

PR #645, opened 2015-01-14
Add @Nullable attribute to BlobStore methods

andrewgaul:
blobMetadata and getBlob return null when a blob does not exist and
callers should check for this.


andrewgaul:
Theoretically FindBugs consumes these attributes and can prevent NullPointerExceptions although it seems to prefer the `@CheckForNull` annotation.  The `putbBlob` methods may benefit from this as well although I am not familiar with any object store returning `null` ETags.

@kahing relevant to your interests.

kahing:
:+1:

PR #647, opened 2015-01-15
Add Rackspace CDN service type

jdaggett:
This PR adds the Early Access Rackspace CDN service `rax:cdn` to the known service types that can be returned in the Cloud Identity Service Catalog.  For additional info on the CDN service, please refer to the [docs](http://docs.rackspace.com/cdn/api/v1.0/cdn-devguide/content/Overview.html).


jdaggett:
Pushed to **master** ade72d4

PR #648, opened 2015-01-19
fix usermetadata check on OS X

kahing:
Java on OS X does not support extended attributes, which the
filesystem blobstore uses to implement user metadata. This disable
the relevant test on OS X

### DIFF
--- a/blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
+++ b/blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
@@ -555,6 +555,10 @@ public class BaseBlobIntegrationTest extends BaseBlobStoreIntegrationTest {
       testPut(payload, new ByteSourcePayload(byteSource), length, new PutOptions().multipart(true));
    }
 
+   protected void checkUserMetadata(Map<String, String> userMetadata1, Map<String, String> userMetadata2) {
+      assertThat(userMetadata1).isEqualTo(userMetadata2);
+   }
+
    private void testPut(Payload payload, Payload expectedPayload, long length, PutOptions options)
          throws IOException, InterruptedException {
       BlobStore blobStore = view.getBlobStore();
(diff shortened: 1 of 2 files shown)


kahing:
@jdaggett @andrewgaul

jdaggett:
@kahing I will pull the changes over later this morning and let you know, thanks! :+1:

blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
```diff
@@ -555,6 +555,14 @@ public void testPutMultipartInputStream() throws Exception {
       testPut(payload, new ByteSourcePayload(byteSource), length, new PutOptions().multipart(true));
    }
 
+   /* Java on OS X does not support extended attributes, which the filesystem backend
+    * uses to implement user metadata */
+   private void checkUserMetadata(Blob blob, Map<String, String> userMetadata1, Map<String, String> userMetadata2) {
+      if ("filesystem".equals(blob.getMetadata().getProviderId()) && !TestUtils.isMacOSX()) {
```
andrewgaul: Can you make this method protected and override it in `FilesystemBlobIntegrationTest`?  We follow this pattern for other attributes like system metadata.

jdaggett:
@kahing Hi! I cherry-picked this commit, built again, and have the following 2 test failures in the `BaseBlobIntegrationTest.testPut()` method:
```
Failed tests:
  FilesystemBlobIntegrationTest>BaseBlobIntegrationTest.testPutByteSource:531->BaseBlobIntegrationTest.testPut:584
Expecting:
 <{}>
to be equal to:
 <{"key1"="value1", "key2"="value2"}>
but was not.
  FilesystemBlobIntegrationTest>BaseBlobIntegrationTest.testPutInputStream:539->BaseBlobIntegrationTest.testPut:584
Expecting:
 <{}>
to be equal to:
 <{"key1"="value1", "key2"="value2"}>
but was not.
```

kahing:
@jdaggett how did you run the test? I don't have OS X but locally I verified with disabling xattrs and faking TestUtils.isMacOSX() to always return true.

jdaggett:
@kahing I just run `mvn clean install` :smile:

kahing:
@jdaggett I just did a `mvn clean install` with xattrs disabled and TestUtils.isMacOSX() return true, and everything passed.

jdaggett:
@kahing Well, that was strange... I had a git remote problem with your repo and now that it is sorted out, `master` is happy again on OS X! +1

jdaggett:
@kahing Thanks again! Can you please address the comment from @andrewgaul above? Once that looks good, we should get it merged.

andrewgaul:
Pushed to master as 3cabb196ce62418ea497ff3087690c802f8ca7e4.

PR #649, opened 2015-01-21
JCLOUDS-794: Use bogus URL for generic S3 endpoint

andrewgaul:
This prevents users from accidentally connecting to AWS.


jdaggett:
@andrewgaul Oh yeah, wise decision to change this endpoint! +1

PR #651, opened 2015-01-22
More robust line-ending detection for Bundles

zack-shoylev:


zack-shoylev:
Seems broken on windows without that fix.

core/src/main/java/org/jclouds/osgi/Bundles.java
```diff
@@ -131,6 +131,6 @@ public T apply(Class<?> in) {
    }
 
    private static Iterable<String> splitOrEmptyAndClose(InputStream in) throws IOException {
-      return Splitter.on('\n').omitEmptyStrings().split(toStringAndClose(in));
+      return Splitter.onPattern("\r?\n").omitEmptyStrings().split(toStringAndClose(in));
```
nacx: Could you use `Splitter.on(System.lineSeparator())` instead?

zack-shoylev: Yep, but this should be better/safer. The reason for this is that the jclouds repo does not use system line endings - so it is remotely possible you could get a bug expecting a system line ending and getting a different one.

zack-shoylev: In fact, "remotely possible" - I have seen that case with jclouds before. The repo specifically has to use core.autocrlf = false on windows or tests fail, and the code usually splits new lines on \n. Which is ok - just does not seem to be enough in this specific case.

nacx: Fair enough :) lgtm!

zack-shoylev:
merged

PR #653, opened 2015-01-27
JCLOUDS-813: fix list private images

andreaturli:


andreaturli:
@nacx does it look good?

andreaturli:
@nacx Some users get a NPE when `listImages` is called.
This is happening [here](https://github.com/andreaturli/jclouds/blob/fix/softlayer-list-private-images/providers/softlayer/src/main/java/org/jclouds/softlayer/domain/OperatingSystem.java#L107). Debugging a little bit, I found out the reason: for some private images of an account [here](https://github.com/andreaturli/jclouds/blob/fix/softlayer-list-private-images/providers/softlayer/src/main/java/org/jclouds/softlayer/compute/strategy/SoftLayerComputeServiceAdapter.java#L290-L296) jclouds can't find the globalIdentifier of an image. 
I think we should simply ignore those kind of images as they are broken (not well configured by the user in SL)

This patch is an attempt to protect jclouds by these manual mistakes.

PR #654, opened 2015-01-27
Fix Softlayer template options

grkvlt:
Changed the builder methods to return `SoftLayerTemplateOptions` to allow chaining. Always set the `diskType` when creating a VM, even if no block device configured, to allow use of `LOCAL` disk.


grkvlt:
@andreaturli Can you have a look at this, please?

andreaturli:
thanks @grkvlt! lgtm +1

merging now

andreaturli:
Merged as https://git1-us-west.apache.org/repos/asf?p=jclouds.git;a=commit;h=9e1d1cbfa9e2d77fe8b7e8feeae0bb4222654692

@grkvlt thanks!

grkvlt:
@andreaturli will this become part of jclouds 1.9.0 as well as 2.0.0 then?

andreaturli:
correct! it should be already available in 2.0.0-SNAPSHOT available in maven central, I think.

nacx:
That's right. When in doubt, you can check the builds at: https://jclouds.ci.cloudbees.com/
Those run after every commit and publish the snapshots to the snapshots repo.

PR #657, opened 2015-01-29
Remove cloudservers from allcompute/pom.xml

andreaturli:


andreaturli:
@jdaggett can you please have a look?

PR #659, opened 2015-01-29
Fixes a missing image. Test now tries to match image by name.

zack-shoylev:


zack-shoylev:
This only affects a live test.

zack-shoylev:
merged

providers/rackspace-cloudservers-us/src/test/java/org/jclouds/rackspace/cloudservers/us/compute/CloudServersUSTemplateBuilderLiveTest.java
```diff
@@ -39,11 +39,11 @@ public CloudServersUSTemplateBuilderLiveTest() {
 
    @Test
    public void testTemplateBuilder() {
-      Template defaultTemplate = this.view.getComputeService().templateBuilder().build();
+      Template defaultTemplate = this.view.getComputeService().templateBuilder().imageNameMatches(".*Ubuntu 14\\.10.*").build();
```
nacx: This is not correct. The test must check the **default** template. That is, what happens when users **don't provide any** criteria in the template builder. This line **must** be reverted to the original one.

The right fix is to configure which is the default template, and that is done in the ProviderMetadata, by setting the default template property with the right criteria. You can take a look at the [CloudSigma2ApiMetadata](https://github.com/jclouds/jclouds-labs/blob/master/cloudsigma2/src/main/java/org/jclouds/cloudsigma2/CloudSigma2ApiMetadata.java#L62) for an example.

nacx:
Thanks for looking at this @zack-shoylev! The fix, however, needs to be reverted and make the ProviderMetadata configure the default template. can you have a look at it?

providers/rackspace-cloudservers-us/src/test/java/org/jclouds/rackspace/cloudservers/us/compute/CloudServersUSTemplateBuilderLiveTest.java
zack-shoylev: Ah this actually makes sense. Thanks for the info @nacx !

zack-shoylev: I won't revert as some of the fixes (such as the OS version) I want to keep, but I will adjust it to make sure it works as intended. Thanks!

nacx: Absolutely :) The important thing is to make sure there are no args to the template builder in this test. Thanks!

zack-shoylev:
Fixes in https://github.com/jclouds/jclouds/pull/661 as referenced.

PR #660, opened 2015-01-30
Fixes some missing cloudservers locations iso codes

zack-shoylev:
"US-IL", "US-TX", "US-VA", "AU-NSW", "HK"


zack-shoylev:
merged

nacx:
Thanks!

zack-shoylev:
There are a few more remaining fixes to do with that provider so stay tuned :)

PR #666, opened 2015-02-04
Fix location assignment in SoftLayer VirtualGuestToNodeMetadata

nacx:
According to the [DatacenterToLocation](https://github.com/jclouds/jclouds/blob/master/providers/softlayer/src/main/java/org/jclouds/softlayer/compute/functions/DatacenterToLocation.java#L52) function, the id of the location is the name of the datacenter. This commit fixes the `VirtualGuestToNodeMetadata` so the location is properly populated.

/cc @andreaturli


andreaturli:
lgtm! Thanks @nacx, merging now!

andreaturli:
Merged as

https://git1-us-west.apache.org/repos/asf?p=jclouds.git;a=commit;h=fa8b9e2332e3a8267d201d61b851f9d91d9e2a27

PR #667, opened 2015-02-04
fix jclouds live tests failing when a keypair is missing

zack-shoylev:
If the keypair is missing, create one.
In some configurations this can be common (cloudbees, windows)

### DIFF
--- a/compute/src/test/java/org/jclouds/compute/ComputeTestUtils.java
+++ b/compute/src/test/java/org/jclouds/compute/ComputeTestUtils.java
@@ -30,6 +30,7 @@ import java.util.concurrent.TimeoutException;
 
 import org.jclouds.compute.domain.NodeMetadata;
 import org.jclouds.rest.HttpClient;
+import org.jclouds.ssh.SshKeys;
 
 import com.google.common.base.Charsets;
 import com.google.common.base.Throwables;
@@ -48,7 +49,12 @@ public class ComputeTestUtils {
       } catch (NullPointerException e) {
          secretKeyFile = System.getProperty("user.home") + "/.ssh/id_rsa";
       }
-      checkSecretKeyFile(secretKeyFile);
+      try {
+         checkSecretKeyFile(secretKeyFile);
+      }
+      catch (IllegalStateException e) {
+         return SshKeys.generate();
+      }
       try {
          String secret = Files.toString(new File(secretKeyFile), Charsets.UTF_8);
          assert secret.startsWith("-----BEGIN RSA PRIVATE KEY-----") : "invalid key:\n" + secret;


zack-shoylev:
The error without this fix:
https://jclouds.ci.cloudbees.com/view/live%20tests/job/computeservice-rackspace/4/console

compute/src/test/java/org/jclouds/compute/ComputeTestUtils.java
```diff
@@ -62,7 +64,19 @@
    public static void checkSecretKeyFile(String secretKeyFile) {
       checkNotNull(emptyToNull(secretKeyFile), "System property: [test.ssh.keyfile] set to an empty string");
       if (!new File(secretKeyFile).exists()) {
-         throw new IllegalStateException("secretKeyFile not found at: " + secretKeyFile);
+         // Create file if missing
+         try {
+            JSch jsch = new JSch();
+            KeyPair kPair = KeyPair.genKeyPair(jsch, KeyPair.RSA);
```
zack-shoylev: The default key size is 1k. I will test with 4k and update if it works.

compute/src/test/java/org/jclouds/compute/ComputeTestUtils.java
```diff
@@ -62,7 +64,19 @@
...
+            kPair.writePrivateKey(secretKeyFile);
+            kPair.writePublicKey(secretKeyFile + ".pub", "jclouds generated key pair");
+            System.out.println("jclouds generated finger print: " + kPair.getFingerPrint());
```
zack-shoylev: This println is probably not necessary?

project/pom.xml
```diff
@@ -365,6 +365,12 @@
       <artifactId>assertj-guava</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.jcraft</groupId>
+      <artifactId>jsch</artifactId>
+      <version>0.1.51</version>
+      <scope>test</scope>
+    </dependency>
```
zack-shoylev: Note: this adds a test dependency in the jclouds project pom (the one that is used as a parent project). The test dependency might be used in multiple providers/apis live tests...

compute/src/test/java/org/jclouds/compute/ComputeTestUtils.java
```diff
@@ -62,7 +64,18 @@
...
       checkNotNull(emptyToNull(secretKeyFile), "System property: [test.ssh.keyfile] set to an empty string");
       if (!new File(secretKeyFile).exists()) {
-         throw new IllegalStateException("secretKeyFile not found at: " + secretKeyFile);
+         // Create file if missing
+         try {
+            JSch jsch = new JSch();
+            KeyPair kPair = KeyPair.genKeyPair(jsch, KeyPair.RSA, 4096);
+            kPair.writePrivateKey(secretKeyFile);
+            kPair.writePublicKey(secretKeyFile + ".pub", "jclouds generated key pair");
+            kPair.dispose();
```
nacx: You should be able to do the same without adding a new dependency:
```java
Map<String, String> keypair = SshKeys.generate();
Files.write(keypair.get("private"), new File(secretkeyFile), Charsets.UTF_8);
Files.write(keypair.get("public"), new File(secretkeyFile + ".pub"), Charsets.UTF_8);
```

But given that this method just assigns the `Map<String, String>` variable, parhaps you don't even need to persist it to disk. Could you test if it works if you just return the result of the `SshKeys.generate()` call? If not, we can persist the generated keys, but it would be better if we can avoid doing that.

zack-shoylev: I notice we generate a 2048 key... Any issues upping that to 4k?
Also, yes, great point - it's possible to pass the keypair directly, but I am not sure the files are used after during testing or not. Still checking.

zack-shoylev:
Passed live tests using the on-demand generated keypair, so this is much better now! Thanks @nacx

zack-shoylev:
Hmm, I will have to rerun tests... messed up my testing environment

zack-shoylev:
Alright, all looks good with the live tests. About to squash/merge.

zack-shoylev:
merged

PR #669, opened 2015-02-06
Make testGetIfUnmodifiedSince a lot more robust (but slower)

zack-shoylev:
This fixes possible eventual consitency bugs


zack-shoylev:
merged

PR #671, opened 2015-02-06
Makes the test testSignGetUrlWithTime more robust

zack-shoylev:


zack-shoylev:
merged

blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobSignerLiveTest.java
```diff
@@ -87,8 +89,9 @@ public void testSignGetUrlWithTime() throws InterruptedException, IOException {
          view.getBlobStore().putBlob(container, blob);
          assertConsistencyAwareContainerSize(container, 1);
          HttpRequest request = view.getSigner().signGetBlob(container, name, timeout);
-         
+
          assertEquals(request.getFilters().size(), 0);
+         Uninterruptibles.sleepUninterruptibly(timeout / 2, TimeUnit.SECONDS);
```
zack-shoylev: Because of eventual consistency (I think) the blob is sometimes not available right away and fails the first time the request is used.

PR #675, opened 2015-02-10
JCLOUDS-258: Support MPU for generic S3

andrewgaul:
Tested against AWS-S3 and DreamObjects.  This commit only moves and
renames code although some classes are duplicated for deprecation
purposes.


andrewgaul:
Ideally I would like to rename instead of duplicating these classes.  While they are part of the public API, they are not likely used by external callers:

```
apis/s3/src/main/java/org/jclouds/s3/binders/BindObjectMetadataToRequest.java
apis/s3/src/main/java/org/jclouds/s3/binders/BindPartIdsAndETagsToRequest.java
apis/s3/src/main/java/org/jclouds/s3/blobstore/strategy/AsyncMultipartUploadStrategy.java
apis/s3/src/main/java/org/jclouds/s3/blobstore/strategy/MultipartUpload.java
apis/s3/src/main/java/org/jclouds/s3/blobstore/strategy/MultipartUploadStrategy.java
apis/s3/src/main/java/org/jclouds/s3/functions/ETagFromHttpResponseViaRegex.java
apis/s3/src/main/java/org/jclouds/s3/functions/ObjectMetadataKey.java
apis/s3/src/main/java/org/jclouds/s3/functions/UploadIdFromHttpResponseViaRegex.java
```

Any opinions on skipping the deprecation process to preserve history?

zack-shoylev:
It is safer to deprecate but in this particular situation just renaming should be fine?

andrewgaul:
Going to rename to preserve history.

PR #676, opened 2015-02-10
Fixes swift - create container with metadata

zack-shoylev:


zack-shoylev:
This fixes a blobstore live test for swift

zack-shoylev:
merged

PR #677, opened 2015-02-10
Fix InitScriptConfigurationForTasks.initScriptPattern on Windows hosts

rdowner:
This method has incorrect results when run on Windows. It is expected to
generate folder names compatible with Unix-like target machines, but by
using File.getParent it will actually be influenced by the host machine
OS type. This results in baseDir being set to a style that will not work
on Unix-like targets.

### DIFF
--- a/compute/src/main/java/org/jclouds/compute/callables/InitScriptConfigurationForTasks.java
+++ b/compute/src/main/java/org/jclouds/compute/callables/InitScriptConfigurationForTasks.java
@@ -16,9 +16,9 @@
  */
 package org.jclouds.compute.callables;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import java.io.File;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.inject.Named;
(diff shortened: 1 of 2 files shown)


compute/src/main/java/org/jclouds/compute/callables/InitScriptConfigurationForTasks.java
```diff
@@ -47,7 +47,20 @@ protected InitScriptConfigurationForTasks() {
    public InitScriptConfigurationForTasks initScriptPattern(
             @Named(PROPERTY_INIT_SCRIPT_PATTERN) String initScriptPattern) {
       this.initScriptPattern = checkNotNull(initScriptPattern, "initScriptPattern ex. /tmp/init-%s");
-      this.basedir = new File(initScriptPattern).getParent();
+      checkArgument(this.initScriptPattern.startsWith("/"), "initScriptPattern must be a UNIX-style path starting at the root (/)");
+
+      int lastSlash = initScriptPattern.lastIndexOf('/');
+      if (lastSlash == -1) {
+         // no slash? impossible if the checkArgument above had worked properly!
```
nacx: Then remove this statement, and leave just the comment?

compute/src/test/java/org/jclouds/compute/callables/InitScriptConfigurationForTasksTest.java
```diff
@@ -76,4 +73,19 @@ public void testIncrementingTimeSupplier() throws InterruptedException {
...
+      assertEquals(config.getInitScriptPattern(), "/var/tmp/jclouds-%s");
+   }
+
+   @Test
+   public void testInitScriptPatternAtRoot() throws Exception {
+      InitScriptConfigurationForTasks config = InitScriptConfigurationForTasks.create();
+      config.initScriptPattern("/jclouds-%s");
+      assertEquals(config.getBasedir(), "/");
+      assertEquals(config.getInitScriptPattern(), "/jclouds-%s");
+   }
```
nacx: I'd also add a test to verify we now require the initscript to start with a slash. Let's add a test that verifies that the IAE is thrown.

nacx:
Just two minors. Apart from that lgtm. Thanks @rdowner!

andreaturli:
@rdowner can you please address the 2 minor comments from @nacx so that we can merge it?
Thanks!

nacx:
Rebased to the latest master version, addressed the minors and merged to master as [199e17ab](http://git-wip-us.apache.org/repos/asf/jclouds/commit/199e17ab).

PR #497, opened 2014-08-27
Implemented compute attach interfaces extension.

RoadRoller:
Attach interfaces API extension let to connect/disconnect running VMs to network ports. See http://developer.openstack.org/api-ref-compute-v2-ext.html#ext-os-interface. Sorry I'm have no time to cover it with unit tests, but I'm check API methods in sandbox project.

PR #679, opened 2015-02-11
Nova Attach Interface Extension API

jdaggett:
This PR integrates #497 that implements the Nova compute attach interfaces extension.

/cc @zack-shoylev


jdaggett:
Pushed to **master** a356ca9 and 59fc232

apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/extensions/AttachInterfaceApiExpectTest.java
```diff
@@ -0,0 +1,89 @@
...
+            .payload(payloadFromResource("/attach_interface_details.json")).build();
+
+      NovaApi novaApi = requestsSendResponses(keystoneAuthWithUsernameAndPasswordAndTenantName,
+            responseWithKeystoneAccess, extensionsOfNovaRequest, extensionsOfNovaResponse, list, listResponse);
+
+      InterfaceAttachment interfaceAttachment = novaApi.getAttachInterfaceApi("az-1.region-a.geo-1").get()
+            .get("instance-1", "ce531f90-199f-48c0-816c-13e38010b442");
+
+      assertEquals(interfaceAttachment, testInterfaceAttachment());
+   }
```
nacx: Add also tests for the get and list fallbacks, a test for the create method, and the tests for the delete method and its fallback.

nacx:
I'm a bit late to the party :) Just a comment about the missing tests. Also there should be a live test for each new API method, right?

PR #682, opened 2015-02-13
JCLOUDS-820: Support multi-delete for generic S3

andrewgaul:
Tested against AWS and DreamObjects.  This commit only moves and
renames code.


andrewgaul:
As with generic MPU, this theoretically changes the public API of some binders and such but practically should not affect callers.

shrinandj:
Awesome!!

PR #683, opened 2015-02-13
Fixes blob-related tests that fail occasionally

zack-shoylev:


blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
```diff
@@ -432,7 +433,13 @@ public void deleteObject(String name) throws InterruptedException {
       String container = getContainerName();
       try {
          addBlobToContainer(container, name, name, MediaType.TEXT_PLAIN);
+         if (view.getConsistencyModel() == ConsistencyModel.EVENTUAL) {
```
jdaggett: It would be great to replace these kinds of comparisons with `Assert.isEqualTo(Object expected)` available to us in assertj. Just an idea, WDYT?

zack-shoylev: Well but this just applies a wait only the case of eventual consistency providers. How will the assert work here?

zack-shoylev:
squash/rebase

zack-shoylev:
needs rebase

zack-shoylev:
merged

PR #691, opened 2015-02-19
JCLOUDS-732: portable object ACLs

andrewgaul:


zack-shoylev:
Swift supports ACLs for containers, right? Maybe add some TODOs - thoughts?

andrewgaul:
@zack-shoylev I already added portable container ACL support, including Swift, in #673.

zack-shoylev:
Ok. This makes sense. Thanks @andrewgaul !

PR #692, opened 2015-02-20
JCLOUDS-832: Don't expose the password credentials in the toString method

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-823


nacx:
merged to master as [7c9d6f76](https://git1-us-west.apache.org/repos/asf?p=jclouds.git;a=commit;h=7c9d6f76).

nacx:
There is a typo in the commit message that references issue 832 instead of 823. I've noticed just after merging, so that can't amended.

PR #693, opened 2015-02-21
fix non-recursive list with empty prefix

kahing:


andrewgaul:
Reworded commit message and pushed to master as be7b9f4cc4e42f1f2e54b7364289839f1b5c64c7.

PR #699, opened 2015-03-03
Adds autovalue/gson builders tests, for sanity.

zack-shoylev:
NOTE: this also adds the 1.1 auto snapshot. This is required for the tests to work. While this is snapshot code, I am fairly certain that generated code (and thus expected functionality) is not expected to change.

### DIFF
--- a/core/src/test/java/org/jclouds/json/JsonTest.java
+++ b/core/src/test/java/org/jclouds/json/JsonTest.java
@@ -98,7 +98,7 @@ public class JsonTest {
       assertEquals(obj2, obj);
       assertEquals(json.toJson(obj2), json.toJson(obj));
    }
-   
+
    static class ExcludeStringValue implements DefaultExclusionStrategy {
       public boolean shouldSkipClass(Class<?> clazz) {
         return false;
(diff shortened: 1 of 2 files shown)


nacx:
-1

While it only affects the tests, we are opening the door to anyone to use it by mistake, and this must **not** be used until there is a fixed release.

Also, we **only** release the source code, and the tarballs we release must produce the same articafts in each build, at any time. We can't guarantee that by adding a SNAPSHOT dependency. It could also break the tests of an already published release!

I'm -1 to have this merged to master but also understand the value of having these tests to allow us to move forward faster once it is released. I suggest to push a branch with this change (say, auto-1.1) and to add a Jenkins build that verifes it, but SNAPSHOT dependencies can't be merged into master.

zack-shoylev:
Alright, but I think it makes more sense to just delay merging this? The snapshot part is to make sure the pr compiles. Also with luck they will release before we do.
________________________________________

andrewgaul:
:-1: Agree with @nacx, this is a good change but let's wait for the release.

nacx:
Fine by me to leave the PR open (although I'd prefer to have a dedicated branch so others can try and play with it easily). The important thing is to make sure we don't add SNAPSHOT dependencies to master.

zack-shoylev:
I am not sure what the dedicated branch adds here - do you mean adding it to cloudbees, or just a separate branch under jclouds?

nacx:
Well, a PR is no more than a branch in your fork. I mean pushing that branch to jclouds, so it is more visible and accessible to others, and add it to cloudbees, to have feedback of its status.

zack-shoylev:
@nacx @andrewgaul Good news, everybody! 1.1 is out.

zack-shoylev:
merged and backported

nacx:
Thanks @zack-shoylev!

We should avoid upgrading versions in patch releases, so I'd say we should revert this commit in 1.9.x. @andrewgaul WDYT?

zack-shoylev:
@nacx A big point when discussing the next 1.9.x release was making sure we have the builders available, as well as any new code that might use them.

andrewgaul:
Generally I don't think we should upgrade dependencies in minor releases, but since this one is compile-only it should not matter.

nacx:
Athough it is compile-only and won't produce classpath issues in existing deployments, take into account that it *might* produce different code, and that could present conflicts.

I agree this is unlikely to happen, so I'm OK leaving it in 1.9.x if you are confident.

Thx @zack-shoylev @andrewgaul!

PR #701, opened 2015-03-10
Update setupPublicCurl.sh

hviejo:
Add the fqdn to /etc/hosts file becouse if you only add the hostname then it is posible to lose the fqdn

### DIFF
--- a/scriptbuilder/src/main/resources/functions/setupPublicCurl.sh
+++ b/scriptbuilder/src/main/resources/functions/setupPublicCurl.sh
@@ -52,7 +52,7 @@ function ensure_hostname_in_hosts() {
     local ipaddr=`hostname -i`
   }
   # NOTE: we blindly trust existing hostname settings in /etc/hosts
-  egrep -q `hostname` /etc/hosts || echo "$ipaddr `hostname`" >> /etc/hosts
+  egrep -q `hostname -s` /etc/hosts || echo "$ipaddr `hostname -f` `hostname -s`" >> /etc/hosts
 }
 
 # download locations for many services are at public dns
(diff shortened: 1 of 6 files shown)


scriptbuilder/src/main/resources/functions/setupPublicCurl.sh
```diff
@@ -52,7 +52,7 @@ function ensure_hostname_in_hosts() {
     local ipaddr=`hostname -i`
   }
   # NOTE: we blindly trust existing hostname settings in /etc/hosts
-  egrep -q `hostname` /etc/hosts || echo "$ipaddr `hostname`" >> /etc/hosts
+  egrep -q `hostname` /etc/hosts || echo "$ipaddr `hostname -f` `hostname`" >> /etc/hosts
```
nacx: Should we better be explicit and use also the `-s` flag? Something like:
```bash
egrep -q `hostname -s` /etc/hosts || echo "$ipaddr `hostname -f` `hostname -s`" >> /etc/hosts
````

nacx:
Thanks @hviejo! Just one comment. Also, could you have a look at the test failures and fix them?
https://jclouds.ci.cloudbees.com/job/jclouds-pull-requests/1618/

hviejo:
Fixed all tests

nacx:
Thanks! WDYT about my comment of explicitly using the `-s` parameter too? In some operating systems the `-f` might be the default behavior.

hviejo:
Added "-s" option.
Now the /etc/hosts line is "ipAddress hostname -f hostname -s"

nacx:
Perhaps the `-s` parameter should also be used in the "grep" part? To make sure we look for the "smaller" hostname?

hviejo:
Yes, yes, the option are on both sides

nacx:
>Yes, yes, the option are on both sides

Oh, I didn't refresh the page and didn't see the last commit :) Cool! If you don't mind squashing the commits into a single one I'll merge the PR.

hviejo:
Ok, now it is squashed

nacx:
Pushed to master as [b54c788e](https://git1-us-west.apache.org/repos/asf?p=jclouds.git;a=commit;h=b54c788e). I've fixed a few more test files in the `compute` project that failed the build locally and amended the commit. I don't know why the pull request builder has not detected that, though.
Thanks @hviejo!

PR #702, opened 2015-03-10
Add utility to crawl a blobstore

andrewgaul:
This avoids error-prone boilerplate when using paginated lists.


andrewgaul:
@kahing Please review.

kahing:
:+1:

blobstore/src/main/java/org/jclouds/blobstore/util/BlobStoreUtils.java
```diff
@@ -85,4 +103,65 @@ private static void convertUserMetadataKeysToLowercase(MutableBlobMetadata metad
...
+         return iterator.hasNext() || marker != null;
+      }
+
+      @Override
+      public StorageMetadata next() {
+         while (true) {
+            if (!iterator.hasNext()) {
+               advance();
+            }
+            StorageMetadata metadata = iterator.next();
```
nacx: The advance method does not guarantee that the new iterator has a next element right?

nacx: To clarify: it is unlikely that the provider returns an empty page if there is a marker, but since it is quite a common pattern to use fallbacks that return empty lists on 404, we should take care and be defensive here. Don't assume that the iterable has content at this point.

Also IMO this class deserves its own unit test to verify it behaves properly when there are 0 pages (the initial advance returns nothing), a single page, multiple pages, and empty pages.

andrewgaul: I amended the pull request with the three tests you suggested and handing the presence of a marker with no more results.

nacx: Changes lgtm now. However, advancing the iterator in the `hasNext` is not the best pattern to use. I'd suggest to use the Guava's `AbstractIterator` to implement the iterator. You can take a look at the GCE's [AdvancingIterator](https://github.com/jclouds/jclouds-labs-google/blob/master/googlecloud/src/main/java/org/jclouds/googlecloud/internal/AdvancingIterator.java), which does pretty much the same. Your `crawl` methods should use that iterator and the [this concat method](https://github.com/jclouds/jclouds-labs-google/blob/master/googlecloud/src/main/java/org/jclouds/googlecloud/internal/ListPages.java#L34-44) to return an Iterator that can advance between pages, without having to change the state of the iterator in the `hasNext` method. WDYT?

andrewgaul: I rewrote this to use `AdvancingIterator` although I do not understand the suggestion to use concat -- could you expand on this?

nacx: Looking at your implementation, that shouldn't be needed. That concat would substitute your `while(true)` loop.

In the links I passed as an example, you have an `Iterator<Page<T>>`, that is, an iterator that knows how to "fetch pages", but does not iterate the elements of each page. The `concat` method is used to "flatten" that iterator in an iterator that iterates over the element of each page and jumps to the next one once finished. Basically, if you model it to be an Iterator of pages (which is what it really is), you can take benefit of Guava's `Iterators.concat(Iterator)` to do the "jump from one page to the next one" for you.

It is just an implementation detail. I just shared how pagination is implemented in other parts of the code (which IMO ends up being simpler code). The current implementation looks fine, BTW :)

andrewgaul:
Abandoning.  I believe we want something like this, but more discoverable, e.g., within `BlobStore` itself.

PR #703, opened 2015-03-11
Promote rackspace cloudfiles

zack-shoylev:
https://issues.apache.org/jira/browse/JCLOUDS-791


nacx:
Just curious... Why are there three merge commits at the end and not just one?

zack-shoylev:
Investigating!

PR #704, opened 2015-03-12
Promote rackspace cloudfiles

zack-shoylev:


zack-shoylev:
@nacx I followed the guide exactly this time and got a very similar result. History is preserved but commits that affected more than one of the merged folders appear multiple times as different hashes (which makes sense, because each of them was filtered differently - I think). I just don't know if this is going to be a problem. I tested against a fork and it seemed ok. Thoughts? Would you mind re-testing this? I also updated the guide to use : as the sed delimeter.

zack-shoylev:
BTW applying this with git am as a patch puts the commit history from this PR as most recent, it seems. As such, merging with git am will not work, from what I see.

zack-shoylev:
Tested in this fork https://github.com/zack-shoylev/jclouds/commits/master

zack-shoylev:
Welp, kinda stuck. Don't think this is working properly.

zack-shoylev:
Also see https://github.com/jclouds/jclouds/pull/705

zack-shoylev:
Superceded by https://github.com/jclouds/jclouds/pull/720

PR #705, opened 2015-03-13
Promote rackspace cloudfiles6

zack-shoylev:
I didn't like the repeating commits that the original guide produced (unless I did something wrong with it).
Instead, I decided to rewrite-history move everything to promote to jclouds/apis and jclouds/providers, then subtree split it. 

```bash
cd jclouds-labs-openstack
git checkout -b move-promote
git filter-branch -f --tree-filter 'if [ -d rackspace-cloudfiles ]; then mkdir -p jclouds/apis && mv rackspace-cloudfiles jclouds/apis; fi' HEAD
git filter-branch -f --tree-filter 'if [ -d rackspace-cloudfiles-us ]; then mkdir -p jclouds/providers && mv rackspace-cloudfiles-us jclouds/providers; fi' HEAD
git filter-branch -f --tree-filter 'if [ -d rackspace-cloudfiles-uk ]; then mkdir -p jclouds/providers && mv rackspace-cloudfiles-uk jclouds/providers; fi' HEAD
git subtree split -P jclouds -b promote-rackspace-cloudfiles-moved
```

switch to branch for PR, then:
```bash
git checkout -b promote-rackspace-cloudfiles6
git merge {remote if needed}/promote-rackspace-cloudfiles-moved
```


zack-shoylev:
I also rebased to get rid of the merge commit.

zack-shoylev:
@nacx Have a look at this when you have some time, and let me know if the steps make sense to you; also if my explanation about the multiple commits in the previous PR #704 make sense.

nacx:
This one looks much more clean. Thanks for the time and pain suffered to get into this nice state @zack-shoylev!

Just one comment though: I wouldn't rebase. When promoting providers with such an amount of commits, I think it is better to keep them "at the moment they were committed", and have the merge commit:

We want to have a linear history, but just because all contributions are assumed to be made against the latest version of the integration branch. That is OK for pull requests, but when promoting a provider, we are not adding *new* contributions, but adding to the repository contributions that were actually made some time before, so having all those commits on the top of the git history does not make sense to me and IMO does not provide any benefit. In this case, I think it is OK to have that merge commit, and I'd really prefer to promote providers and APIs without rebasing.

WDYT?

zack-shoylev:
@nacx I agree about not having the commits appear at the top of the list. However, without rebasing, I am not sure you will be able to keep a linear history in the repo? I can make another PR without the rebase so we can have a look at that.

nacx:
Without rebasing it won't be a linear history. But I think in this case it is OK, because we are not *adding new code on top of the current one* (which is what we do with PRs and patches); we are joining two repos, and in this case (IMO) makes much more sense *merging* them instead of trying to keep a linear history for all the incoming commits.

Also the merge commit shouldn't *complicate* the rebasing of existing PRs, as most of the commits being merged into this repo are older than the given PRs, and there won't be conflicts as the incoming files will be seen as *new* files here.

zack-shoylev:
@nacx Alright, this makes sense!

zack-shoylev:
https://github.com/jclouds/jclouds/pull/720

zack-shoylev:
Superceded by https://github.com/jclouds/jclouds/pull/720

PR #708, opened 2015-03-18
Deprecate Nova API for volumes in favour of Cinder API for volumes

everett-toews:
Helps reduce user confusion over which API to use too.


zack-shoylev:
Ah that @Deprecated @Beta. Looks good though, +1.

ccustine:
:+1:  HP provider has full Cinder support and is well tested.

everett-toews:
@nacx Shall we pull this into 1.9.0?

It's related to the bug that broke the rc and it's just a doc change so it's zero risk.

nacx:
Pushed to master as [fdb2cfe5](https://git1-us-west.apache.org/repos/asf?p=jclouds.git;a=commit;h=fdb2cfe5). Thanks!

PR #709, opened 2015-03-19
Modernize the VolumeAttachmentApiLiveTest

everett-toews:
To run the test on OpenStack (DevStack)

    mvn clean test -Plive -Dtest.openstack-nova.endpoint=http://$DEVSTACK_HOST:5000/v2.0/ -Dtest.openstack-nova.identity=demo:demo -Dtest.openstack-nova.credential=devstack -Dtest.openstack-nova.image-id=$IMAGE_ID -Dtest.openstack-nova.flavor-id=1 -Dtest=VolumeAttachmentApiLiveTest

To run the test on Rackspace

    mvn clean test -Plive -Dtest.openstack-nova.endpoint=https://identity.api.rackspacecloud.com/v2.0/ -Dtest.openstack-nova.identity=$RACKSPACE_USERNAME -Dtest.openstack-nova.credential=$RACKSPACE_PASSWORD -Dtest.openstack-nova.image-id=$IMAGE_ID -Dtest.openstack-nova.flavor-id=$FLAVOR_ID -Dtest.openstack-nova.region=IAD -Dtest.openstack-nova.volume-size-gb=100 -Dtest=VolumeAttachmentApiLiveTest

As expected, due to [this change](https://github.com/jclouds/jclouds/commit/2db654fbe3e32c3799cf2cb8999cd422878d3928#diff-5d6a573e2a00f18e2381f2e66c8469c7L62), this test currently fails on Rackspace with the error

    java.lang.IllegalStateException: Optional.get() cannot be called on an absent value
        at com.google.common.base.Absent.get(Absent.java:47)
        at org.jclouds.openstack.nova.v2_0.extensions.VolumeAttachmentApiLiveTest.setup(VolumeAttachmentApiLiveTest.java:72)

@zack-shoylev to provide the fix in a subsequent PR

### DIFF
--- a/apis/openstack-nova/pom.xml
+++ b/apis/openstack-nova/pom.xml
@@ -73,6 +73,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.apache.jclouds.api</groupId>
+      <artifactId>openstack-cinder</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.apache.jclouds</groupId>
       <artifactId>jclouds-compute</artifactId>
       <version>${project.version}</version>
--- a/apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/extensions/VolumeAttachmentApiLiveTest.java
+++ b/apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/extensions/VolumeAttachmentApiLiveTest.java
@@ -17,7 +17,6 @@
 package org.jclouds.openstack.nova.v2_0.extensions;
 
 import com.google.common.collect.FluentIterable;
-import com.google.common.collect.Iterables;
 import org.jclouds.ContextBuilder;
 import org.jclouds.openstack.cinder.v1.CinderApi;
 import org.jclouds.openstack.cinder.v1.domain.Volume;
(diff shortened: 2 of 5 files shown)


everett-toews:
Hopefully this is in reasonably good shape now. I'll be on vacation until Monday and won't have my laptop with me. If changes need to be made, go ahead and push on top of my commits or just cherry pick or whatever. I just don't want to hold the release up.

nacx:
Let's see if I understand this.

The volume test has started failing now, because modern OpenStack distributions come without the volume API in Nova, and the Available Volume API is through Cinder?

In that case, from what I see int he changed test we are no longer exercising the deprecated Nova Volume API. And having an API completely untested does not seem like a good approach. If those tests now belong to Cinder, what about moving the tests to that API (if missing there) and just leave the current ones unchanged? Instead of configuring them in the "live" group, we could move them to a "live-deprecated" group so they don't run by default, but anyone with an old Nova installation can run them if needed.

Am I understanding this correctly? WDYT?

zack-shoylev:
Superceded by https://github.com/jclouds/jclouds/pull/710

zack-shoylev:
Well one problem is that the current tests were broken from what I could tell. Which is why the idea was to replace them with up-to-date tests which follow what users would do (which is why @everett-toews also added the deprecation PR).

I don't think this tests a cinder feature; it just tests a nova feature that requires cinder to be tested properly.

We could add new working tests that test the old deprecated API - I think that would make sense. This might be a bit more difficult to setup for live testing.
Hopefully I am making sense :)

PR #710, opened 2015-03-20
Fix rackspace volume attach with modernized base tests

zack-shoylev:


zack-shoylev:
This supercedes https://github.com/jclouds/jclouds/pull/709
Meaning we can just merge this and it should fix the issues.

nacx:
Following the discussion in the other PR, I think I get it now. Cinder is used just to setup the volume, but the attachment api, which is the one being tested here, is the one in Nova, right? In that case lgtm!

zack-shoylev:
Alright @nacx. I am about to rebase and merge this

zack-shoylev:
merged

nacx:
I've tried to run the live test on HPCloud as follows:

```bash
mvn test -Plive -Dtest.openstack-nova.endpoint="https://region-a.geo-1.identity.hpcloudsvc.com:35357/v2.0/" \
    -Dtest.openstack-nova.identity="<identity>" \
    -Dtest.openstack-nova.credential="<credential>" \
    -Dtest.openstack-nova.image-id=bec3cab5-4722-40b9-a78a-3489218e22fe \
    -Dtest.openstack-nova.flavor-id=standard.small  \
    -Djclouds.keystone.credential-type=apiAccessKeyCredentials \
    -Dtest.openstack-nova.region=region-a.geo-1
    -Dtest=VolumeAttachmentApiLiveTest
```

But I get the following failure, in [this line](https://github.com/jclouds/jclouds/blob/master/apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/extensions/VolumeAttachmentApiLiveTest.java#L84). Is it normal that the attachment api is not available in HPCloud? /cc @ccustine @everett-toews 

```bash
setup(org.jclouds.openstack.nova.v2_0.extensions.VolumeAttachmentApiLiveTest)  Time elapsed: 7.562 sec  <<< FAILURE!
java.lang.IllegalStateException: Optional.get() cannot be called on an absent value
	at com.google.common.base.Absent.get(Absent.java:47)
	at org.jclouds.openstack.nova.v2_0.extensions.VolumeAttachmentApiLiveTest.setup(VolumeAttachmentApiLiveTest.java:84)
```

nacx:
I've run the same test with the RC1 version (without the applied patches) and the volume attachment API is also absent for HP cloud there.

@ccustine @everett-toews @zack-shoylev Is this ok? If not, I think we should report it as a known issue (as no one have complaint about it before and it already existed) and release 1.9.0 with the provided patch. WDYT?

ccustine:
@nacx @everett-toews @zack-shoylev I am delivering a project this week so I have been heads down crunching on that, but to be honest I don't totally understand the issue as it related to HP Cloud.  HPs nova supports os-volumes extension and IIRC cinder already has the attachment support baked in. Let me know if I am wrong about that, and specifically let me know if there is any specific test you want me to  run and report.  I tried to run the tests against RC1 and I had failures for other (unrelated) issues so it seems.

nacx:
Ok, I've opened https://github.com/jclouds/jclouds/pull/711 to fix the attachment in HPCloud. Basically it requires the same attachment extension namespace than Rackspace, and the test needed to be a bit more generic, as the Cinder version provided by HPCloud is 1.0, while the Compute version is v2.

@zack-shoylev @everett-toews I've seen that the RAX UK provider was not updated with the volume attachment changes. Should it be updated too? If so, can you please raise a PR? (I don't have an account there to test it).

PR #711, opened 2015-03-23
Fixed VolumeAttachment in HPCloud Compute

nacx:
Fixes the volume extension namespace in the HPCloud Compute provider.
This PR also adds a custom method to make it easier to override the default OpenStack namespaces without having to copy/paste the entire Guice module, and allows to configure the Cinder version in the tests (which could be different than the provider's compute version, for example).

/cc @zack-shoylev @ccustine @everett-toews

While fixing this I've noticed that the RAX UK provider was not updated with the volume attachment fixes. Should that be done?

### DIFF
--- a/apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/config/KeystoneHttpApiModule.java
+++ b/apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/config/KeystoneHttpApiModule.java
@@ -49,11 +49,11 @@ import com.google.common.base.Suppliers;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
-import com.google.common.collect.ImmutableMultimap;
-import com.google.common.collect.Multimap;
 import com.google.inject.AbstractModule;
+import com.google.inject.Binder;
 import com.google.inject.Provides;
 import com.google.inject.assistedinject.FactoryModuleBuilder;
+import com.google.inject.multibindings.MapBinder;
 
 /**
  * Configures the Keystone API.
@@ -91,17 +91,16 @@ public class KeystoneHttpApiModule extends HttpApiModule<KeystoneApi> {
       }
    }
 
+   // Allow providers to cleanly contribute their own aliases
+   public static MapBinder<URI, URI> aliasBinder(Binder binder) {
+      return MapBinder.newMapBinder(binder, URI.class, URI.class, Aliases.class).permitDuplicates();
+   }
+
    @Override
    protected void configure() {
       bind(ImplicitOptionalConverter.class).to(PresentWhenExtensionAnnotationNamespaceEqualsAnyNamespaceInExtensionsSet.class);
       super.configure();
-   }
-
-   @Provides
-   @Singleton
-   public Multimap<URI, URI> aliases() {
-       return ImmutableMultimap.<URI, URI>builder()
-          .build();
+      aliasBinder(binder());
    }
 
    @Provides
--- a/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/config/NovaHttpApiModule.java
+++ b/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/config/NovaHttpApiModule.java
@@ -16,6 +16,8 @@
  */
 package org.jclouds.openstack.nova.v2_0.config;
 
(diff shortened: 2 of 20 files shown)


nacx:
Hey jenkins, rebuild please

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/config/NovaHttpApiModule.java
```diff
@@ -62,7 +62,7 @@ protected void configure() {
    @Provides
    @Singleton
    public Multimap<URI, URI> aliases() {
-       return ImmutableMultimap.<URI, URI>builder()
+       return customAliases(ImmutableMultimap.<URI, URI>builder()
```
zack-shoylev: Good approach. It works for both the HP and rackspace providers. Thanks @nacx !

nacx:
The last change completely refactors how the extension namespace aliases are configured. Now aliases are configured using Guice multibindings to allow providers to contribute their own aliases in a clean way.

This also fixes the HPCloud volume attachment configuration and adds the corresponding live test, configures the Rackspace US provider accordingly, and supercedes https://github.com/jclouds/jclouds/pull/712.

nacx:
Jenkins, rebuild please

nacx:
Also note that the bindings configured in the providers are *added* to the aliases map; they don't override the existing keys, but that should be aligned with the current behavior, that already expects more than one namespace URI for a given extension type (see [here](https://github.com/jclouds/jclouds/blob/master/apis/openstack-keystone/src/main/java/org/jclouds/openstack/v2_0/functions/PresentWhenExtensionAnnotationNamespaceEqualsAnyNamespaceInExtensionsSet.java#L63-L69) and [here](https://github.com/jclouds/jclouds/blob/master/apis/openstack-keystone/src/main/java/org/jclouds/openstack/v2_0/predicates/ExtensionPredicates.java#L88)).

nacx:
Once this is in, [this commit](https://github.com/nacx/jclouds-labs-openstack/commit/8b252a4f3487a5533ace98da14f76c8e799810ad) will need to be cherry-picked in jclouds-labs-openstack to align Neutron with the changes in this PR. /cc @zack-shoylev

everett-toews:
:+1: 

Success on OpenStack Juno (with DevStack)

    mvn clean test -Plive -Dtest.openstack-nova.endpoint=http://$DEVSTACK_HOST:5000/v2.0/ -Dtest.openstack-nova.identity=demo:demo -Dtest.openstack-nova.credential=devstack -Dtest.openstack-nova.image-id=$IMAGE_ID -Dtest.openstack-nova.flavor-id=2 -Dtest=VolumeAttachmentApiLiveTest

Success on Rackspace

    mvn clean test -Plive -Dtest.rackspace-cloudservers-us.identity=$RACKSPACE_USERNAME -Dtest.rackspace-cloudservers-us.credential=$RACKSPACE_API_KEY -Dtest.rackspace-cloudservers-us.image-id=$IMAGE_ID -Dtest.rackspace-cloudservers-us.flavor-id=performance1-1 -Dtest.rackspace-cloudservers-us.region=IAD -Dtest.rackspace-cloudservers-us.volume-size-gb=100 -Dtest=CloudServersUSVolumeAttachmentExtensionLiveTest

zack-shoylev:
Seems to work with both UK and US Rackspace providers!

apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/config/KeystoneHttpApiModule.java
```diff
@@ -91,17 +91,16 @@ protected void configure() {
       }
    }
 
+   // Allow providers to cleanly contribute their own aliases
```
zack-shoylev: Maybe add some more info about the mechanics of it.

nacx:
Pushed to master as [4db2b75a](https://git1-us-west.apache.org/repos/asf?p=jclouds.git;a=commit;h=4db2b75a).

apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/config/KeystoneHttpApiModule.java
nacx: The [MapBinder](http://google.github.io/guice/api-docs/latest/javadoc/index.html?com/google/inject/multibindings/MapBinder.html) javadoc already has the detailed documentation of how it works and several examples. Replicating it here seems unnecessary. The object returned by this method already has the docs describing its usage.

nacx:
Thanks for helping with the testing @zack-shoylev, @everett-toews!

PR #712, opened 2015-03-23
Volume attachment fix for cloudservers-uk

zack-shoylev:
Also uses @nacx better approach to making NovaHttpApi aliases overrideable.


nacx:
This PR should be closed now as the changes here are already included in https://github.com/jclouds/jclouds/commit/4db2b75addf9730d66ddde166a5d0e464434478f. Thanks @zack-shoylev!

PR #713, opened 2015-03-24
Forwarding and read-only BlobStores

andrewgaul:


blobstore/src/main/java/org/jclouds/blobstore/util/ForwardingBlobStore.java
```diff
@@ -0,0 +1,210 @@
...
+      this.blobStore = checkNotNull(blobStore);
+   }
+
+   protected BlobStore delegate() {
+      return blobStore;
+   }
+
+   @Override
+   public BlobStoreContext getContext() {
+      return delegate().getContext();
```
andrewgaul: Note that there is an unintentional escape hatch from the `ForwardingObject`:

```java
BlobStore rwBlobStore = ...;
BlobStore roBlobStore = ReadOnlyBlobStore.newReadOnlyBlobStore(rwBlobStore);
roBlobStore.getContext().getBlobStore().putBlob(...);
```

PR #715, opened 2015-03-29
Enable Google error-prone for builds

andrewgaul:


andrewgaul:
Tested with JDK 1.8.0_25.

andrewgaul:
rebuild please

PR #716, opened 2015-03-30
Return ETag from Azure setBlobMetadata

andrewgaul:
Also fix up live tests.


providers/azureblob/src/main/java/org/jclouds/azureblob/AzureBlobClient.java
```diff
@@ -415,7 +415,8 @@ BlobProperties getBlobProperties(
    @PUT
    @Path("{container}/{name}")
    @QueryParams(keys = { "comp" }, values = { "metadata" })
-   void setBlobMetadata(
+   @ResponseParser(ParseETagHeader.class)
+   String setBlobMetadata(
```
andrewgaul: Theoretically this represents an interface change although practically it does not.

andrewgaul:
rebuild please

andrewgaul:
rebuild please

PR #718, opened 2015-03-31
Allows copying an object and modifying metadata (user and object)

zack-shoylev:


zack-shoylev:
https://issues.apache.org/jira/browse/JCLOUDS-872

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/ObjectApi.java
```diff
@@ -262,4 +263,36 @@ boolean copy(@PathParam("destinationObject") String destinationObject,
...
+   @Named("object:copy")
+   @PUT
+   @Path("/{destinationObject}")
+   @Headers(keys = OBJECT_COPY_FROM, values = "/{sourceContainer}/{sourceObject}")
+   @Fallback(FalseOnContainerNotFound.class)
+   boolean copy(@PathParam("destinationObject") String destinationObject,
+         @PathParam("sourceContainer") String sourceContainer,
+         @PathParam("sourceObject") String sourceObject,
+         @BinderParam(BindObjectMetadataToHeaders.class) Map<String, String> userMetadata,
+         @BinderParam(BindHeaderMetadataToHeaders.class) Map<String, String> objectMetadata);
```
andrewgaul: Should we reverse these parameters?  Also does this allow an easy way to copy the metadata from an existing object?  I would like to make the simple case of copying all objects from one container to another as easy as possible.

zack-shoylev: A very good point. I will investigate.

zack-shoylev:
I have added a live test that demonstrates how the copy would work with real conditions.
Unfortunately, the content- metadata is handled separately and is also filtered from the headers. Which makes all of this very clunky! However, I did not want to address the header-filtering in this PR (or maybe at all).

andrewgaul:
Pushed to master as a1cbec10925a05bf0a05c6ddb3f9eb51c7f155ea and 576005a33553d3b682bdd7088c10c5e7da1c004e.

PR #719, opened 2015-04-03
JCLOUDS-651: Copy Swift system metadata

andrewgaul:


andrewgaul:
Follow-on to #718.

zack-shoylev:
Have you had a chance to run the live tests? I thought swift copies the metadata and we just need to override it, based on this in the docs:
"All metadata is preserved during the object copy."

zack-shoylev:
I will get back to this tomorrow if needed, but go ahead and merge if you think it's ready @andrewgaul

andrewgaul:
Agree with your read of the documentation and its attached example, although running without this change I see this failure:

```
testCopyBlobCopyMetadata(org.jclouds.rackspace.cloudfiles.us.blobstore.integration.CloudFilesUSBlobIntegrationLiveTest)  Time elapsed: 38.811 sec  <<< FAILURE!
java.lang.NullPointerException: null
        at org.jclouds.blobstore.integration.internal.BaseBlobIntegrationTest.checkContentDisposition(BaseBlobIntegrationTest.java:666)
        at org.jclouds.blobstore.integration.internal.BaseBlobIntegrationTest.checkContentMetadata(BaseBlobIntegrationTest.java:646)
        at org.jclouds.blobstore.integration.internal.BaseBlobIntegrationTest.testCopyBlobCopyMetadata(BaseBlobIntegrationTest.java:762)
```

Maybe we are triggering this behavior by accidentally specifying a header which triggers replacement of others?

zack-shoylev:
I suspect that we send the content metadata headers regardless and overwrite. However, you will not see them in the log

zack-shoylev:
So yeah, I suspect this is catching a bug with how we send the copy request when no options are specified. This arises with how contentmetadata is handled in core jclouds, if i remember correctly.

andrewgaul:
@zack-shoylev Perhaps we are hitting one of the consequences of [JCLOUDS-595](https://issues.apache.org/jira/browse/JCLOUDS-595).  I investigated removing a default Content-Type but this needs more work:

https://github.com/andrewgaul/jclouds/tree/no-default-content-type

Maybe you would like to pick this up since it enables Swift/CloudFiles to use its auto-detect Content-Type?

zack-shoylev:
Yep, exactly. However, I also wanted to address all the content metadata in general. But. That's a lot of work and affects multiple providers. Some choices:
1. Refactor properly and change the provider/api code as needed. Remove default behavior. Treat all headers the same way, maybe. Get rid of how content- headers work now. Hope for the best.
2. Keep the default behavior but add some options that allow us to modify that behavior from a provider/api, so not all providers are affected. For example a NO_DEFAULT_CONTENT_META property or similar.
3. Other ideas?

PR #722, opened 2015-04-07
Allow overriding content metadata during copyBlob

andrewgaul:
Add Azure, S3, and Swift variants.

### DIFF
--- a/blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
+++ b/blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
@@ -62,6 +62,7 @@ import org.jclouds.blobstore.options.PutOptions;
 import org.jclouds.crypto.Crypto;
 import org.jclouds.encryption.internal.JCECrypto;
 import org.jclouds.http.HttpResponseException;
+import org.jclouds.io.ContentMetadataBuilder;
 import org.jclouds.io.Payload;
 import org.jclouds.io.Payloads;
 import org.jclouds.io.payloads.ByteSourcePayload;
@@ -777,7 +778,11 @@ public class BaseBlobIntegrationTest extends BaseBlobStoreIntegrationTest {
             .blobBuilder(fromName)
             .userMetadata(ImmutableMap.of("key1", "value1", "key2", "value2"))
             .payload(payload)
-            .contentLength(payload.size());
+            .contentLength(payload.size())
+            .contentDisposition("attachment; filename=original.jpg")
+            .contentEncoding("compress")
+            .contentLanguage("fr")
+            .contentType("audio/ogg");
       addContentMetadata(blobBuilder);
       Blob blob = blobBuilder.build();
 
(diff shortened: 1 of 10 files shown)


blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
```diff
@@ -796,9 +808,8 @@ public void testCopyBlobReplaceMetadata() throws Exception {
          } finally {
             Closeables2.closeQuietly(is);
          }
-         // TODO: S3 overrideMetadataWith also overrides system metadata
-         // TODO: Swift does not preserve system metadata
-         //checkContentMetadata(toBlob);
+         // TODO: Swift copy object *appends* user metadata, does not overwrite
```
andrewgaul: @zack-shoylev Curious Swift *appends* user metadata so this test fails with:

```
Failed tests:
  CloudFilesUSBlobIntegrationLiveTest>BaseBlobIntegrationTest.testCopyBlobReplaceMetadata:813
Expecting:
 <{"key2"="value2", "key3"="value3", "key1"="value1", "key4"="value4"}>
to be equal to:
 <{"key3"="value3", "key4"="value4"}>
but was not.
```

Swift documents this behavior so this is expected.  I guess we just need to change the test expectations?

zack-shoylev: I am not completely sure. Shouldn't we be trying to have the same outcome? The question is if metadata should be completely overridden or appended when using the abstraction, right?

andrewgaul: Alternatively we could try to fix up the metadata after the copy.  Agreed that the portable abstraction should act the same between providers.

zack-shoylev: You could query the source metadata before copy and apply that during the copy operation that overrides it. Otherwise you will have to do two operation that modify an object, instead of a read and a copy. Thoughts?

andrewgaul: I don't understand how to do this fixup with the current API -- `ObjectApi.updateMetadata` writes user metadata but overwrites content metadata and `ObjectApi.updateRawMetadata` has some strange semantics.

andrewgaul:
This works pretty well across the major providers -- should we merge as-is and open a JIRA issue for the Swift quirk?

zack-shoylev:
@andrewgaul I'm ok with that. Figuring out the swift quirk might need some more work.

andrewgaul:
OK I'll open a JIRA issue and skip the test on Swift before merging.

andrewgaul:
Opened [JCLOUDS-883](https://issues.apache.org/jira/browse/JCLOUDS-883).

zack-shoylev:
backport?

andrewgaul:
Backported all the copyBlob commits to 1.9.x.

PR #724, opened 2015-04-07
JCLOUDS-876: Properly clone the login credentials

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-876

@ilgrosso I opted to add your code snippet to the `copyTo` method. The other approach looked cleaner, but initialising the options presented some issues [here](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/internal/BaseComputeService.java#L697-L710): the ComputeService overrides the node credentials if the user provided them in the template options, but if we initialise them to absent, there is no clean way of knowing if credentials were manually set by the user or not (some could override the credentials to `absent()` to force the authentication using the SSH agent).

Given this, I think it is better to fix the copy issue in the `copyTo` method. I will run some live tests against several providers to make sure this change does not break existing tests. Can you run the Azure tests with this change?


nacx:
rebuild please!

nacx:
Weird failures... Let's try once again. rebuild please.

ilgrosso:
@nacx Everything working as expected: looking forward for this PR to get merged

nacx:
Pushed to master and 1.9.x.

PR #725, opened 2015-04-07
Update cloudfiles poms

zack-shoylev:


zack-shoylev:
Rebuild please?

nacx:
Looks good. A couple comments:

* Should the exising `cloudfiles` providers be removed? Are they superceded by the promoted ones?
* Should the promoted ones be added to the `allblobstore` pom?

zack-shoylev:
That would be the plan going forward, though maybe not as part of this PR. I am not sure what the right timing would be to do that, yet.

nacx:
Do you refer to the removal of the existing cloudfiles provider, or to adding the new provides to the blobstore aggregator pom?

zack-shoylev:
Both

nacx:
+1 to merge it as-is then. Thanks!

What are waiting to remove the old providers? Is it something to deprecate/announce/discuss in the mailing list?

zack-shoylev:
merged and backported

PR #726, opened 2015-04-07
Ensures files generated on Windows while building are not checked.

zack-shoylev:
For example:

```
[INFO] --- maven-checkstyle-plugin:2.15:check (default) @ jclouds-resources ---
[WARNING] target\maven-archiver\pom.properties[1] (regexp) RegexpMultiline: Do not use Windows line endings
[WARNING] target\maven-archiver\pom.properties[2] (regexp) RegexpMultiline: Do not use Windows line endings
[WARNING] target\maven-archiver\pom.properties[3] (regexp) RegexpMultiline: Do not use Windows line endings
[WARNING] target\maven-archiver\pom.properties[4] (regexp) RegexpMultiline: Do not use Windows line endings
[WARNING] target\maven-archiver\pom.properties[5] (regexp) RegexpMultiline: Do not use Windows line endings
[INFO] ------------------------------------------------------------------------
```

### DIFF
--- a/resources/checkstyle.xml
+++ b/resources/checkstyle.xml
@@ -91,5 +91,6 @@
     <module name="RegexpMultiline">
         <property name="format" value="\r\n"/>
         <property name="message" value="Do not use Windows line endings"/>
+        <property name="fileExtensions" value="java,xml,json"/>
     </module>
 </module>


andrewgaul:
Odd, why does Checkstyle look at the target directory?

zack-shoylev:
I would like to know that too! It would be better to exclude it, rather than do it by extensions. Just unsure how to do that, though.

zack-shoylev:
@andrewgaul is this good to merge? Or do we need to track it down.

zack-shoylev:
merged
not backported (not applicable)

PR #727, opened 2015-04-08
replace ExpectTests with MockWebServerTests

andreaturli:
- refactor AccountApiExpectedTest to AccountApiMockTest
- refactor DatacenterApiExpectTest to DatacenterApiMockTest
- refactor SoftwareDescriptionApiExpectTest to SoftwareDescriptionApiMockTest
- refactor VirtualGuestApiExpectTest to VirtualGuestApiMockTest
- refactor VirtualGuestBlockDeviceTemplateGroupApiExpectTest to VirtualGuestBlockDeviceTemplateGroupApiMockTest

@nacx could you have a quick look, please?


providers/softlayer/src/test/java/org/jclouds/softlayer/internal/BaseSoftLayerMockTest.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.inject.Module;
+import com.squareup.okhttp.mockwebserver.MockWebServer;
+import com.squareup.okhttp.mockwebserver.RecordedRequest;
+
+/**
+ * Base class for all Docker mock tests.
+ */
+public class BaseSoftLayerMockTest extends BaseMockWebServerTest {
+
+   protected static final String API_VERSION = "1";
```
nacx: Better get an instance of the `SoftLayerApiMetadata` and get its version.

nacx:
Looks good @andreaturli but the mock tests are missing the tests that exercise the defined fallbacks. Could you add those too?

andreaturli:
@nacx thanks for the review.

I think we agreed on not writing tests for pre-defined fallbacks as those fallbacks are already tested in isolation. At this regard, what's the rule-of-thumb for `list` methods: is it always needed to have something like `@Fallback(EmptySetOnNotFoundOr404.class)` to avoid NPE or it depends on the API?

nacx:
By pre-defined fallbacks we mean implicit ones. For example, if you don't define a fallback, you don't need to add a test that verifies that an HTTPException is thrown, or a ResourceNotFoundException is thrown on a 404 response. jclouds already has a set of default fallbacks that throw the appropriate exception and we don't want to test that as it is already tested in jclouds-core.

However, when explicitly adding fallbacks to an API method, we need to test they're taking place. A `list` method could use the "empty list" fallback, but also the "empty paged iterable" if it is a paginated api call, so there is no real default here. If a fallback is explicitly defined (that means, the default one is overridden) then a test should be added. Makes sense?

andreaturli:
absolutely @nacx! thanks again!

andreaturli:
@nacx could we merge that now?

nacx:
Yes, lgtm. Huge thanks for this PR @andreaturli!

zack-shoylev:
So much cleanup, thanks!

andreaturli:
Merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/7480d6e3) and backport to [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a0ada77e)

PR #731, opened 2015-04-15
Exclude Eclipse annotation processor files that are git ignored

nacx:


nacx:
Merged and backported to 1.9.x

PR #733, opened 2015-04-23
Adds windows ACL code for file permissions, fixes build.

zack-shoylev:


apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -154,32 +161,95 @@ public boolean createContainerInLocation(String container, Location location, Cr
...
-   public ContainerAccess getContainerAccess(String container) {
+   public synchronized ContainerAccess getContainerAccess(String container) {
       Path path = new File(buildPathStartingFromBaseDir(container)).toPath();
-      Set<PosixFilePermission> permissions;
-      try {
-         permissions = getPosixFilePermissions(path);
-      } catch (IOException ioe) {
-         throw Throwables.propagate(ioe);
+
+      if ( System.getProperty("os.name", "").toLowerCase().contains("windows") ) {
```
nacx: Extract this to a method in the [TestUtils](https://github.com/jclouds/jclouds/blob/master/core/src/test/java/org/jclouds/utils/TestUtils.java) class, similar to what we already do with OSX.

nacx: Uops, didn't notice this was not a test. Forget about that comment.

nacx: Anyway, extracting this to a method so it can be reused seems like a good idea.

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -154,32 +161,95 @@ public boolean createContainerInLocation(String container, Location location, Cr
...
-            permissions.add(PosixFilePermission.OTHERS_READ);
+
+      if ( System.getProperty("os.name", "").toLowerCase().contains("windows") ) {
+         UserPrincipal everyone = null;
+         UserPrincipal me = null;
+         try {
+            everyone = getDefault().getUserPrincipalLookupService()
+                  .lookupPrincipalByName("Everyone");
+         } catch (IOException e) {
+            e.printStackTrace();
```
nacx: Use a logger or remove this.

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -154,32 +161,95 @@ public boolean createContainerInLocation(String container, Location location, Cr
...
+            return;
```
nacx: When reading, if this exception happens, we propagate an exception, but when writing, we just silently return and ignore it? Can we change one or the other to have a consistent behavior?

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -154,32 +161,95 @@ public boolean createContainerInLocation(String container, Location location, Cr
...
+                     AclEntryPermission.READ_ACL,
+                     AclEntryPermission.READ_ATTRIBUTES,
+                     AclEntryPermission.READ_NAMED_ATTRS)
+                     .setType(AclEntryType.ALLOW)
+                     .build());
+               aclFileAttributes.setAcl(list);
+            }
+         } catch (IOException e) {
+            e.printStackTrace();
+            return;
```
nacx: Same comments about exception handling.

nacx:
There is several repeated logic in this PR. Could you extract it to a WindowsUtils (or whatever) class, to encapsulate it there?

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
zack-shoylev: Yep, I messed up. I think we should propagate in both cases.

zack-shoylev:
It should be easy to refactor.

zack-shoylev:
About to rebase, merge, backport.

zack-shoylev:
merged and backported

nacx:
Much cleaner. Thanks @zack-shoylev!

PR #734, opened 2015-04-23
JCLOUDS-890: Prevent NPE when generating the Chef attributes file

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-890


zack-shoylev:
looks good

nacx:
Thx @zack-shoylev!

PR #738, opened 2015-05-01
JCLOUDS-897: Remove the Rocoto dependency

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-897

Remove the Rocoto dependency but keep the property expand feature.

Note that this patch also adds an exclusion for the `StringBuffer` constructor, as it is the object required by the java `Matcher` to perform the string replacements. We should be able to remove the exclusion once https://github.com/andrewgaul/modernizer-maven-plugin/issues/3 is fixed. /cc @andrewgaul


nacx:
Pushed to master.

PR #739, opened 2015-05-01
Remove the deprecated methods in Chef

nacx:
Removes legacy code:

* Removes the deprecated methods and objects for the next release.
* Removes the uses of deprecated Guava methods.
* Drops support for Chef 9. It is a very very old version now, and supporting it does no longer make sense. There have been three major Chef releases (10, 11 and 12) since then which we support, and the vast majority of the ecosystem has upgraded to one of them.


nacx:
Merged to master as [4863bb83](http://git-wip-us.apache.org/repos/asf/jclouds/commit/4863bb83).

PR #741, opened 2015-05-06
JCLOUDS-207: Key Pair and Security Groups created by jclouds are not rem...

stuarthendren:
...oved when the node is destroyed

The names created do not match those searched for.

They are created in FormatSharedNamesAndAppendUniqueStringToThoseWhichRepeat.java
and have are of this form jclouds#I-0#e96.

But for example jclouds#I-0#us-east-1#* is used as the search term.

### DIFF
--- a/apis/ec2/src/main/java/org/jclouds/ec2/compute/EC2ComputeService.java
+++ b/apis/ec2/src/main/java/org/jclouds/ec2/compute/EC2ComputeService.java
@@ -235,7 +235,7 @@ public class EC2ComputeService extends BaseComputeService {
       for (KeyPair keyPair : client.getKeyPairApi().get().describeKeyPairsInRegionWithFilter(region,
               ImmutableMultimap.<String, String>builder()
                       .put("key-name", Strings2.urlEncode(
-                              String.format("jclouds#%s#%s*", group, region).replace('#', delimiter)))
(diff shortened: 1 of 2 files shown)


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/31749cba) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/31749cba). Thanks @stuarthendren!

nacx:
I'm not very familiar with the issue, so can the jira issue be considered as fixed now?

stuarthendren:
I think this only fixes a sub-issue with the ticket.

On some tests I ran, you sill get SecurityGroups and KeyPairs left behind (but much less).

However, I think the remaining instances are really related to the issue [JCLOUDS-529](https://issues.apache.org/jira/browse/JCLOUDS-529) which has been closed, but still causes things to be left behind. The fix that closes it just changes the exception type, I don't think it solves the underlying issue. 

The test code I added to that ticket showed it can sometimes take 60s or more for a SecurityGroup to be released (i.e. no longer be dependent on the instance being deleted and therefore not deletable). I currently catch that exception and retry deleting the security group for longer which solves the problem for me. 

Maybe pulling that timing out into a variable would be a sensible solution. So the time trying to delete it can be set to be longer if you care more about it, but don't have to wait so long if you don't mind the artefacts being left behind.

PR #747, opened 2015-05-12
JCLOUDS-904 Fixing OSGi Header by adding org.jclouds.io import

agrz:
This commit fixes JCLOUDS-904 by adding the missing import in pom.xml for generation of correct OSGI Header during build


nacx:
Seems an unrelated build failure. Let's see if jenkins is happy now: rebuild please

BTW, is there a test that can be added in a follow-up PR in jclouds-karaf, to make sure this issue does not show up again?

andrewgaul:
@iocanel Can you also review this?

nacx:
@agrz ping :)

agrz:
Sorry, I wasn't aware you are waiting for me :)
The issue only shows up during provisioning a real S3 service on Karaf. In fact the AwsS3LiveTest.java already in the project would already test it. But the live tests are excluded in the pom.xml. Are they executed at any time during build/release process?

nacx:
Live tests are still not added to our CI, as we don't have credentials for every provider, so we run them manually when performing the releases. It is clear that the AWS karaf live tests weren't run in the last one, but it's good to know the test is there. Thanks for the feedback!

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a809c11a) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/b1231257). Thanks @agrz!

PR #752, opened 2015-05-27
implemented some basic range validations in local blob store

kahing:
also fixed a range get bug in local blob store


andrewgaul:
Pushed to master as 1103a778d2936f262b900767e823279677d5aec0.

PR #753, opened 2015-05-27
Propagate an AuthorizationException on OAuth client errors

nacx:
OAuth will be used as an authentication filter, so all client errors produced in the OAuth calls should be treated as an `AuthorizationException`. This will allow providers and clients to properly handle and react to OAuth specific errors.

/cc @danbroudy


nacx:
Pushed to master as [97c16087](http://git-wip-us.apache.org/repos/asf/jclouds/commit/97c16087).

PR #754, opened 2015-05-27
Update labs dependencies for rackspace-cloudfiles

andrewgaul:


andrewgaul:
@zack-shoylev Should we remove the labs providers now that they are in core?

andrewgaul:
rebuild please

zack-shoylev:
@andrewgaul I think we should, but I don't know if that should also happen in 1.9.x

PR #757, opened 2015-05-28
metadata and content metadata should have the same length

kahing:
for range get in a LocalBlobStore, metadata.getSize() has the
original blob size and contentMetadata.getContentLength() has
the actual payload size. Other blobstores have the same size
in both


andrewgaul:
Pushed to master as 573ceb51c36b78bd1c84ed095567c67975c275cf and fixed Atmos in 454a30a4846785dbfa214dc2c1250c86225dc3b5.

PR #755, opened 2015-05-27
tighten the isUrlEncoded check

kahing:
ideally we shouldn't need this function and instead never double
encode strings, but auditing for that is beyond what I have time
for. currently, putBlob(" ") and putBlob("%20") behave the same
way which is arguably incorrect

PR #758, opened 2015-05-29
tighten the isUrlEncoded check

kahing:
ideally we shouldn't need this function and instead never double
encode strings, but auditing for that is beyond what I have time
for. currently, putBlob(" ") and putBlob("%20") behave the same
way which is arguably incorrect

Old PR is https://github.com/jclouds/jclouds/pull/755 . This worksaround some of the test failures

### DIFF
--- a/core/src/main/java/org/jclouds/util/Strings2.java
+++ b/core/src/main/java/org/jclouds/util/Strings2.java
@@ -90,11 +90,34 @@ public class Strings2 {
    public static boolean isCidrFormat(String in) {
       return CIDR_PATTERN.matcher(in).matches();
    }
-      
+
+   // taken from https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#legal-chars
+   private static final Pattern URL_VALID_PATTERN = Pattern.compile("[a-zA-Z0-9_\\-!.~'()*,;:$&/\\[\\]@%]+");
    private static final Pattern URL_ENCODED_PATTERN = Pattern.compile(".*%[a-fA-F0-9][a-fA-F0-9].*");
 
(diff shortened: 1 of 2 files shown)


kahing:
tried running the core tests and s3 integration test, everything passed for me

demobox:
Still lots of test failures, I'm afraid :-(

https://jclouds.ci.cloudbees.com/job/jclouds-pull-requests/1799/testReport/

Before we try to dig into those, could you perhaps explain _why_ this change is needed? I don't disagree that the behaviour is arguably not correct, but as you can see this is used all over the place in jclouds, so it would seem sensible not to change it unless there's a compelling use case...

kahing:
Ugh this is messier than I thought.

The original motivation was to make https://github.com/bouncestorage/swiftproxy pass swift's functional tests, which tries to put a blob named "dealde%2Fl04 011e%204c8df/flash.png", which won't be escaped by jclouds and causes a number of failures. I ended up escaping blobs before passing to jclouds, which causes other problems.

In particular, when we call putBlob("dealde%252Fl04%20011e%25204c8df/flash.png") jclouds would send the path as is, which causes the server side to interpret it as "dealde%2Fl04 011e%204c8df/flash.png", and so the latter is what what would be returned from list(). So now what we put() and what we get from list() are different. 

Some o this is captured in https://issues.apache.org/jira/browse/JCLOUDS-918

demobox:
> The original motivation was to make https://github.com/bouncestorage/swiftproxy pass swift's functional tests

@kahing Thanks for clarifying. Yes, the list/put divergence is not fun. Hm.

andrewgaul:
Superseded by #851.

PR #759, opened 2015-05-30
Move logger and payer S3 tests to AWS provider

andrewgaul:
Most S3 clones do not support these operations.


andrewgaul:
:-1: We should move the functionality along with the tests.

PR #760, opened 2015-05-31
remove blob uri from transient blobstore

kahing:
container may not be valid hostnames and the current blob uri
code can fail. there isn't really a way to use uri for transient
blobstore anyway, and removing it is the easier option. for
reference, filesystem blobstore also doesn't do this


andrewgaul:
Agreed that this is useless.

PR #735, opened 2015-04-24
Use ssh agent if privatekey and password are not provided

hlta:
fixed according to comments. Thanks

PR #761, opened 2015-06-01
JCLOUDS-921 prioritise key over password in SSHClientConnection

stuarthendren:
From ticket:
If keyboard interactive login is not allowed on the box but the user also requires a sudo password the ssh fails as it prioritises the password.
If you remove the password then the sudo fails in the SudoAwareInitManager.
It would seem better to prioritise the key over the password in SSHClientConnection or possibly try both if they are both present, and the first fails.

This commit swaps the order of the if else check to use the ssh key if present.

### DIFF
--- a/drivers/sshj/src/main/java/org/jclouds/sshj/SSHClientConnection.java
+++ b/drivers/sshj/src/main/java/org/jclouds/sshj/SSHClientConnection.java
@@ -157,12 +157,12 @@ public class SSHClientConnection implements Connection<SSHClient> {
          ssh.setTimeout(sessionTimeout);
       }
       ssh.connect(hostAndPort.getHostText(), hostAndPort.getPortOrDefault(22));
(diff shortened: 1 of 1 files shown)


nacx:
The approach looks good to me. Thanks @stuarthendren!
Looking at how the [jsch driver](https://github.com/jclouds/jclouds/blob/master/drivers/jsch/src/main/java/org/jclouds/ssh/jsch/SessionConnection.java#L172-L180) manages the credentials I'd say it will already work. Anyway, could you try using it, just to make sure all the ssh drivers are fixed in this issue?

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/f6a97139) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/f88acd67). I've also added a commit after yours to align the Jsch driver, which was required after merging https://github.com/jclouds/jclouds/pull/735. Thanks @stuarthendren!

PR #765, opened 2015-06-10
fix range get for manifest object

kahing:
swift doesn't return the etag if the client is doing a range
get of a manifest (SLO or DLO), this fixes SwiftObject so
etag is optional

### DIFF
--- a/blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
+++ b/blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
@@ -426,6 +426,32 @@ public class BaseBlobIntegrationTest extends BaseBlobStoreIntegrationTest {
       }
    }
 
+   @Test(groups = { "integration", "live" })
+   public void testGetRangeMultipart() throws InterruptedException, IOException {
+      String container = getContainerName();
+      InputStream expect = null;
+      InputStream actual = null;
+      try {
+         String name = "apples";
+         long length = getMinimumMultipartBlobSize();
+         ByteSource byteSource = TestUtils.randomByteSource().slice(0, length);
+         Blob blob = view.getBlobStore().blobBuilder(name)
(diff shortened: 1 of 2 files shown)


blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
```diff
@@ -427,6 +427,25 @@ public void testGetTwoRanges() throws InterruptedException, IOException {
...
+         String name = "apples";
+         long length = getMinimumMultipartBlobSize();
+         ByteSource byteSource = TestUtils.randomByteSource().slice(0, length);
+         Blob blob = view.getBlobStore().blobBuilder(name)
+                 .payload(byteSource)
+                 .contentLength(length)
+                 .build();
+         view.getBlobStore().putBlob(container, blob, new PutOptions().multipart(true));
+         blob = view.getBlobStore().getBlob(container, name, range(0, 5));
+         validateMetadata(blob.getMetadata(), container, name);
```
andrewgaul: Compare the `blob` payload against the partial `byteSource`?

kahing: done

blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
```diff
@@ -427,6 +427,27 @@ public void testGetTwoRanges() throws InterruptedException, IOException {
...
+         assertThat(blob.getPayload().openStream())
+                 .hasContentEqualTo(byteSource.slice(0, 6).openStream());
```
andrewgaul: Wrap these in try-with-resources to close the socket?

kahing: done

PR #767, opened 2015-06-10
use awaitConsistency in more places so provider can override it

kahing:


andrewgaul:
Pushed to master as 23ef39416d4bd9ef43f50fe16dc4d4cf909afbc4.

PR #768, opened 2015-06-10
Avoid AutoValue shaded imports

nacx:
Google Auto provides some shaded dependencies, such as Guava, that should not be directly imported. This pull request sets a checkstyle rule to make sure we don't add them by mistake.


zack-shoylev:
+1

Great too, these are easy to miss.

PR #770, opened 2015-06-11
use segment names similar to those created by python-swiftclient

kahing:
Currently swift creates blob-1, blob-2... blob-n for multipart
upload segments, which are very common names that can easily
collide with normal objects. This changes the naming convention
to that of python-swiftclient (the `swift' command line).

python-swiftclient also uses a different segment container by
default, that's not implemented by this patch

### DIFF
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
@@ -27,7 +27,6 @@ import static org.jclouds.openstack.swift.v1.options.PutOptions.Builder.metadata
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.UUID;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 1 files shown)


apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -410,7 +410,16 @@ public long countBlobs(String containerName) {
 
    @Override
    public MultipartUpload initiateMultipartUpload(String container, BlobMetadata blobMetadata) {
-      String uploadId = UUID.randomUUID().toString();
```
andrewgaul: Call `initiateMultipartUpload(container, blobMetadata, 0)`?

andrewgaul:
@kahing Overall agree with the approach.

@zack-shoylev What do you think?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
kahing: done

andrewgaul:
Pushed to master as b8b20c16f5e7d0248e7611125b10c4ec45895910.

PR #772, opened 2015-06-11
send HEAD instead of GET for blobMetadata

kahing:


andrewgaul:
Good catch!  Pushed to master as 320742c6b5ad8258ddf5ed621a2ad4cb556a49ba and 1.9.x as 59291175968e044ef3aabf69f924d585ebd971de.

PR #773, opened 2015-06-12
Handle lowercase user metadata headers

andrewgaul:
HTTP headers are not case-sensitive.


andrewgaul:
@kahing @zack-shoylev Please review.

kahing:
:+1:

nacx:
Things like this one are better secured with a test :)

zack-shoylev:
Maybe I am missing something. Shouldn't we be looking specifically for "X-Object-Meta-" in the beginning of the header key?

andrewgaul:
@nacx Not sure the best way to address this -- servers can send any combination of cases.  I encountered jclouds case sensitivity in bouncestorage/swiftproxy#36 but generally https://github.com/bouncestorage/chaos-http-proxy will uncover these kinds of errors.  Some kind of taint or typing seems like a better approach though.

@zack-shoylev Agree that this code is wonky...

andrewgaul:
The original motivation for this pull request has disappeared so I am closing this.  Tightening the check as per @zack-shoylev suggestion requires more work as some code like `getKnowingServerMessesWithMetadataKeyCaseFormat` uses `EntriesWithoutMetaPrefix` to parse headers starting with `X-Account-Meta-`.

PR #774, opened 2015-06-15
Improve image check in the BaseImageExtensionLiveTest

nacx:
Do not perform an image list when we have the imageId. This behavior was also causing some failures in the image extension live tests due to [JCLOUDS-512](https://issues.apache.org/jira/browse/JCLOUDS-512), even if the extension is properly implemented. Appropriate tests to validate the issue must be added when it is fixed, but that should not prevent all existing image extension live tests from working.


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/8182f8c4) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/6b3aed3c).

PR #775, opened 2015-06-15
Use a configurable timeout for awaitConsistency

andrewgaul:
This allows setting to zero for strongly-consistency implementations
of s3 and swift stores.


andrewgaul:
@zack-shoylev @kahing @timuralp Please review.

zack-shoylev:
Should the pom setting be set for other apis/providers with this PR?
Otherwise, +1

andrewgaul:
@zack-shoylev Other providers use the default of 10 seconds which matches the previous behavior.

PR #779, opened 2015-06-19
Add missing value in Content-Range header

lmivan:
This is related with #776. It's necessary to add
the 'bytes' string before the bytes.


lmivan:
This is related with this issue https://github.com/andrewgaul/s3proxy/issues/63

andrewgaul:
Thank you for your contribution @lmivan!

lmivan:
Thanks! :+1:

PR #780, opened 2015-06-21
JCLOUDS-941: Auto-detect filesystem Content-Type

lmivan:
Auto-detect the content-type if it's null.

I think the feature is working in my first commit (135b609). 

In the second one I tried to add the property @andrewgaul suggested in the JIRA but I'm not really sure if this is the right way to do it. Besides, if I run the tests only for the class they all pass but when I execute the `install` target some tests fail because this Guice error that I don't know how to fix it:

```
com.google.inject.CreationException: Guice creation errors:

1) No implementation for java.lang.Boolean annotated with @com.google.inject.name.Named(value=jclouds.filesystem.auto-detect-content-type) was bound.
  while locating java.lang.Boolean annotated with @com.google.inject.name.Named(value=jclouds.filesystem.auto-detect-content-type)
    for parameter 2 at org.jclouds.filesystem.strategy.internal.FilesystemStorageStrategyImpl.<init>(FilesystemStorageStrategyImpl.java:121)
  at org.jclouds.filesystem.config.FilesystemBlobStoreContextModule.configure(FilesystemBlobStoreContextModule.java:43)
```

@andrewgaul I've been looking how to inject a property from a property file but I don't know how to do it. Could you please point me to an example or give me a hint how to do it?

Regards, Iván.


nacx:
If the property is not annotated as `@Nullable` Guice will fail to create the injector when missing. 

You should assign a default value in the FileSystemApiMetadata class by overriding the `defaultProperties` method

nacx:
To complete the answer, users will be able to override the value when creating the context, and you'll be able to provide a custom value when running the tests, if needed,  passing the value to the command line as a system property "-D....".

lmivan:
Thank you @nacx! I fixed it and now it is working!

I've built my own version of s3proxy with my compile jclouds and it detects correctly the content-type and plays the video in the browser.

I've done two commits, do you prefer that I ammed the, in only one?

andrewgaul:
Squashed commits, reworded commit message, changed `Boolean` to `boolean`, and pushed to master as 37a014ae00788fd8402f444dbfb93c364c9f1a4d.  Thank you for your contribution @lmivan!

lmivan:
Great! Thank you very much! :smile:

PR #782, opened 2015-06-24
list MPU parts with prefix listing instead of marker

kahing:
we now use segment conventions that allow us to do more
efficient listing


andrewgaul:
Corrected whitespace and pushed to master as e0bafd4c10c20312d8117e193adbc915bbac5621.

PR #784, opened 2015-06-25
Fix template equals assertion to consider location scope

devcsrj:
The current assertion assumes that the default location and the built one is always in the same scope. 

See [PR145](https://github.com/jclouds/jclouds-labs/pull/145#issuecomment-115174692) for use case.

/cc @nacx


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/107c9d0d) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/17c51c35). Thanks @devcsrj!

PR #787, opened 2015-06-25
JCLOUDS-172: Promote Google Compute Engine

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-172

This promotes the Google Compute Engine provider.

* The `googlecloud` goes to the `common` folder.
* The `google-compute-engine` project goes to `providers`.

Also all pom.xml have been updated and the Maven coordinates configured accordingly.

This PR keeps the commit history and brings all the commits in the jclouds-labs-google project. The relevant ones are the last two, being the first one the merge commit that promotes the provider and the last one the one that fixes the pom.xml files.

/cc @danbroudy @zack-shoylev


nacx:
Merged to master.

danbroudy:
Thanks for all the work that went into this! I'm excited to see this go through!

grkvlt:
@nacx What happened to the 1.9.x branches! There is now no `google-compute-engine` if you build a 1.9.1-SNAPSHOT release, and since you moved the whole commit history there is now no trace of it ever existing on `jclouds-labs` so it can't be rolled back there. I suggest copying the version from master here into the 1.9.x branch.

grkvlt:
@nacx Actually, I think you need to put the `google-compute-engine` back into jclouds-labs version 1.9.x and previous, it's just too confising to have the provider vanish. I should be able to build old versions on jclouds and you'll want to be able to add patches. Can you restore the commit history to the labs repo?

grkvlt:
@nacx Sorry, I broke the build all by myself. Ignore the previous comments...

nacx:
>Thanks for all the work that went into this! I'm excited to see this go through!

There is a huge amount of work and a great effort behind GCE. Kudos to you @danbroudy!

>Sorry, I broke the build all by myself. Ignore the previous comments...

NP! Early feedback is always welcome! :)

PR #789, opened 2015-06-26
JCLOUDS-931: Sleep only after creating containers.

timuralp:
jclouds should issue the requests to create all of the containers in
the container pool before sleeping. The patch modifies the
awaitConsistency method to be cognizant of whether the container pool
has been initialized or not. If the pool has not been initialized, the
method returns immediately and does not sleep.


blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobStoreIntegrationTest.java
```diff
@@ -531,6 +532,9 @@ protected Module createHttpModule() {
    }
 
    protected void awaitConsistency() {
+      if (!initialized) {
```
andrewgaul: This seems magical.  Instead should we pass a flag to `createContainerAndEnsureEmpty` which controls the call to `awaitConsistency`?

timuralp: Done

andrewgaul:
Pushed to master as e63474a8242aabab38785567d53467dd825d956f and 1.9.x as a05f369807a75c16f68837fbe00d03b7dbdcb934.

PR #791, opened 2015-06-30
JCLOUDS-930: Regression in a File System test.

timuralp:
Fix a regression in the file system blob store, where a test does not
expect the trailing delimiter character.

### DIFF
--- a/apis/filesystem/src/test/java/org/jclouds/filesystem/FilesystemBlobStoreTest.java
+++ b/apis/filesystem/src/test/java/org/jclouds/filesystem/FilesystemBlobStoreTest.java
@@ -525,8 +525,7 @@ public class FilesystemBlobStoreTest {
         checkForContainerContent(CONTAINER_NAME, dirs);
     }
 
-    // TODO: disable test due to unexpected trailing slash
-    @Test(dataProvider = "ignoreOnMacOSX", enabled = false)
+    @Test(dataProvider = "ignoreOnMacOSX")
     public void testListDirectoryBlobsS3FS() {
         blobStore.createContainerInLocation(null, CONTAINER_NAME);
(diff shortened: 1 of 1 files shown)


andrewgaul:
rebuild please

nacx:
@timuralp It seems there are also regressions in [jclouds-labs/jdbc](https://jclouds.ci.cloudbees.com/view/public/job/jclouds-labs/1028/org.apache.jclouds.labs$jdbc/testReport/). Mind fixing it too?

andrewgaul:
Timur and I are out at a conference all day, @rcoedo could you investigate these failing tests in JDBC blobstore?

timuralp:
@nacx I will try to look at the failures tonight unless anyone gets to it before I do.

timuralp:
Also, looking at the failures, this is identical to the test that was fixed for the file system blobstore (the test expected common prefixes to not include the delimiter -- the one significant change from the JCLOUDS-930 pull request).

nacx:
Yes, I fixed thay locally but then a couple integration test failures appeared in the `testDirectory` method that escape to my knowledge right now.

I preferred to comment it here so it can have a proper fix. Thanks for having a look at this!

timuralp:
@nacx I submitted a pull request for the BaseJdbcBlobStoreTest.

I found a two failing tests in the EclipselinkH2ContainerIntegrationTest and HibernateHsqldbContainerIntegrationTest will work on fixing those up as well.

rcoedo:
I can take a look at this later today.

timuralp:
@rcoedo @nacx figured out the integration tests failure in jdbc. It's related to the integration tests failure for the Transient blob store. I'm working on that now.

PR #792, opened 2015-06-30
Allow other version strings with arbitrary labels

grkvlt:
This allows building private forks with non-conflicting versions, since the current regular expression matching sematic versions is very restrictive. Versions such as `1.9.1-org-name.1` (for the first build of version 1.9.1 by some organisation) should be allowed. All of the checks for alpha or beta and the major, minor and patch properties in `JcloudsVersion` will still work as expected.

### DIFF
--- a/core/src/main/java/org/jclouds/JcloudsVersion.java
+++ b/core/src/main/java/org/jclouds/JcloudsVersion.java
@@ -37,11 +37,11 @@ public class JcloudsVersion {
     private static final String VERSION_PROPERTY_NAME = "version";
 
     /*
-     * x.y.z or x.y.z-alpha.n or x.y.z-beta.n or x.y.z-rc.n or x.y.z-SNAPSHOT -
-     * see http://semver.org.
+     * Matches x.y.z or x.y.z-alpha.n or x.y.z-beta.n or x.y.z-rc.n or x.y.z-SNAPSHOT - see http://semver.org/
+     * Also matches x.y.z-* as fallback, with any non-whitespace character sequence after the hyphen.
      */
     private static final Pattern SEMANTIC_VERSION_PATTERN =
-        Pattern.compile("(\\d+)\\.(\\d+)\\.(\\d+)(?:-(alpha|beta|rc)\\.(\\d+)|-SNAPSHOT)?");
+        Pattern.compile("(\\d+)\\.(\\d+)\\.(\\d+)(?:-(alpha|beta|rc)\\.(\\d+)|-SNAPSHOT|-\\S+)?");
     private static final String ALPHA_VERSION_IDENTIFIER = "alpha";
     private static final String BETA_VERSION_IDENTIFIER = "beta";
 


grkvlt:
Forgot to adjust tests in `JcloudsVersionTest` to accept new strings. But, since I'm not sure this is the right solution, having tried writing a new regexp that parses semver metadata using `+` syntax, it introduces too much complexity, so closing for now.

nacx:
Instead of hacking the version, I think the right approach for forks would be to change the groupId. If at some point other software ends up depending on a fork that has more features (or features not yet merged in jclouds) it is *technically* not depending on jclouds, so I think the right approach would be to change the groupIds. WDYT?

demobox:
> I think the right approach for forks would be to change the groupId

Agreed. jclouds tries to stick to semver, and if private users want to use a different version scheme, I'd say that it should be easy enough for them to patch `JcloudsVersion` as needed? Not sure why that change would need to make it back to jclouds itself?

nacx:
Also, take into account that if at some point the forked version needs to be published, say in Maven Central, then having a groupId different than the jclouds one is a must, so I'd say using a custom groupId is the easiest way to go for both jclouds and its forks.

grkvlt:
@nacx @demobox You're right, and we actually also do that. We would like a predictable set of dependencies, so we end up making effectively a pre-release version of jclouds that can be used as a fixed dependency, with a different groupId. It works fine if we have a single release made for that jclouds version, but because the release cycles of Brooklyn and Clocker are different from that of jclouds, we may want to depend on the code present in a -SNAPSHOT version at different stages. So changing the groupId is not ideal, since we don't want to encode the patch level there, it really belongs in the version, maybe as `+` separated metadata, as suggested by the sematic versioning spec.

Basically we are looking for the equivalent of nightly version numbers and builds, or something similar if not as frequent, fixed to a particular tag or commit id on the -SNAPSHOT development branch. Is that possible?

demobox:
> Is that possible?

@grkvlt I think nightly builds sound like a pretty good idea, in general, assuming we can get the release process sufficiently automated, of course. I'm not sure what the ASF thoughts would be on where to publish those, though, since I doubt the ASF release repo is the right place for that (@abayer: any thoughts there)?

Irrespective of that discussion, relaxing the regex to allow for other SemVer-compliant strings doesn't sound unreasonable to me. Still would like to know what the reason for having this upstream would be: are you finding that cherry-picking an appropriate patch to `JcloudsVersion` is too brittle?

Or is there a different problem? Or more because the idea of patchiing `JcloudsVersion` just feels nasty?

nacx:
> I'm not sure what the ASF thoughts would be on where to publish those, though, since I doubt the ASF release repo is the right place for that

Interesting. There is a recent discussion about this [in the general@ list](http://markmail.org/message/qtjqmzvgsg7w6j5s). But Take into account that nightly builds can't be considered a release or something we should encourage to use. There has not been any vote on those artifacts and although the state of the branch should be always releaseable, there has been no explicit check/validation, etc.

Regarding the version thing, what about changing the way artifacts are built to add a classifier? Would that be an option?

PR #796, opened 2015-07-02
JCLOUDS-930: Transient store integration test fixes

timuralp:
Two fixes to integration tests. The first one changes the tests to expect the delimiter to be included in the common prefixes. The second change is to fix directory handling in the local blob store and a unit test to catch the regressions.


nacx:
This seems to fix the build. Merged to master. Thanks @timuralp!

PR #800, opened 2015-07-07
Swift: do not assume only 1 container in tests.

timuralp:
Openstack Swift ContainerApiLiveTest testListWithOptions assumes that
there are no other containers and that ContainerApiLiveTest container
will _always_ be the first container.

The patch changes the behavior of the test to instead pull out the
container from the returned list.


apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ContainerApiLiveTest.java
```diff
@@ -98,7 +99,13 @@ public void testListWithOptions() throws Exception {
...
          ListContainerOptions options = ListContainerOptions.Builder.marker(lexicographicallyBeforeName);
-         Container container = api.getContainerApi(regionId).list(options).get(0);
+         Container container = api.getContainerApi(regionId).list(options).firstMatch(new Predicate<Container>() {
+            @Override
+            public boolean apply(Container container) {
+               return container.getName().equals(name);
+            }
+         }).get();
+         assertNotNull(container);
          assertEquals(container.getName(), name);
```
andrewgaul: This is now always true.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ContainerApiLiveTest.java
```diff
@@ -98,7 +99,13 @@ public void testListWithOptions() throws Exception {
...
       for (String regionId : regions) {
          ListContainerOptions options = ListContainerOptions.Builder.marker(lexicographicallyBeforeName);
-         Container container = api.getContainerApi(regionId).list(options).get(0);
+         Container container = api.getContainerApi(regionId).list(options).firstMatch(new Predicate<Container>() {
+            @Override
+            public boolean apply(Container container) {
+               return container.getName().equals(name);
+            }
+         }).get();
+         assertNotNull(container);
```
andrewgaul: Needed?  `get` will throw `IllegalStateException` if not present.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ContainerApiLiveTest.java
timuralp: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ContainerApiLiveTest.java
timuralp: Done

andrewgaul:
Pushed to master as 86af0753bf51bba42d5c18657e1f0e3dd72630ee.

PR #801, opened 2015-07-07
JCLOUDS-930: Handle subdir entries correctly.

timuralp:
When listing blobs, Swift returns an array of "subdir" entries for
every common prefix. The openstack-swift provider needs to process
such entries and add them to the result set. The subdir array is an
array of strings, so additional metadata needs to be added to the
objects (bogus ETag, bogus LastModifiedDate, and so on).

When directory marker blobs are used, this means that potential _two_
entries are generated for every directory if:
1. the delimiter is set and matches the directory blob (e.g. dir/ and
   delimiter /)
2. there are objects under the directory name (e.g. dir/blob), which
   will result in results that include common prefixes (subdir)

In the above example, we should expect two results: "dir" and "dir/"
representing the directory marker blob and the common prefix,
respectively.

This is caught in the testDirectory integration test.

The patch changes the behavior of the Swift provider to correctly
handle the results in the subdir stanza and changes the test to expect
the directory marker to be returned in the list.


timuralp:
The Azure PR will fail in integration tests if merged before this commit (as the change to the test is in this commit).

blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseContainerIntegrationTest.java
```diff
@@ -234,9 +234,9 @@ public void testDirectory() throws InterruptedException {
          awaitConsistency();
 
          container = view.getBlobStore().list(containerName);
-         // we should still have only the directory under root
+         // we should get back the subdir entry and the directory marker
          assert container.getNextMarker() == null;
-         assert container.size() == 1 : container;
+         assertThat(container.size()).isEqualTo(2);
```
andrewgaul: `assertThat(container).hasSize(2)`

timuralp: Done

andrewgaul:
Pushed to master as 6d27fbb18ac682df4579c98b7c3f3fefe8ca11e6.

PR #802, opened 2015-07-07
Fix filtering of directories in the local blob store

timuralp:
The local blob store should not explicitly filter out directories. The changed behavior aligns the Transient blob store with Swift. However, the file system blob store does not have a notion of marker blobs and the second commit disables the directory test for the filesystem blob store.


andrewgaul:
I see a failure:

```
  TransientContainerIntegrationTest>BaseContainerIntegrationTest.testDirectory:239 [[type=RELATIVE_PATH, id=null, name=directory, location=null, uri=null, userMetadata={}], [type=RELATIVE_PATH, id=null, name=directory/, location=null, uri=null, userMetadata={}]]
```

timuralp:
Sorry, I need to base this PR off the Swift subdirectory change as that's where the change to the ```testDirectory``` test happened. I can submit a different PR based off that branch.

andrewgaul:
Pushed to master as 669feb4d0c459b1612818869ce82b9b5d1915f42 and 5383148d9ef92fb39d8958abbe9cae19c80ca609.

PR #804, opened 2015-07-09
Removes CopyObjectException in favor of standard blobstore exceptions

zack-shoylev:


zack-shoylev:
https://issues.apache.org/jira/browse/JCLOUDS-957

zack-shoylev:
@andrewgaul ^

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ObjectApiLiveTest.java
```diff
@@ -141,20 +142,17 @@ public void testCopyObject() throws Exception {
          // test exception thrown on bad source name
          try {
             destApi.copy(destinationObject, badSource, sourceObjectName);
-            fail("Expected CopyObjectException");
```
andrewgaul: Can you use try .. catch instead of `expectedExceptions`?  This narrows the exception handling scope:

http://konigsberg.blogspot.com/2007/11/testng-and-expectedexceptions-ive.html

(also the for loop above will not complete with this change)

andrewgaul:
:+1:

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ObjectApiLiveTest.java
zack-shoylev: Ah, right. Thanks, I will look into that.

zack-shoylev:
Rebuild, please.

zack-shoylev:
merged but not backported

PR #805, opened 2015-07-09
Use unique yet deterministic container names

andrewgaul:
Previously jclouds allocated the initial container pool as
$USERNAME-blobstore-1, 2, etc. and a different system for scratch
containers.  The former caused errors with azureblob during multiple
runs of integration tests since containers do not recycle immediately.
We now allocate all container names as jclouds-$USERNAME-$TIMESTAMP-1,
2, etc.


andrewgaul:
The previous approach has the advantage of removing initial pool buckets from previous runs if they were interrupted.

timuralp:
@andrewgaul each run could remove all buckets that begin with ```CONTAINER_PREFIX```? This seems like the right thing to do.

PR #809, opened 2015-07-14
Updates live tests and template builder settings for compute

zack-shoylev:


zack-shoylev:
This will have to be backported

zack-shoylev:
merged and backported

PR #810, opened 2015-07-14
Move the delimiter/prefix tests to BaseContainerIntegrationTest

timuralp:
BaseContainerLiveTest is not run in every provider (notably file system API). The commit moves the test to BaseContainerIntegrationTest and updates the affected providers accordingly. We also shouldn't use the file separator character when testing the file system provider, as it will generate directories for every blob name of the form "foo/bar". This presents a challenge, since when we list the container, we will observe the directory entries (unlike in other providers). Using a "-" works around this issue.

The first commit adds an additional test for the case of two blobs "bar" and "bazar" with prefix set to "ba" and delimiter to "a". The test verifies the delimiter behavior in the case of a prefix ending with the delimiter character.


timuralp:
Unfortunately, I need to squash the two PRs that amend the tests and the local blob store together. Closing this and opening a new one for all three commits together.

PR #811, opened 2015-07-14
Propagate delimiter changes to LocalBlobStore.

timuralp:
Propagates the changes to add delimiter support to the local blob store. The prefix option no longer works with the non-recursive option. The caller _must_ set the delimiter option to enable the same functionality.


timuralp:
Unfortunately, I'm gonna have to squash this together with the PR that amends the tests.

PR #812, opened 2015-07-14
Handle delimiter and prefix being set in LocalBlobStore

timuralp:
If the prefix is set, we need to also handle the delimiter option. One commit plumbs this through. Another commit adds a test for the scenario. The last commit moves the ```testContainerListWithPrefix``` and ```testDelimiterList``` tests from ```BaseContainerLiveTest``` to ```BaseContainerIntegrationTest```. The ```BaseContainerLiveTest``` is not run for the file system API and regressions, such as the one above occur.

Lastly, the delimiter tests are changed from using a (possible) file separator character to use "-", as creating a blob with a file separator character in its name will cause jclouds to create a directory, which is visible when listing blobs.


andrewgaul:
Pushed to master as 613b6bf045eb0ed040e1202c92bdb3644c1276c9, e4161a59bbfd84e834081b0efe85d5b938798a86, and 19d54369db3ac20ad0314339fd6255b1c4f39e5f.

PR #814, opened 2015-07-15
range get of manifest object has no etag

kahing:


andrewgaul:
Pushed to master as ad38e482058fd6a5c9463363ad4eae06874d3321.

PR #815, opened 2015-07-15
 Split availability Zone API to return Basic Info and Detailed info 

arvindn05:
Split availability Zone API to return Basic Info and Detailed info.
My previous checkin did not show detail for availability zone which is a different path than just plain availability zone info(name and state)

Introduced AvailabilityZone class which just shows name and state
Introduced AvailabilityZoneDetail class which show name, state and host information

updated tests for same


apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/domain/regionscoped/AvailabilityZoneDetails.java
```diff
@@ -0,0 +1,116 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.gson.annotations.SerializedName;
+
+import java.beans.ConstructorProperties;
+import java.util.Date;
+import java.util.Map;
+
+public class AvailabilityZoneDetails {
+
+   public static final class HostService {
```
zack-shoylev: At this point, maybe move this to its own file.

arvindn05: just curious, Host service is only valid within AvailabilityZoneDetails why move it out to its own file?
wouldnt that pollute the namespace if someone introduces an independent host service class?

zack-shoylev: Well, yes, there will be some namespace pollution, but it usually is not a problem with jclouds. It helps that the namespaces are fairly well organized. It's a trade-off, but I think in this case we might get better readability. Your call, though.

apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/extensions/AvailabilityZoneApiExpectTest.java
```diff
@@ -34,7 +36,7 @@
 @Test(groups = "unit", testName = "AvailabilityZoneApiExpectTest")
 public class AvailabilityZoneApiExpectTest extends BaseNovaApiExpectTest {
 
-   public void testAvailabilityZonesList() throws Exception {
+   public void testListAvailabilityZones() throws Exception {
```
zack-shoylev: Both of these tests need fail tests as well.

zack-shoylev:
Can we add some of your first comment in this PR as in-code docs for the domain classes?

arvindn05:
Thanks for the review. @zack-shoylev Added few more negetive test cases and comments.

Leaving the host service as is for now. If there are other domain classes which need to use host service domain class, we can move it out.

zack-shoylev:
Going to try rebuilding...

zack-shoylev:
merged, backported

PR #816, opened 2015-07-15
JCLOUDS-462: Fix a bunch of AWS live tests

nacx:
This PR fixes most of the remaining live test failures in AWS. After having discussed the remaining ones with @abayer, we've agreed that they are transient ones or specific to particular regions, and are not blockers.


nacx:
rebuild please

abayer:
+1, cherry picking it around myself now.

PR #818, opened 2015-07-16
JCLOUDS-962: Do not override provider methods

olivergondza:
[JCLOUDS-962](https://issues.apache.org/jira/browse/JCLOUDS-962)

Fix guice 4 support.

### DIFF
--- a/apis/atmos/src/main/java/org/jclouds/atmos/blobstore/config/AtmosBlobStoreContextModule.java
+++ b/apis/atmos/src/main/java/org/jclouds/atmos/blobstore/config/AtmosBlobStoreContextModule.java
@@ -45,7 +45,7 @@ public class AtmosBlobStoreContextModule extends AbstractModule {
 
    @Provides
    @Singleton
-   protected LoadingCache<String, Boolean> isPublic(final AtmosClient client) {
+   protected final LoadingCache<String, Boolean> isPublic(final AtmosClient client) {
       return CacheBuilder.newBuilder().expireAfterWrite(30, TimeUnit.SECONDS).build(new CacheLoader<String, Boolean>() {
 
          @Override
--- a/apis/atmos/src/main/java/org/jclouds/atmos/config/AtmosHttpApiModule.java
+++ b/apis/atmos/src/main/java/org/jclouds/atmos/config/AtmosHttpApiModule.java
@@ -56,6 +56,10 @@ public class AtmosHttpApiModule extends HttpApiModule<AtmosClient> {
 
    @Provides
    @TimeStamp
+   protected final String guiceProvideTimeStamp(@TimeStamp Supplier<String> cache) {
+      return provideTimeStamp(cache);
+   }
+
    protected String provideTimeStamp(@TimeStamp Supplier<String> cache) {
       return cache.get();
    }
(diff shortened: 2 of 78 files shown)


nacx:
Thanks for looking at this @olivergondza!

Looking at the change, I'd suggest a different approach: what about leaving the provider annotations in the parent class, make the method `final` to make sure it is not overridden my mistake, and make it call another protected method that actually does the work? This way we can leave the guice configuration to the parent class and keep the current behavior. This should be less error-prone regarding Guice misconfigurations. WDYT?

nacx:
Looking a bit deeper, and to make sure we don't allow mistakes in Guice 4, I think the right approach would be to make **all** provider methods in the code base `final`, even if they are still not subclassed, to avoid failures in the future.

Running the following command will help identifying all provider methods defined in the module classes:

```bash
find . -name '*.java' -exec grep '@Provides' -nH -A3 {} \; | grep -i module
```

WDYT?

nacx:
Also, I'll setup a Jenkins build that uses Guice 4 so we can use it to track the status of this issue.

nacx:
rebuild please

olivergondza:
@nacx, Looking back at the amount of boilerplate this generates, keeping the `@Provider` annotation in parent class is a must. Refactoring:

```java
@Provide
public X provideX() { ... }
```

into

```java
@Provide
public final X guiceProvideX() { ... }
public X provideX() { ... }
```
 
should not break clients and limit the scope of the change to parent classes.

nacx:
Agree.

I've already configured a build that uses Guice 4.0 to track this pull request: https://jclouds.ci.cloudbees.com/view/public/job/jclouds-guice-4/
Let's see if it is automatically triggered and reports feedback here: rebuild please

olivergondza:
Alright, I have marked all provider methods final to identify the overridden ones and created a delegate where needed. Still, there is a lot of methods that _was not_ overriden and thus have no delegate methods. I am undecided if we should keep them `final` to identify problems in dependent libraries or revert back to non-final not to break the clients.

This patch now works for me.

nacx:
Yep, build is passing: https://jclouds.ci.cloudbees.com/view/public/job/jclouds-guice-4/37/

Regarding the final methods, clients can expect to break when upgrading to 2.0.0; we just need to make sure there is a way for them to recover. It would be ideal to have all provider methods implement the pattern that allows inheritance, but I think that would introduce a lot of unnecessary boilerplate code, as most of those provider methods (if not all) are unlikely to be overridden, so I'd say it is ok to leave them final.

I'm ok with merging this as-is. It allows us to support Guice 4.0 now and will give users enough time to plan the migration and open any issues they find, but I'd like to have some feedback from the community too, so I'd leave the PR open a couple days and share this in the ML. Sounds good?

demobox:
Created a new matrix build to test this PR against Guice 3.0 and 4.0:

https://jclouds.ci.cloudbees.com/job/jclouds-guice-3-4/

Thanks, @olivergondza and @nacx!

olivergondza:
Would it be possible to publish alpha/rc with the announcement so people can start to experiment and provide more useful feedback? It is not a deal breaker but it blocks https://github.com/jenkinsci/openstack-cloud-plugin/pull/15, and the release would help.

nacx:
As soon as this gets merged the 2.0.0-SNAPSHOT will be automatically released to the Apache snapshots repo, so you'll be able to start testing it (see http://jclouds.apache.org/start/install/).

A formal release in Apache is a bit more complicated as the release has to be proposed and voted by the community first. If the snapshot approach is not an option I'd encourage you to start a guice4-friendly release preview discussion in the jclouds dev@ mailing list.

nacx:
@demobox good to merge?

nacx:
Squashed the fix commit and pushed to master. Thanks @olivergondza and apologies for the delay! I've also updated the Jenkins build to build the master branch with both versions of Guice.

PR #819, opened 2015-07-18
add os-hypervisors extension

arvindn05:
add os-hypervisors extension for openstack-nova
fixes https://issues.apache.org/jira/browse/JCLOUDS-851


apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/config/NovaHttpApiModule.java
```diff
@@ -96,6 +94,8 @@ private void bindDefaultAliases() {
             URI.create("http://docs.openstack.org/compute/ext/os-volume-attachment-update/api/v2"));
       aliases.addBinding(URI.create(ExtensionNamespaces.ATTACH_INTERFACES)).toInstance(
             URI.create("http://docs.openstack.org/compute/ext/interfaces/api/v1.1"));
+      aliases.addBinding(URI.create(ExtensionNamespaces.HYPERVISORS)).toInstance(
+            URI.create("http://docs.openstack.org/compute/ext/hypervisors/api/v1.1"));
```
arvindn05: no idea why we cant use ExtensionNamespaces.HYPERVISORS instead of specifying the URI again ...but since all the other binding do the same, i am following the same pattern FYI

arvindn05:
In NovaHttpApiModule,
no idea why we cant use ExtensionNamespaces.HYPERVISORS instead of specifying the URI again ...but since all the other binding do the same, i am following the same pattern FYI

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/domain/regionscoped/HypervisorDetails.java
```diff
@@ -0,0 +1,70 @@
...
+ * Hypervisor which shows id, name and other details
+ */
+@AutoValue
+public abstract class HypervisorDetails {
+
+   public abstract String getId();
+
+   public abstract String getName();
+
+   public abstract int getCurrentWorkload();
```
zack-shoylev: Are you sure all of these ints will be returned by the service? If not, use Integer.

apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/extensions/HypervisorApiExpectTest.java
```diff
@@ -0,0 +1,163 @@
...
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+
+@Test(groups = "unit", testName = "HypervisorApiExpectTest")
+public class HypervisorApiExpectTest extends BaseNovaApiExpectTest {
+
+   public void testWhenNamespaceInExtensionsListHypervisorPresent() throws Exception {
```
zack-shoylev: Even though nova uses expect tests, maybe new tests should be using the mock test approach in place of expect tests.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/domain/regionscoped/HypervisorDetails.java
arvindn05: good point.
Took the reference from here. added nullable annotation and made fields objects for nullable properties.
https://github.com/openstack/nova/blob/master/nova/objects/compute_node.py

apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/extensions/HypervisorApiExpectTest.java
arvindn05: I am not sure about how i would go about introducing mock tests. It might require more changes like introducing a base mock test to be more comprehensive. 
Since there are lot of expect tests here might make sense to do a refactor at a later time as one change

zack-shoylev: If I remember correctly, the nova base mock test (or something like that) will inherit from the base openstack mock test. The openstack base mock test should implement all the tools you need.
In addition to that, I would prefer that we stop accumulating more crud if we can avoid it. The other tests will have to be refactored anyways eventually.

arvindn05:
```Java    
  MockWebServer server = mockOpenStackServer();
      server.enqueue(addCommonHeaders(new MockResponse().setBody(stringFromResource("/keystoneAuthResponse.json"))));
      server.enqueue(addCommonHeaders(new MockResponse().setBody(stringFromResource("/extension_list_full.json"))));
      try {
         NovaApi novaApi = api(server.getUrl("/").toString(), "openstack-nova");

         assertEquals(novaApi.getConfiguredRegions(), ImmutableSet.of("az-1.region-a.geo-1", "az-2.region-a.geo-1", "az-3.region-a.geo-1"));
         Optional<HypervisorApi> hypervisorApi = novaApi.getHypervisorApi("az-1.region-a.geo-1");
         assertTrue(hypervisorApi.isPresent());
         assertAuthentication(server);
         assertExtensions(server);

      } finally {
         server.shutdown();
      }
```
I am trying to write the mock test based on your suggestion. See Above.
I am getting stuck on the below line.
```Java
Optional<HypervisorApi> hypervisorApi = novaApi.getHypervisorApi("az-1.region-a.geo-1");
```
 The test just wait for a minute or so and give the following error
```Java
org.jclouds.http.HttpResponseException: Connection timed out: connect connecting to GET https://az-1.region-a.geo-1.compute.hpcloudsvc.com/v2/3456/extensions HTTP/1.1
	at org.jclouds.http.internal.BaseHttpCommandExecutorService.invoke(BaseHttpCommandExecutorService.java:117)
```

zack-shoylev: This happens when you are doing an additional request somewhere. The mock web server will not answer a request if a response for it is not queued. Maybe add an additional response and see where you are doing something you don't expect.

arvindn05: enqued a few more responses, but still get the same error. It looks like it is actually trying to communicate with https://az-1.region-a.geo-1.compute.hpcloudsvc.com/v2/3456/extensions...shouldnt the mock server intercept this request?

zack-shoylev: You might be missing a piece such as https://github.com/jclouds/jclouds-labs-openstack/blob/master/openstack-heat/src/test/java/org/jclouds/openstack/heat/v1/internal/BaseHeatApiMockTest.java

arvindn05: figured out the issue. It was because of me using keystoneAuthResponse.json. Somehow the mock tests only work with access.json

Also tried to make use of beforemethod and aftermethod annotations to do some of the setup. since the tests run in parallel, i could not use them directly. so had to get creative to reduce code duplication.
if you like this pattern maybe future tests can use it...

zack-shoylev: ?Hmm yes, I like the pattern actually. I want to run it through some tests, but looks good. If it works ok, we might have to refactor some other mock tests to use it as well.

________________________________

zack-shoylev:
merged and backported

PR #822, opened 2015-07-21
marker should be null when there are no more items

kahing:


blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseContainerIntegrationTest.java
```diff
@@ -187,6 +187,8 @@ public void testListContainerMarker() throws InterruptedException {
                container);
          assert container.getNextMarker() == null;
 
+         container = view.getBlobStore().list(containerName, afterMarker("z"));
+         assert container.getNextMarker() == null;
```
andrewgaul: Could you write this as `assertThat(container.getNextMarker()).isNotNull()` for a more informative error message?

andrewgaul:
:+1:

@zack-shoylev any comments?

blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseContainerIntegrationTest.java
kahing: done

PR #824, opened 2015-07-27
Do not buffer range get in LocalBlobStore

andrewgaul:
This avoids `OutOfMemoryError` with large blobs and resolves issues with
ranges exceeding array limits.


andrewgaul:
@kahing @timuralp Please review.

timuralp:
Looks good to me. @andrewgaul is there a way to add a test for this behavior?

andrewgaul:
It would be hard to add this to the transient unit tests since we would have to make assumptions about the Java heap.  We could definitely do this in the filesytem integration tests if the environment provided sufficient temporary storage.  Probably not worth the effort though.

PR #826, opened 2015-07-27
JCLOUDS-972. Fix parsing of spot reqs, set sane default for validUntil.

abayer:
Get the faultCode and faultMessage to actually be parsed (though I'm
not sure they're ever used), add statusCode, statusMessage and
statusUpdateTime, and have AWSEC2TemplateOptions default to a sane 30
minute lifetime for spot instance requests, so they don't get orphaned
forever if the price is too low etc.


nacx:
Looke like an unrelated failure. +1.
Thanks @abayer!

PR #828, opened 2015-07-27
JCLOUDS-967: Chef Client Model update with public key attribute

adrianbn:
Requests to chef's API endpoint organizations/<orgname>/clients/<clientname> return the client's public key but the Client model has no attribute to store it. This PR adds the public key attribute to the client model and updates the Client class accordingly.


apis/chef/src/main/java/org/jclouds/chef/domain/Client.java
```diff
@@ -73,27 +80,34 @@ public Builder isValidator(boolean validator) {
...
    private final String orgname;
    private final String clientname;
    private final String name;
    private final boolean validator;
 
-   @ConstructorProperties({ "certificate", "orgname", "clientname", "name", "validator", "private_key" })
+   @ConstructorProperties({ "certificate", "orgname", "clientname", "name", "validator", "private_key", "public_key"})
    protected Client(X509Certificate certificate, String orgname, String clientname, String name, boolean validator,
-         @Nullable PrivateKey privateKey) {
+         @Nullable PrivateKey privateKey, PublicKey publicKey) {
```
nacx: If the public key is not returned in Chef Server versions < 12 it should be marked as nullable. Otherwise the deserialization of an object missing the public key will fail.

If this is the case, then could you add a test to verify this? In the test class you updated, keep a test for a client without the public key and one for a client that has it set.

nacx:
Thanks @adrianbn! Just one comment; otherwise looks good.

nacx:
Thanks @adrianbn! lgtm.
I'll merge this PR in a few days, when 1.9.1 is released (I hope this to be by the end of the week) and the 1.9.x branch is unlocked so I can merge it to master and 1.9.x at once.
Mind squashing the commits into a single one meanwhile?

adrianbn:
Done! Let me know if you need anything else. Thanks!

nacx:
Jenkins failure was unrelated.
Pushed to master as [eafdeb52](http://git-wip-us.apache.org/repos/asf/jclouds/commit/eafdeb52). Thanks @adrianbn!

PR #829, opened 2015-07-28
Fixes an error caused by not using an assert by @CheckReturnValue.

zack-shoylev:
Also see jsr305


zack-shoylev:
https://issues.apache.org/jira/browse/JCLOUDS-975

zack-shoylev:
merged, backported

PR #830, opened 2015-07-29
Fix bug where password information is printed in logs in case of exce…

arvindn05:
Fix bug where password information is printed in logs in case of exceptions in keystone

fixes bug https://issues.apache.org/jira/browse/JCLOUDS-958


core/src/main/java/org/jclouds/http/HttpResponseException.java
```diff
@@ -83,8 +83,13 @@ static String requestPayloadIfStringOrFormIfNotReturnEmptyString(HttpRequest req
             && request.getPayload().getContentMetadata().getContentLength() != null
             && request.getPayload().getContentMetadata().getContentLength() < 1024) {
          try {
-            return String.format(" [%s] ", request.getPayload() instanceof StringPayload ? request.getPayload()
-                  .getRawContent() : Strings2.toStringAndClose(request.getPayload().openStream()));
+            String logStatement;
+            if (request.getPayload() instanceof StringPayload) {
+               logStatement = request.getPayload().isSensitive() ? "---- SENSITIVE INFORMATION ----" : request.getPayload().getRawContent().toString();
```
nacx: The sensitive flag should apply to all payload types, not just string ones.

nacx:
Thanks @arvindn05! The approach looks pretty clean.

You will also want to check the sensitive flag [in the Wire class](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/logging/internal/Wire.java#L120-L137), so the sensitive information does not get printed in the logs with the wire logs are enabled.

Regarding the change, this is something users might want to be able to turn on/off. I mean, this fix does not only mask the sensitive information, but removes the entire payload from the logs, and that might not be ideal/wanted by everyone. We should provide a way to disable this (I'm OK having it enabled by default), such a property one can configure when creating the context.

/cc @andrewgaul @demobox

core/src/main/java/org/jclouds/http/HttpResponseException.java
arvindn05: will make the change

zack-shoylev:
Looks good to merge/backport, is there any work left on this?

zack-shoylev:
Also I see this is against 1.9.x, can it be merged to master as well?

arvindn05:
I dont have any other work left on this.

@zack-shoylev we should be able to merge this to master. Is there anything i need to do for helping you merge this to master?

zack-shoylev:
I will merge and run tests, then backport

arvindn05:
thanks

zack-shoylev:
merged and backported

nacx:
I'm ok with the merge but not with the backport. In 1.9.x the defsult for the property should be set to true.

We can't just remove the payload logs in a bugfix release. There might be tooling, etc, that process the logs and we don't want to break any integration. We should provide the feature but default to the current behavior in 1.9.x.

arvindn05:
@nacx I am fine if you want the default behavior to be that you print out the password information in plaintext for 1.9.x

But since that will cause a security issue i would suggest to not do that. I understand your concern about tooling and tools which process logs but they should not be relying on security issue being present in jclouds. I think its good to break their tooling in this case.

maybe we can add a release note to say this issue has been fixed and might require updates to tooling.

nacx:
I'd agree if the fix just masked the sensitive part of the payload just like the "toString" does. But we are removing the entire payload.

Wouldn't a note in the release notes/release announcement saying the issue is fixed and it can be turned on by setting the property be enough for those worried about clear text sensitive payloads?

arvindn05:
we do add this message in its place instead of removing the payload "Sensitive data in payload, use PROPERTY_LOGGER_WIRE_LOG_SENSITIVE_INFO override to enable logging this data." so its going to be obvious i  hope.

like i said i am fine either way, though i prefer keeping the authentication payload out of the logs until explicitly requested with an ovveride.

zack-shoylev:
Hmm I thought that the backport should be fine because that is what the original PR targeted (it targeted the 1.9.x branch), which is why it was backported. Sorry for any confusion.

Instead of reverting the change, we should just merge a PR to 1.9.x specifically with the default targeted behavior that we want, but keep the option in.

Please link such a PR here so we can keep track.

As for the acceptable solutions, I really would prefer to err on the side of more security. I am OK with the code as is; but is there a way to filter just sensitive data from the payload? Like search-and-replace passwords? Or is the whole payload stream considered sensitive? 

I am not really sure what parts of the payload are considered sensitive, so let me know if I'm missing something.

arvindn05:
>>but is there a way to filter just sensitive data from the payload? Like search-and-replace passwords?
by the time we get the payload data, it has been converted into JSON and the only way to selectively mask the sensitive payload, we would have to go with a search and replace solution.

I dont like using search and replace because it will only handle one off cases and not be a generic solution.

Regardless, created a PR to change default logging behavior in 1.9.x 

see https://github.com/jclouds/jclouds/pull/834

PR #831, opened 2015-07-29
Fix a small softlayer bug that triggers a guava 19 rc bugcheck.

zack-shoylev:


nacx:
+1

zack-shoylev:
merged and backported

providers/softlayer/src/main/java/org/jclouds/softlayer/compute/strategy/SoftLayerComputeServiceAdapter.java
```diff
@@ -272,7 +272,7 @@ public int compare(Hardware h1, Hardware h2) {
                     .compare(getBootableDeviceType(h1), getBootableDeviceType(h2));
             if (!volumes1.isEmpty() && !volumes2.isEmpty() && volumes1.size() == volumes2.size()) {
```
andrewgaul: @andreaturli Outside the scope of this commit, but this can't be right -- two otherwise identical objects, one with and without a volume, will compare to zero instead of -1 or +1.

andreaturli: @andrewgaul thanks for the comment but I'm not sure I got your suggestion. Could you rephrase it, please?

andrewgaul: @andreaturli Consider

h1: cores=1, ram=256, etc., volumes = [volume(type=FOO)]
h2: cores=1, ram=256, etc., volumes = []

These will compare to 0 currently.  I expect them to compare to either -1 or +1 since they are different.

andreaturli: I see your point, I will fix it later! thanks!

PR #832, opened 2015-07-30
Add new Softlayer API method to return arbitrary VirtualGuest fields

grkvlt:
Adds new method to `VirtualGuestApi` that allows the user to set the object-mask, therefore enabling arbitrary fields to be returned; for example the `NetworkVlan` object. Includes unit and live tests.

### DIFF
--- a/providers/softlayer/src/main/java/org/jclouds/softlayer/features/VirtualGuestApi.java
+++ b/providers/softlayer/src/main/java/org/jclouds/softlayer/features/VirtualGuestApi.java
@@ -19,6 +19,7 @@ package org.jclouds.softlayer.features;
 import static org.jclouds.Fallbacks.FalseOnNotFoundOr404;
 import static org.jclouds.Fallbacks.NullOnNotFoundOr404;
 import static org.jclouds.Fallbacks.VoidOnNotFoundOr404;
+
 import java.util.Set;
 
 import javax.inject.Named;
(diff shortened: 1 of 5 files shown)


grkvlt:
@andreaturli Who should I ask to review this?

nacx:
Lgtm. Does it make sense to provide "filtered" methods for all methods in SoftLayer?

andreaturli:
thanks @grkvlt ! I think it makes sense to have such a method. Probably it should be the only mechanism supported *but* in order to do that we need to improve the domain objects (moving to AutoValue for sure) so that they actually support arbitrary fields.

Main problem is that DTO are quite complex in SoftLayer and they not only support `local properties` but also `relational and count properties` because of the rich query language they support from API.
There is a risk that the query passed using *Filtered method may be valid but requires a field that is not modelled in the domain class. See [VirtualGuest def](http://sldn.softlayer.com/reference/datatypes/SoftLayer_Virtual_Guest) and [VirtualGuest jclouds](https://github.com/jclouds/jclouds/blob/master/providers/softlayer/src/main/java/org/jclouds/softlayer/domain/VirtualGuest.java)

Wdyt?

nacx:
If the properties in the DTO are fixed, even if they are many, I'd model all them in the domain object to allow that. A proper builder should leave the domain object interface clean enough to make it usable.
Mind opening a JIRA to track and discuss this?
I'll merge the PR once 1.9.1 is released.

jclouds-mirror:
Agreed! https://issues.apache.org/jira/browse/JCLOUDS-984

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/cf0e659c) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/b7020439). Thanks @grkvlt and apologies for the delay!

grkvlt:
NP thanks @nacx

PR #833, opened 2015-07-31
Add missing equals and hashCode methods

andrewgaul:
Possible cause of odd PageSet behavior.


andrewgaul:
@kahing Please review and see if this resolves odd `PageSet` behavior.

PR #834, opened 2015-07-31
change default behavior of log to print sensitive information

arvindn05:
change default behavior of log to print sensitive information


nacx:
Thanks for the PR @arvindn05. Given your (and others) last feedback received in the other PR, I agree there should be no problem in releasing 1.9.x with the current default, so let's close this PR and release 1.9.1 with the sensitive payloads disabled by default.
Thanks for discussing!

PR #836, opened 2015-08-02
1.9.x

zhangdong1981:


demobox:
@zhangdong1981 Could you explain what this commit is for? You seem to be merging changes into master from 1.9.x - is that intentional?

demobox:
It looks like this PR was unintentional - will close this for now.

@zhangdong1981 Please reopen and provide some detail if there is a change you're trying to merge!

PR #837, opened 2015-08-04
JCLOUDS-979: Fix range get for LocalBlobStore

rcoedo:


rcoedo:
Tests and style pass on my machine, am I missing something?

zack-shoylev:
Seemed like a jenkins problem to me.

zack-shoylev:
Looks good to me!

andrewgaul:
:+1: Uses unbounded memory for the jdbc blobstore but we can fix this up later.

zack-shoylev:
This should be good to merge, but I guess a backport will be needed as well?

andrewgaul:
We can skip backporting since JDBC blobstore does not exist in 1.9.x and I did not backport the original memory saving patch for range gets.

PR #839, opened 2015-08-06
Fix LocalBlobStore

rcoedo:
The ByteSource used to build the payload when using ranges was being recalculated for every range. This commit moves that logic outside the loop.


rcoedo:
This commit will fix the problems in the jdbc api and in the h2 provider.

zack-shoylev:
+1

zack-shoylev:
merged

PR #840, opened 2015-08-06
JCLOUDS-883: Replace metadata during Swift copy

andrewgaul:
Previously Swift copy would append metadata instead of replacing it.


andrewgaul:
@zack-shoylev @kahing Please review.

kahing:
:+1:

TerryHowe:
lgtm

zack-shoylev:
This does change the default behavior of copy, though. Isn't it better to add an additional copy method that has fresh metadata as a parameter? I don't like changing the behavior of an existing method that is already being used. Thoughts? Am I missing something?

andrewgaul:
@zack-shoylev We could add a a third method or a flag to expose the full Swift support for appending metadata.  However I think the default should be to replace metadata when provided as this what other providers do and I imagine what most people expect.

zack-shoylev:
@andrewgaul Alright, I think it makes sense to switch the default behavior for consistency's sake, but let's document that in the method doc and make sure we add another method or a flag for the full functionality.

PR #841, opened 2015-08-07
JCLOUDS-827 updating sshj

stuarthendren:
Moved sshj to the new com.hierynomus version.
This only needed a minor code change due to a method removal.

However, due to the changed groupId but the same package declarations
maven treats them as different and does not exclude the versions imported by
jsch.agentproxy so thse are manually excluded and updated to the 0.0.9 versions

Also sshj is depends on a newer version of BouncyCastle. This has been updated to 1.51.
The inclusion of the extended version in the bouncycastle driver causes a duplicate
class clash. This could be ignored in the project pom but I have reduced it to the
normal version. Users could still provide the extended version if they require it.

### DIFF
--- a/drivers/jsch/pom.xml
+++ b/drivers/jsch/pom.xml
@@ -90,12 +90,12 @@
     <dependency>
       <groupId>com.jcraft</groupId>
       <artifactId>jsch.agentproxy.jsch</artifactId>
-      <version>0.0.8</version>
+      <version>0.0.9</version>
     </dependency>
     <dependency>
       <groupId>com.jcraft</groupId>
       <artifactId>jsch.agentproxy.connector-factory</artifactId>
-      <version>0.0.8</version>
+      <version>0.0.9</version>
     </dependency>
   </dependencies>
 
--- a/project/pom.xml
+++ b/project/pom.xml
@@ -232,6 +232,7 @@
     <auto-service.version>1.0-rc2</auto-service.version>
     <auto-value.version>1.1</auto-value.version>
     <java-xmlbuilder.version>1.1</java-xmlbuilder.version>
+    <sshj.version>0.12.0</sshj.version>
     <http.proxyHost />
     <http.proxyPort />
     <jclouds.wire.httpstream.url>http://archive.apache.org/dist/commons/logging/binaries/commons-logging-1.1.1-bin.tar.gz</jclouds.wire.httpstream.url>
@@ -360,6 +361,18 @@
         <groupId>com.jamesmurty.utils</groupId>
         <artifactId>java-xmlbuilder</artifactId>
         <version>${java-xmlbuilder.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.hierynomus</groupId>
+        <artifactId>sshj</artifactId>
+        <version>${sshj.version}</version>
+        <exclusions>
+        <!-- provided by the jclouds-bouncycastle driver -->
+        <exclusion>
+          <groupId>org.bouncycastle</groupId>
+          <artifactId>bcprov-jdk15on</artifactId>
+        </exclusion>
+      </exclusions>
       </dependency>
     </dependencies>
   </dependencyManagement>
@@ -577,12 +590,12 @@
                 <dependency>
                   <groupId>com.jcraft</groupId>
                   <artifactId>jsch.agentproxy.core</artifactId>
-                  <version>0.0.8</version>
+                  <version>0.0.9</version>
                 </dependency>
                 <dependency>
                   <groupId>com.jcraft</groupId>
                   <artifactId>jsch.agentproxy.connector-factory</artifactId>
-                  <version>0.0.8</version>
+                  <version>0.0.9</version>
                 </dependency>
               </conflictingDependencies>
               <packages>
--- a/drivers/bouncycastle/pom.xml
+++ b/drivers/bouncycastle/pom.xml
@@ -62,7 +62,7 @@
     <dependency>
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcprov-ext-jdk15on</artifactId>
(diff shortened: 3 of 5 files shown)


stuarthendren:
May also be relevant:
https://issues.apache.org/jira/browse/JCLOUDS-451
https://github.com/jclouds/jclouds/pull/274
https://github.com/jclouds/jclouds/pull/278

gionn:
:thumbsup:

nacx:
Thanks @stuarthendren!

>The inclusion of the extended version in the bouncycastle driver causes a duplicate
class clash

Where do the conflicting classes come from? I'm wondering if it would be safer to ignore that in the pom instead of changing the dependency.

stuarthendren:
These are the offending classes:
```
[WARNING] Found duplicate and different classes in [org.bouncycastle:bcprov-ext-jdk15on:1.51, org.bouncycastle:bcprov-jdk15on:1.51]:
[WARNING]   org.bouncycastle.crypto.agreement.jpake.JPAKERound1Payload
[WARNING]   org.bouncycastle.crypto.agreement.jpake.JPAKERound2Payload
[WARNING]   org.bouncycastle.math.ec.endo.GLVTypeBParameters
[WARNING] Found duplicate classes/resources in test classpath.
```

However, looking at the source from the two (source) jar all the classes are the same.

stuarthendren:
I've added the exclusion. After adding those specific classes, more were found, then more. I guess the whole jar is detected as a clash. This might be to do with the jars being signed. So I have not put any restrictions and just left the whole jars as an exception.

stuarthendren:
Is there a way to re-trigger the build seems like the node went offline during the build?

nacx:
You can trigger the build by commenting: rebuild please
Do you know which jars put the `bcprov-jdk15on` dependency? Perhaps we can exclude it from there/replace it by the ext one so all jclouds deps use the same BC dependency?

nacx:
Or by closing/reopening the PR if it does not work...

stuarthendren:
It's the com.hierynomus sshj jar that has the dependency on bcprov-jdk15on and the jclouds-bouncycastle driver that is dependent on bcprov-ext-jdk15on

nacx:
Would it make sense then to declare the `com.hierynomus` in the `dependencyManagement` section just to configure the exclusion of the `bcprov-jdk15on` dependency? This way we could use the ext version without having to ignore the conflicts?

nacx:
Thanks @stuarthendren! This lgtm.
@demobox mind having a look?

demobox:
> @demobox mind having a look?

Sure! I'll try to get round to that this evening or tomorrow. Thanks for the ping, @nacx, and of course thanks to @stuarthendren for the fix! ;-)

drivers/jsch/pom.xml
```diff
@@ -90,12 +90,12 @@
...
       <groupId>com.jcraft</groupId>
       <artifactId>jsch.agentproxy.jsch</artifactId>
-      <version>0.0.8</version>
+      <version>0.0.9</version>
     </dependency>
     <dependency>
       <groupId>com.jcraft</groupId>
       <artifactId>jsch.agentproxy.connector-factory</artifactId>
-      <version>0.0.8</version>
+      <version>0.0.9</version>
```
demobox: Do we need the net.schmizz exclusion here too? If not here, then also not below?

project/pom.xml
```diff
@@ -361,6 +362,18 @@
         <artifactId>java-xmlbuilder</artifactId>
         <version>${java-xmlbuilder.version}</version>
       </dependency>
+      <dependency>
```
demobox: Wondering if this adds anything here? `sshj` is only declared as a dependency in one library. If anything, the version that keeps reappearing is the `jsch.agentproxy` version, not the sshj version.

project/pom.xml
```diff
@@ -361,6 +362,18 @@
...
+        <groupId>com.hierynomus</groupId>
+        <artifactId>sshj</artifactId>
+        <version>${sshj.version}</version>
+        <exclusions>
+        <!-- provided by the jclouds-bouncycastle driver -->
+        <exclusion>
+          <groupId>org.bouncycastle</groupId>
+          <artifactId>bcprov-jdk15on</artifactId>
+        </exclusion>
+      </exclusions>
```
demobox: [nit] Indenting

demobox:
@stuarthendren, @nacx: Overall, looks good. I'd probably make the following small changes, but I don't think we need to update the PR for that - we can apply and test those during the merge:

1. Move the declaration of the `sshj` dep back into `drivers/sshj/pom.xml`
1. Figure out whether or not we need exclusions on `jsch.agentproxy.connector-factory`, and apply them consistently in the `jsch` and `sshj` drivers
1. Minor indenting fix

Does that sound like it makes sense?

drivers/jsch/pom.xml
stuarthendren: I think one or the other is required. You can see earlier versions had the exclusion here but not below. If you don't have either exclusion you get the duplicate class issue.

demobox: > I think one or the other is required.

Interesting...it doesn't _seem_ as though `connector-factory` should [depend on `sshj`](https://github.com/ymnk/jsch-agent-proxy/blob/master/jsch-agent-proxy-connector-factory/pom.xml#L16-L42) in any way. Wonder what's going on there...

demobox:
> I'd probably make the following small changes

See #844 for this - locally at least, it seems like we can actually remove the "conflicting deps" stuff. Not 100% convinced about my local environment, though, so I'd like to run the PR builder to check.

demobox:
Committed to master - see [this comment for #841](https://github.com/jclouds/jclouds/pull/844#issuecomment-130388904)

PR #842, opened 2015-08-10
JCLOUDS-872: Swift copy object and append metadata

andrewgaul:
The existing method replaces metadata.


andrewgaul:
@zack-shoylev Follow on to #840.  I added some live tests but running `mvn integration-test -pl :rackspace-cloudfiles-us` does not seem to run tests from `ObjectApiLiveTest`.  Does the `rackspace-cloudfiles-us` provider need to subclass this class?

zack-shoylev:
It seems the two providers are not extending the features tests from rackspace-cloudfiles, which is why you are seeing this.

We should just extend all of these to the providers, instead of testing them separately (hassle). It might make sense to directly extend from swift where applicable.

-Zack

________________________________

andrewgaul:
Subclasses `ObjectApiLiveTest` in #843.  We can address other classes in subsequent commits.

andrewgaul:
rebuild please

PR #846, opened 2015-08-18
JCLOUDS-826: Remove filesystem

rcoedo:


andrewgaul:
:-1: The filesystem blobstore is useful to present existing data via a blobstore interface as well as for larger tests.  Further I have existing users of it via S3Proxy.

rcoedo:
This PR was opened to discuss the removal of filesystem with the community, which was one of the points on my GSoC proposal.

When I started the JDBC Blobstore one of the requisites was portability between different database technologies. Also, we decided to store the blobs in the database. Unfortunately, the blob storage apis offered by databases are not standard, so it's really hard to offer a generic jdbc blobstore without an object mapper. We ended up with something closer to a JPA Blobstore instead of a JDBC blobstore.

IMHO jdbc blobstore and filesystem blobstore are different things and both should stay. Maybe in the future we can come up with an hybrid database which stores the blobs in the filesystem and the metadata in an embedded database. Storing only metadata in the database would make way easier to implement the jdbc calls in a generic way. It would be a better replacement.

zack-shoylev:
@andrewgaul Let's discuss this a bit more. The jdbc is supposed to be a simpler os-neutral implementation and replacement of filesystem. It should have the same capabilities. Is there a problem with the underlying idea?

zack-shoylev:
@nacx ping as well.

zack-shoylev:
Here is a tentative plan:

1. Promote jdbc to jclouds.
2. Keep both jdbc and filesystem for a while, transition to the jdbc API in projects.
3. Eventually figure out if we want to keep filesystem (maybe consider this for 2.0 or after).

kahing:
The filesystem blobstore has the advantage of working with existing files. There maybe some advantages of a jdbc only blobstore, but I think people will vastly prefer one that is closer to the existing filesystem blobstore. The optimization that @rcoedo suggested (storing metadata in a database) is a great idea for the filesystem blobstore.

nacx:
The issue basically tries to address the limitation that some filesystems are unable to store the metadata. Personally I don't see the jdbc as a replacement for the filesystem blobstore, but I also understand it can be useful in some test scenarios. I'd like to get jdbc promoted to the main repo but I think the filesystem provider comes in handy in many situations and we should try to keep it.

demobox:
For what it's worth, I would much rather have a discussion at this point about **adding** an additional blobstore implementation (and then testing it out to see how useful it actually is) than about **removing** the filesystem blobstore.

The latter certainly has its disadvantages, and I think we need to communicate clearly what they are. But as far as I can see that doesn't mean that it's "useless", "fundamentally broken" or "needs to be removed."

A database-based blobstore is an interesting alternative that we should certainly experiment with (I wonder whether we'll see a variant based on Docker data containers - @andreaturli ping? ;-)), but requiring a database **and** an object mapper as we currently do significantly raises the convenience bar for usage.

I would certainly not say that the current JPA blobstore, as far as I understand it, is a "drop-in replacement" for the filesystem blobstore at present.

zack-shoylev:
Thank you all for the feedback! With the (big) exception of managing existing files, I think a drop-in replacement should be more-or-less possible functionally (I am not sure what other limitations there are). 

The current filesystem metadata implementation, while limited, works. JDBC is a nice alternative to have. I am not so sure about implementing a hybrid - why not just use the JDBC one if you need more metadata?

I will be closing this PR. It seems we should focus on promoting JDBC to the main repo for now. I still think long-term the JDBC provider should prove to be more resilient - better metadata storage and no recurring cross-platform problems, but perhaps a full replacement is not reasonable at this point.

PR #849, opened 2015-08-20
replace old style asserts with AssertJ's assertThat

kahing:
This is a straightforward replace of

assert x -> assertThat(x).isTrue()

It doesn't specialize cases like:

assert x == null -> assertThat(x).isNull()


andrewgaul:
I don't think that we should merge this; instead we should actually specialize the cases.

PR #850, opened 2015-08-20
Always return ETag in filesystem object store get response

flandr:
Adds a property that instructs filesystem object store providers to
compute the ETag on demand when the content-md5 is not available in
an extended attribute. Defaults to 'false', preserving existing
behavior.


flandr:
I've left this optional to preserve existing behavior; would happily remove the property and do this by default if suggested.

andrewgaul:
Can you change this to always provide an ETag for consistency with real blobstores?  Currently we return the ETag if the `putBlob` request included Content-MD5 or generate it if the filesystem does not support `UserDefinedFileAttributeView` (Mac OS X, NFS mounts).  We can either remove the property and unconditionally provide ETag or keep the property and use it on the non-`UserDefinedFileAttributeView` code path.

flandr:
> Can you change this to always provide an ETag for consistency with real blobstores? Currently we return the ETag if the putBlob request included Content-MD5 or generate it if the filesystem does not support UserDefinedFileAttributeView (Mac OS X, NFS mounts). We can either remove the property and unconditionally provide ETag or keep the property and use it on the non-UserDefinedFileAttributeView code path.

I'll happily remove the option. The current behavior is slightly different than you describe: the filesystem object store always computes the actual MD5 hash on puts when it writes the extended attribute. In further regrettable news, `getUserDefinedFileAttributeView` returns non-null on OSX, even though user extended attribute views are not actually supported by the JVM on that platform. 

Will remove the property and the conditional block.

andrewgaul:
Pushed to master as 496e27f1afa32b90d0656c3f21387cf68a30cb31.  Thank you for your contribution @flandr!

kahing:
`LocalBlobStore` calls getBlob on the storage strategy during listContainer, this change will make it unbearably slow for the cases when the md5 needs to be generated at run time.

flandr:
> LocalBlobStore calls getBlob on the storage strategy during listContainer, this change will make it unbearably slow for the cases when the md5 needs to be generated at run time.

That's unfortunate, though it looks like this patch simply closed a corner case on OS X, where the user extended attribute view is non-null but still extended attributes are not supported; on a system where `getUserDefinedFileAttributeView` was returning null anyway, we'd already have recomputed the hash.

I'm honestly fine with behavior going either way, but would prefer uniformity.

kahing:
The case I am concerned about is when jclouds is used to expose files not written through jclouds. In those cases getUserDefinedFileAttributeView() would not return null but the extended attribute would not have been written.

flandr:
> The case I am concerned about is when jclouds is used to expose files not written through jclouds. In those cases getUserDefinedFileAttributeView() would not return null but the extended attribute would not have been written.

I'm fine with it if you'd like to revert this change. I _would_ like to understand why the expected null view is not in fact null on OS X, but I can approach that separately.

andrewgaul:
Reverted in 86e947ddedfa1a13976f9debeadae1da6f2f0190.

PR #856, opened 2015-09-11
JCLOUDS-217: do not double decode strings.

timuralp:
This PR removes the decoding multi map from URI and makes changes primarily to tests to not work around its behavior. This does not fix an existing issue, but I believe it simplifies the internals of handling parameters in jclouds.

Background
-----------------

Currently, calling ```addQueryParam()``` when constructing an HttpRequest results in inserting the key and the value into a decoding MultiMap. What it means is that the parameter value is decoded prior to being inserted and then encoded again on retrieval.

Further, there is some additional logic that checks whether a string is encoded or not in the ```Strings2.urlDecode()``` method. This behavior is also problematic such that providers need to make sure they encode strings before adding them as a query parameter, as they may not get the expected result back (e.g. a string that contains "+" is decoded into " " and encoded into "%20").

The jclouds code currently works around these issues, however, I believe the parameter handling is not clear to developers and leads to confusion. The PR suggests a more straightforward approach:
1. all query parameters will be encoded by ```addQueryParam()```
2. the query parameters are _never_ decoded by ```addQueryParam()```

I tested the PR with Azure, AWS S3, Swift, and the Filesystem backend. I removed one workaround from AWS S3, but otherwise all providers were functional. The majority of the changes are in fixing up tests to not use encoded strings (as these now become doubly encoded, so a parameter "my%20account" is turned into "my%2520account".

### DIFF
--- a/apis/cloudstack/src/test/java/org/jclouds/cloudstack/compute/CloudStackComputeServiceAdapterExpectTest.java
+++ b/apis/cloudstack/src/test/java/org/jclouds/cloudstack/compute/CloudStackComputeServiceAdapterExpectTest.java
@@ -56,7 +56,7 @@ public class CloudStackComputeServiceAdapterExpectTest extends BaseCloudStackCom
         .addQueryParam("command", "queryAsyncJobResult")
         .addQueryParam("jobid", "50006")
         .addQueryParam("apiKey", "APIKEY")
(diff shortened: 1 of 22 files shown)


andrewgaul:
:+1:

nacx:
+1 Thanks for all this fixes @timuralp!
Are there required changes in the labs repos too, once this is merged?

timuralp:
Rebased. I'll look into GCS in labs and see if anything else may be broken.

PR #858, opened 2015-09-16
JCLOUDS-987: Create security groups in AWS if subnetId is specified

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-987

The changes required to make this work are basically:

* Call the logic to create the security groups also when the subnet id is set.
* Get the VPC id for the given subnet, as it is mandatory when creating security groups for instances that run in the VPC.
* Use the security group id instead of the name, as it is mandatory when creating them in an VPC.

The changes are pretty straightforward, but there is a subtle change: AWS-EC2 and the generic EC2 providers have a cache `"region/group" -> String` where the resulting String is the security group name for that group. In AWS-EC2 now the resulting String is the security group ID instead of its name, allowing the creation of security groups in VPCs (and saving an API call in most cases). The id form is always used to create security groups in AWS-EC2 (the generic EC2 provider remains unchanged as there is no VPC support there).

@abayer a quick review would be highly appreciated.
@richardcloudsoft Could you test if this change fixes the issue?


nacx:
Looks like a transient Jenkins failure: rebuild please!

nacx:
Rebased, amended the commit message and pushed to master as [985d4512](http://git-wip-us.apache.org/repos/asf/jclouds/commit/985d4512).

PR #860, opened 2015-09-22
Fix openstack nova when floating ips are not available

andreaturli:
- improve destroyNode that could remove keypair and/or securityGroup if required

### DIFF
--- a/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeService.java
+++ b/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeService.java
@@ -20,9 +20,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_NODE_RUNNING;
 import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_NODE_SUSPENDED;
 import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_NODE_TERMINATED;
-import static org.jclouds.openstack.nova.v2_0.predicates.KeyPairPredicates.nameMatches;
-
-import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
@@ -42,7 +39,6 @@ import org.jclouds.compute.domain.NodeMetadata;
 import org.jclouds.compute.domain.TemplateBuilder;
 import org.jclouds.compute.extensions.ImageExtension;
 import org.jclouds.compute.extensions.SecurityGroupExtension;
-import org.jclouds.compute.functions.GroupNamingConvention;
 import org.jclouds.compute.internal.BaseComputeService;
 import org.jclouds.compute.internal.PersistNodeCredentials;
 import org.jclouds.compute.options.TemplateOptions;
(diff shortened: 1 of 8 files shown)


apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
```diff
@@ -81,14 +82,16 @@ public AllocateAndAddFloatingIpToNode(@Named(TIMEOUT_NODE_RUNNING) Predicate<Ato
 
       Optional<FloatingIP> ip = allocateFloatingIPForNode(floatingIpApi, poolNames, node.getId());
       if (!ip.isPresent()) {
-         throw new InsufficientResourcesException("Failed to allocate a FloatingIP for node(" + node.getId() + ")");
+        RegionAndId regionAndId = RegionAndId.fromSlashEncoded(node.getId());
+        novaApi.getServerApi(regionAndId.getRegion()).delete(regionAndId.getId());
```
nacx: Should we do here the same security group and key pair cleanup?

nacx:
The test failures are real test failures:
https://jclouds.ci.cloudbees.com/job/jclouds-pull-requests/2053/org.apache.jclouds.api$openstack-nova/testReport/

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/handlers/NovaErrorHandler.java
```diff
@@ -97,6 +97,13 @@ else if (message.indexOf("already exists") != -1)
                exception = new ResourceNotFoundException(message, exception);
             }
             break;
+        case 500:
```
zack-shoylev: Is this a different fix or part of the same fix?

zack-shoylev:
Looks like good cleanup to me overall

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/handlers/NovaErrorHandler.java
andreaturli: it is part of the same fix, when floating IP are finished, openstack returns 500 instead of https://github.com/andreaturli/jclouds/blob/fix/openstack-nova-floating-ips/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/handlers/NovaErrorHandler.java#L84 so I added that case to trigger the rollback cleanly.

zack-shoylev: Can you add this comment to the code. It seems to me like somewhat inconsistent openstack behavior.

andreaturli:
thanks @zack-shoylev, I've added the comment to justify the case 500

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeService.java
```diff
@@ -58,34 +54,18 @@
...
 import com.google.common.util.concurrent.ListeningExecutorService;
 
 @Singleton
 public class NovaComputeService extends BaseComputeService {
-   protected final NovaApi novaApi;
-   protected final LoadingCache<RegionAndName, SecurityGroupInRegion> securityGroupMap;
-   protected final LoadingCache<RegionAndName, KeyPair> keyPairCache;
-   protected final Function<Set<? extends NodeMetadata>, Multimap<String, String>> orphanedGroupsByRegionId;
-   protected final GroupNamingConvention.Factory namingConvention;
+   protected CleanupServer cleanupServer;
```
nacx: Make it final.

nacx:
Just a minor comment. lgtm @andreaturli. Thanks!

andreaturli:
@nacx still good for merging?

nacx:
Go ahead!

andreaturli:
Merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a4a255fa)

PR #868, opened 2015-10-22
UserAdd: guard groupadd with check if group exists

aledsage:
On SUSE, the “-f” force option is not available for groupadd,
so `groupadd -f wheel` returns exit code 9 if the group already
exists. To avoid this, first check if the group exists.

In normal usage, this doesn’t matter: the script continues with the
next command anyway.

However, if the statements generated by UserAdd or AdminAccess are
used outside of that context (e.g. by code external to jclouds), then
this can cause them to fail.


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/502d96d2) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/bdbed1ce). Thanks @aledsage!

PR #869, opened 2015-10-23
[elasticstack] fix for incorrect logging message

stuarthendren:
Was logging when it was present rather than when it had failed to find it.


andrewgaul:
Pushed to master as ada17d3e476e70423460b2828c836eeb7d2a0203.  Thank you for your contribution @stuarthendren!

PR #872, opened 2015-10-27
Move QueryParam encoding to a separate class.

timuralp:
The patch implements a QueryValue class, which encodes the underlying
value based on whether the "encoded" flag is set. This class is used
by the RestAnnotationProcessor to propagate the @Encoded value set on
any parameters.

Since the encoding is now handled by the QueryValue instances, we
should no longer call encodeQueryLine() in the URI builder and instead
call buildQueryLine(). The caveat is that we need to make sure all of
the parameters that may need to be encoded are converted to QueryValue
objects. This is done by converting Object instances to QueryValue by
an instance of the TransformObjectToQueryValue when adding any query
parameters to the URI.


timuralp:
@nacx is this in line with what you had in mind?

core/src/main/java/org/jclouds/http/utils/QueryValue.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.jclouds.util.Strings2;
+
+import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableList;
+
+public class QueryValue implements Comparable {
+   private boolean encoded;
+   Object value;
+   Iterable<Character> skipChars;
```
nacx: Make the variables private and final?

nacx:
Yes, I like this approach. Thanks @timuralp!

nacx:
Pushed to [master](https://git-wip-us.apache.org/repos/asf?p=jclouds.git;a=commit;h=1fc9b0e2597ee0e11d8ae09db478c2dbf516912b). Thanks @timuralp!

PR #873, opened 2015-10-27
Added context information for exceptions that are hard to debug

nacx:
Throw an exception with a more detailed message to help debugging reflection issues. See:
http://markmail.org/message/wsthwxkrggoqafmh

/cc @gionn


gionn:
```
java.lang.IllegalStateException: Incorrect number of names on @org.jclouds.json.SerializedNames(value=[slug, name, sizes, available, features]).
 Class [org.jclouds.digitalocean2.domain.AutoValue_Region]. Annotation config: [slug,name,sizes,available,features]. Fields in object; [java.lang.String,java.lang.String,java
.util.List,boolean,java.util.List,[Z]
```

No idea of what [Z] means, but it's probably a syntetic field added by jacoco as explained in [FAQs](http://eclemma.org/jacoco/trunk/doc/faq.html):

> My code uses reflection. Why does it fail when I execute it with JaCoCo?
> 
> To collect execution data JaCoCo instruments the classes under test which adds two members to the classes: A private static field $jacocoData and a private static method $jacocoInit(). Both members are marked as synthetic.
>
> Please change your code to ignore synthetic members. This is a good practice anyways as also the Java compiler creates synthetic members in certain situation.

How can we proceed?

nacx:
That could explain the failure. I'd say we can completely ignore synthetic fields.

The deserialization strategy just tries to match the number of "names" in the annotation with the number of parameters in the constructor (or static factory method), to be able to call it passing the right value to each parameter.

The serialization strategy, when using auto value, assumes that the fields are declared in the same order than the parameters of the constructor/factory method and the annotation. We live with this assumption, and it is used to determine, for each field, which value in the annotation corresponds to it, in order to serialize it under the proper json key. 

This said, since synthetic methods are not in the source code, we should ignore them. I'll raise a JIRA issue for this and update this PR accordingly.

nacx:
See: https://issues.apache.org/jira/browse/JCLOUDS-1030

nacx:
@gionn I've just pushed a commit that filters out the synthetic members. Could you try again building this branch again?

gionn:
@nacx worked like a charm! :beer::thumbsup:

nacx:
Squashed, amended the commit message and pushed to master as [afd3c322](http://git-wip-us.apache.org/repos/asf/jclouds/commit/afd3c322).
@gionn kudos to you for your investigation of the issue!

PR #876, opened 2015-11-08
Fix aggregator poms

nacx:


demobox:
Thanks for catching this, @nacx!

PR #877, opened 2015-11-10
Remove HPCloud from the aggregator poms

nacx:


all/pom.xml
```diff
@@ -44,11 +44,6 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>org.apache.jclouds.api</groupId>
-      <artifactId>chef</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-    <dependency>
```
nacx: This one was duplicated.

PR #878, opened 2015-11-13
Delete parts when removing Swift multipart objects

andrewgaul:
Swift implements multi-part upload with user-visible parts and an
explicit manifest.  When deleting an MPU blob it can delete only the
manifest or both the manifest and parts.  For consistency with other
providers, we now do the latter in the portable abstraction.  Swift
ignores the multipart-manifest=delete parameter for single-part
objects.  Fixes andrewgaul/s3proxy#92.

### DIFF
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
@@ -344,7 +344,8 @@ public class RegionScopedSwiftBlobStore implements BlobStore {
 
    @Override
    public void removeBlob(String container, String name) {
-      api.getObjectApi(regionId, container).delete(name);
+      // use SLO API to delete blob regardless of whether its a single- or multi-part object
+      api.getStaticLargeObjectApi(regionId, container).delete(name);
    }
 
    @Override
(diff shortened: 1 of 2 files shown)


andrewgaul:
@zack-shoylev Please review.

zack-shoylev:
+1

andrewgaul:
@zack-shoylev We might need to revert this; a user reported the following error in andrewgaul/s3proxy#92:

```
11:04:28.854 [S3Proxy-22 - /testbucket/s3proxy] DEBUG jclouds.wire - << "{"Number Not Found": 0, "Response Status": "400 Bad Request", "Errors": [["/testbucket/s3proxy", "Not an SLO manifest"]], "Number Deleted": 0, "Response Body": ""}"
```

Can you help verify this?  I am not available for another week.

andrewgaul:
I reproduced the symptoms and reverted this pull request in 4abd5dd16f72b8233b914ace18dee02d1a9e57dc.  jclouds incorrectly drops the 400 bad request error, likely when correctly ignoring 404 file not found errors.

zack-shoylev:
@andrewgaul : is this the error retry handler messing up the 400? or something else

blop:
Any progress ?
Current 2.0.0-SNAPSHOT does not remove a multipart blob completely.

Shouldn't we create a jclouds issue for this?

andrewgaul:
@blop This pull request only deals with mapping delete single-part object to delete multi-part object.  If you have a problem with delete multi-part object please open an issue with a test case that fails.  I believe that `StaticLargeObjectApiLiveTest.testDelete` exercises this feature already?

blop:
I'm using the org.jclouds.openstack.swift.v1.blobstore.RegionScopedSwiftBlobStore through the org.jclouds.blobstore.BlobStore interface, and I have no way to remove a multi-part object other than calling removeBlob(String container, String name) because I cannot list all the part names either (which would allow me to call removeBlobs(String container, Iterable<String> names)).

When calling removeBlob() (like I would on a normal object) some of the parts are left in the container.

Or did I get that wrong? :)

andrewgaul:
@blop I see, you have the same problem that this pull request intended to solve.  I reverted it since Swift does not allow deleting single-part objects with the multi-part delete call.  I agree that the portable `BlobStore.removeBlob` should work for both single- and multi-part objects although this is difficult due to Swift API decisions.  We could issue a HEAD request to determine whether to call the correct single- or multi-part object delete call although this requires an extra RPC and does not work for the `BlobStore.deleteObjects` multi-delete RPC.  Could you open an issue to track this?

PR #880, opened 2015-11-18
Fixes tenantId when using the keystone admin extension

zack-shoylev:


zack-shoylev:
Also should be backported.

apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/options/CreateUserOptions.java
```diff
@@ -97,7 +97,7 @@ private ServerRequest(String name, String password) {
       if (password != null)
          user.password = password;
       if (tenant != null)
-         user.tenant = tenant;
+         user.tenantId = tenant;
```
demobox: Would it make sense to also rename the other variable to `tenantId`, or should we leave that as it is?

demobox:
@zack-shoylev: you beat me to it ;-)

apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/options/CreateUserOptions.java
zack-shoylev: I would rather not change the other one, as then it would be consistent to also change the getters/setters - and break backwards compatibility.

zack-shoylev:
merged and backported

PR #881, opened 2015-11-21
don't remove directory if it's a blob

kahing:
the filesystem blob store was always removing a/ when a/b was
removed, even if the client explicitly created a/. this patch
fixes that


zack-shoylev:
Ah, that's a good change. +1

PR #882, opened 2015-11-21
JCLOUDS-1038: Fix date deserialization in EC2 apis

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1038

This change assumes the fact that dates can come in different formats and uses a method that keeps the current behaviour and falls back to the `iso8601SecondsDateParse` parser.


nacx:
Pushed to master as [86dadc3a](http://git-wip-us.apache.org/repos/asf/jclouds/commit/86dadc3a).

PR #884, opened 2015-11-24
Fix filesystem api in OSX

nacx:
/cc @andrewgaul @kahing

### DIFF
--- a/apis/ec2/src/main/java/org/jclouds/ec2/xml/AttachmentHandler.java
+++ b/apis/ec2/src/main/java/org/jclouds/ec2/xml/AttachmentHandler.java
@@ -22,8 +22,7 @@ import javax.annotation.Resource;
 import javax.inject.Inject;
 
 import org.jclouds.aws.util.AWSUtils;
-import org.jclouds.date.DateCodec;
-import org.jclouds.date.DateCodecFactory;
+import org.jclouds.date.DateService;
 import org.jclouds.ec2.domain.Attachment;
 import org.jclouds.http.functions.ParseSax;
 import org.jclouds.location.Region;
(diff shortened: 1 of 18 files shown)


demobox:
@nacx Were both of these commits intended to be part of this PR? It looks like only one small change is actually about OS X filesystem support?

nacx:
For some reason the github PR preview shows both commits when the PR was actually rebased to the latest master. Look at the commit hash of the ec2 fix (whcih comes from a previous PR); it matches the commit that was currently on master.

The pr did actually contain only 1 commit, but the github preview showed both. Don't know why. The master commit history looks clean too.

demobox:
> The master commit history looks clean too.

I'm sure all is well there ;-) Just wanted to check that the relevant bit of the change is indeed only:
```
-            if (!view.list().isEmpty()) {
+            if (view == null) { // OSX HFS+ does not support UserDefinedFileAttributeView
+                logger.debug("Could not look for attributes from %s", directory);
+            } else if (!view.list().isEmpty()) {
```

Regarding that: what's our feeling on making `logger.debug` calls that are not wrapped by `logger.isDebugEnabled`?

nacx:
>Regarding that: what's our feeling on making logger.debug calls that are not wrapped by logger.isDebugEnabled?

The [jclouds loggers](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/logging/BaseLogger.java#L47-L50) already take care of that.

demobox:
> The jclouds loggers already take care of that.

Er...ah, yes, of course - this is _jclouds_ logging, not SLF4J. Doh! ;-)

Thanks, Ignasi!

nacx:
jclouds FTW! ^^

kahing:
:+1:

PR #885, opened 2015-11-26
JCLOUDS-512: Implement the ImageCache

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-512

TL;DR: A proper image cache for the compute abstraction.

I've been thinking about this for a long time and approached the problem in several ways. This is the less intrusive fix I've come up with after several painful refactors. The core changes are not obvious in the PR diff, so I'll try to summarise them.

**Background:**
Images from a provider are exposed to the user as a `Set`. Internally, that set is accessed through a memoized `Supplier` that acts as a cache, and jclouds uses it to improve response times when talking to providers where getting the image list is an expensive operation.

This cached set is *immutable* and gives little control over its lifecycle, and this presents issues with the `ImageExtension`. Users can use it to create and delete images in the provider, but those changes are not reflected in the cache (given its nature), and inconsistent behaviors start to appear in subsequent uses of the `TemplateBuilder`.

**Proposed fix:**
Stop using a cached set and use a proper `Cache<String, Image>` structure. To avoid having to refactor all the providers, as the injection of the cached `Supplier<Set<Image>>` is a common practice, the cache  will also implement `Supplier<Set<Image>>` and replace the old binding to make the change transparent for the cache consumers.

**Implementation details and considerations:**
* The `ComputeService` returns a `Set<Image>`, and it is convenient to keep that signature; we don't want to couple it to a `Cache` and the fact that images are cached by jclouds.
* We want to keep the "old" `Supplier<Set<Image>>` to allow users to call it to (lazily) load the images in the provider, but we want to move the caching logic out of it.
* The `ImageCacheSupplier` (the `LoadingCache` implementation) uses the `GetImagesStrategy` to load individual images, and is initialised with the values of the `Supplier<Set<Image>>`. As the supplied set is lazily loaded, we don't want to call it when building the cache: it has to be initialised only when the supplier "loads live values". This has been resolved by generating an event when the supplier loads new values, and by subscribing the image cache to that event. This way the cache is populated with the images exactly when new values are fetched.
* A default implementation of the `ImageExtension` has been added to update the cache every time images are added or deleted.
* The `TemplateBuilder` now provides a method to bypass and reload the cache, as this has been asked several times:
```java
Template template = compute.templateBuilder().forceCacheReload(). ... .build();
```

I still have to run live tests in several providers before merging, but I'd like to get some early feedback:

@demobox @abayer @ccustine Your comments would be highly appreciated
@felfert Could you try this branch and see if it fixes the caching issues you were experiencing with Jenkins?


nacx:
I've added a commit that replaces the usage of the event bus to notify new values loaded by a regular callback. The event bus can be configured to be async, and we want the value loaded notification to be notified as soon as the loading happens, so async is not convenient and we shouldn't rely on the event bus being configured to be sync.

nacx:
I've added a commit that properly tests the image cache in EC2. It is tricky because in EC2 images are filtered by owner by default, and the images created by the ImageExtension are owned by the current user. When the image cache is refreshed due to the session expiration, the created image is no longer found (the current user is not in the list of queried owners, and we don't want to rebuild the context as we want to test the singleton image cache).

The test in EC2 directly inspects the ImageCache using reflection to avoid falling back to its lazy loading and avoiding an eventual cache refresh.

This is the latest change. I've tested several providers and in all them the new tests for the ImageCache are working as expected. I'll merge this PR once the build is green.

nacx:
Squashed and pushed to master as [40f31786](http://git-wip-us.apache.org/repos/asf/jclouds/commit/40f31786). Also pushed [f67d2aa7](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/f67d2aa7) to jclouds-labs to align it with the changes introduced in this PR.

PR #886, opened 2015-11-29
create OpenStack API by "Access" object or token

Xianguang-Zhou:
https://issues.apache.org/jira/browse/JCLOUDS-1040
https://issues.apache.org/jira/browse/JCLOUDS-1039


nacx:
Thanks for your contribution @Xianguang-Zhou! There several things that have to be addressed before we can review the PR in detail:

* First of all, remove all the unrelated changes: those `logback.xml` changed files and all the `Serializable` additions to all entities in the domain model. I don't think they are in the scope of this issue, so please, remove those changes from the PR.
* This change couples `jclouds-core` to very specific authentication details of a single provider, and we can't do that. The changes in core should be removed. (Could you also explain what that `accessReceiver` function is for?)
* Instead of reusing the `AuthenticatePasswordCredentials` to authenticate with the token, you should create a class that handles your new credential typeand register it [here](https://github.com/jclouds/jclouds/blob/master/apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/config/KeystoneAuthenticationModule.java#L218-L223).
* There are 33 unit tests failing (see the PR build details). All tests must be passing.

Once these points have been addressed we'll be able to do a detailed review. If you need help or guidance, please, comment here or find us in the jclouds dev@ mailing list or the #jclouds Freenode IRC channel. Thx!

andrewgaul:
@Xianguang-Zhou do we have a path forward on this pull request?

Xianguang-Zhou:
I have used other OpenStack client library, I don't need the pull request any more, thank you.

PR #887, opened 2015-12-03
[JCLOUDS-1044] fix NULL values handling in TypeAdapterFactories

kwart:
JIRA: https://issues.apache.org/jira/browse/JCLOUDS-1044

The `MapTypeAdapter` and `IterableTypeAdapter` classes under the `NullFilteringTypeAdapterFactories` didn't handle null values correctly and they tried to read an JSON object  (or array) instead of null.

That was source of exceptions like:

    com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_ARRAY but was NULL ...

### DIFF
--- a/apis/chef/src/main/java/org/jclouds/chef/config/ChefParserModule.java
+++ b/apis/chef/src/main/java/org/jclouds/chef/config/ChefParserModule.java
@@ -16,8 +16,6 @@
  */
 package org.jclouds.chef.config;
 
-import static com.google.common.base.Objects.equal;
-import static com.google.common.base.Objects.toStringHelper;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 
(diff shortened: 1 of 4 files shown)


nacx:
Thanks @kwart! The change itself LGTM.

The chef provider overrides this behavior, so the fix should be applied [here](https://github.com/jclouds/jclouds/blob/master/apis/chef/src/main/java/org/jclouds/chef/config/ChefParserModule.java#L236) too. Also, you'll see that the ChefParserModule [copies the entire map adapter](https://github.com/jclouds/jclouds/blob/master/apis/chef/src/main/java/org/jclouds/chef/config/ChefParserModule.java#L208-L209) (this is a legacy code created when Chef lived in a different Git repo). Could you make the `NullFilteringTypeAdapterFactories.MapTypeAdapter` non-final, so we can properly override just the relevant bits?

Just FYI, the difference in the Chef adapter is that it doesn't use an immutbale map to allow repeated keys and it returns just the last one. Some json returned in Chef use key repetition to put "comments", and leave the value in the last field. Pretty ugly, I know!

kwart:
Commit was updated to include also @nacx 's comments.

nacx:
Thanks @kwart! Pushed to master as [0fb1b459](http://git-wip-us.apache.org/repos/asf/jclouds/commit/0fb1b459).

PR #888, opened 2015-12-03
JCLOUDS-1045 remove swift dependency from allblobstore module

kwart:
JIRA: https://issues.apache.org/jira/browse/JCLOUDS-1045

The [JCLOUDS-786](https://issues.apache.org/jira/browse/JCLOUDS-786) removed swift api, but it's still used as dependency in the `allblobstore` module.


andrewgaul:
Thank you for your contribution @kwart!

PR #895, opened 2015-12-17
JCLOUDS-1042: Remove unneeded ACL swizzling

andrewgaul:
Test shows that the bucket access supersedes the blob access policy.

### DIFF
--- a/apis/s3/src/main/java/org/jclouds/s3/blobstore/S3BlobStore.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/blobstore/S3BlobStore.java
@@ -74,7 +74,6 @@ import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
 import com.google.common.base.Supplier;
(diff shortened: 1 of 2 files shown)


apis/s3/src/main/java/org/jclouds/s3/blobstore/S3BlobStore.java
```diff
@@ -273,13 +272,6 @@ public String putBlob(String container, Blob blob, PutOptions overrides) {
 
       // TODO: Make use of options overrides
       PutObjectOptions options = new PutObjectOptions();
-      try {
-         AccessControlList acl = bucketAcls.getUnchecked(container);
-         if (acl != null && acl.hasPermission(GroupGranteeURI.ALL_USERS, Permission.READ))
-            options.withAcl(CannedAccessPolicy.PUBLIC_READ);
-      } catch (CacheLoader.InvalidCacheLoadException e) {
-         // nulls not permitted from cache loader
-      }
```
andrewgaul: This commit also makes the above MPU code consistent with the single-part code.

PR #900, opened 2016-01-19
JCLOUDS-1053: Fallback to -1 when sshj exit status returns null

devcsrj:
[JIRA Ticket](https://issues.apache.org/jira/browse/JCLOUDS-1053).

Change: Returns -1, in case the sshj driver returns a null exit status.


drivers/sshj/src/main/java/org/jclouds/sshj/SshjSshClient.java
```diff
@@ -460,7 +460,7 @@ public ExecResponse create() throws Exception {
             Command output = session.exec(checkNotNull(command, "command"));
             String outputString = IOUtils.readFully(output.getInputStream()).toString();
             output.join(sshClientConnection.getSessionTimeout(), TimeUnit.MILLISECONDS);
-            int errorStatus = output.getExitStatus();
+            int errorStatus = output.getExitStatus() != null ? output.getExitStatus() : -1;
```
nacx: Move the logic to the [ExecResponse constructor](https://github.com/devcsrj/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/domain/ExecResponse.java#L29-L33)? This way the approach would be ssh driver agnostic.

nacx:
Thanks @devcsrj!
I'd also add a constant in the ExecResponse class itself for the -1 value, and try to unit test this using mocks.

devcsrj:
Added mock test. Also, updated `ssh_test.clj` for it seemed like clojure always casts "numbers" to `long`. Throws this exception without the explicit cast:
```
1) ClassCastException on node 9:
java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Integer
	at org.jclouds.ssh_test.NoOpClient.exec(ssh_test.clj:63)
	at org.jclouds.compute.callables.RunScriptOnNodeUsingSsh.runCommand(RunScriptOnNodeUsingSsh.java:107)
	at org.jclouds.compute.callables.RunScriptOnNodeUsingSsh.call(RunScriptOnNodeUsingSsh.java:80)
...omitted..
```

nacx:
Pushed to master as [a4b8a737](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a4b8a737). Thanks @devcsrj!

PR #903, opened 2016-01-22
JCLOUDS-1065: Deprecate inDirectory option

andrewgaul:
Superseded by prefix and delimiter support.

### DIFF
--- a/blobstore/src/main/java/org/jclouds/blobstore/options/ListContainerOptions.java
+++ b/blobstore/src/main/java/org/jclouds/blobstore/options/ListContainerOptions.java
@@ -31,7 +31,7 @@ import com.google.common.base.Objects;
  * import static org.jclouds.blobstore.options.ListContainerOptions.Builder.*
  * <p/>
  * BlobStore connection = // get connection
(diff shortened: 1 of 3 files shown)


andrewgaul:
@timuralp Please review.

timuralp:
I like this change and believe we should be able to remove the directory notions out of jclouds!

Having said that, there are some providers where the behavior between delimiter/prefix and directory notions is not quite identical (I won't mention the file system provider, because it's pretty convoluted there anyway). In swift, the path parameter (what is used for directories) may return different results than when using delimiter/prefix. It supports the notion of directory blobs, which prefix and delimiter would not actually implicitly create. However, I don't know how many people actually rely on that functionality. I do think jclouds should deprecate support for directory blobs and move away from using the path parameter. None of the above precludes deprecating the support, but to actually switch over, consumers may need to do additional work on top of changing the parameters.

andrewgaul:
As a rule, jclouds only includes a feature in the portable abstraction when three providers support it.  Since only Swift and Atmos natively support directories and other providers are unlikely to support this odd feature, we should not expose it in the portable abstraction.  Callers can use the provider-specific interface if needed.

PR #906, opened 2016-01-23
Fixes a NPE when null is expected.

zack-shoylev:
https://issues.apache.org/jira/browse/JCLOUDS-1064

### DIFF
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
@@ -476,7 +476,12 @@ public class RegionScopedSwiftBlobStore implements BlobStore {
    @Override
    public MultipartPart uploadMultipartPart(MultipartUpload mpu, int partNumber, Payload payload) {
       String partName = getMPUPartName(mpu, partNumber);
+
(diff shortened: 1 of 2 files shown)


zack-shoylev:
This might need some more work? I am not sure if we want to do additional retries, different behavior, or more tests.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -476,7 +476,12 @@ public String completeMultipartUpload(MultipartUpload mpu, List<MultipartPart> p
    @Override
    public MultipartPart uploadMultipartPart(MultipartUpload mpu, int partNumber, Payload payload) {
       String partName = getMPUPartName(mpu, partNumber);
+
       String eTag = api.getObjectApi(regionId, mpu.containerName()).put(partName, payload);
+      if (eTag == null) {
```
andrewgaul: Sorry I do not understand this.  Why is the ETag nullable for a successful put?  Otherwise we should raise an exception?

zack-shoylev: That is the question. From https://issues.apache.org/jira/browse/JCLOUDS-1064

I thought it simply indicated (service-side?) failure, even with no fallback annotation on the swift put.

PR #909, opened 2016-01-25
Delete single- and multi-part objects in Swift

andrewgaul:
Due to Swift's wonky MPU implementation we must try the multi-part
delete first followed by a single-part delete if it fails.  Previously
delete would leave orphan subobjects.  Fixes andrewgaul/s3proxy#92.

### DIFF
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
@@ -66,6 +66,7 @@ import org.jclouds.openstack.swift.v1.blobstore.functions.ToBlobMetadata;
 import org.jclouds.openstack.swift.v1.blobstore.functions.ToListContainerOptions;
 import org.jclouds.openstack.swift.v1.blobstore.functions.ToResourceMetadata;
 import org.jclouds.openstack.swift.v1.domain.Container;
+import org.jclouds.openstack.swift.v1.domain.DeleteStaticLargeObjectResponse;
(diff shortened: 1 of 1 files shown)


andrewgaul:
@zack-shoylev This commit makes me sad but I do not see another way around this.  Thoughts?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -347,9 +348,20 @@ public Blob getBlob(String container, String name, GetOptions options) {
...
+      // not an MPU we fall back to single-part delete.
+      DeleteStaticLargeObjectResponse response = api.getStaticLargeObjectApi(regionId, container).delete(name);
+      if (!response.status().equals("200 OK")) {
+         api.getObjectApi(regionId, container).delete(name);
+      }
    }
 
+   /**
+    * Delete multiple single-part objects.  Note that this cannot remove the
+    * subobjects of a mulit-part upload.
```
zack-shoylev: Small typo: mulit

zack-shoylev:
It looks fine to me, there shouldn't be an alternative especially considering how removeBlob should work. From what I understand, with the way different providers work with MPU, swift just shifts some of the complexity to dealing with manifests (instead of, say, assembling the MPU objects). Oh well.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
andrewgaul: Fixed.

PR #910, opened 2016-01-31
Use parseLong instead of parseInt in range parser

quod3:
@andrewgaul Use parseLong instead of parseInt when parsing open-ended byte ranges in LocalBlobStore. Without this fix, any "from byte x to end" or "from beginning to byte x" getBlob() call will throw a NumberFormatException if x is too big to fit into an int (2 GB).

Fixes https://issues.apache.org/jira/browse/JCLOUDS-1073


andrewgaul:
Thank you for your contribution @quod3!

PR #911, opened 2016-02-01
JCLOUDS-691: Remove 404 fallbacks of PUT and POST calls from OpenStac…

devcsrj:
…k Apis

openstack cinder, keystone, nova, nova-ec2, swift, trove


zack-shoylev:
+1

Just a note, we shouldn't be back-porting this (obviously). But it should roll out with v2.

nacx:
This is great. Thanks @devcsrj!
Just one comment: remove all tests that verify the propagated exceptions. It is the jclouds default behavior (not anything introduced by these APIs) so we shouldn't test this. That is already tested in jclouds-core.

devcsrj:
Just to be clear, does this mean that tests like [this](https://github.com/jclouds/jclouds/pull/911/files#diff-346bc00a46231f33755b2d3ec420cfbbR129) are no longer needed? Since the `ResourceNotFoundException` is a Fallback in the jclouds-core?

nacx:
Exactly. It is jclouds-core the one that throws the exception in the absence of a fallback. We shouldn't test that, as if at some point we change the default behavior we could end up having to update hundreds of tests in every provider.
Let's avoid re-testing core behavior in concrete providers.

devcsrj:
Removed *unneeded* tests, mostly checking `AuthorizationException`s and `ResourceNotFoundException`s. Please do tell if I went overboard (2,515 deletions).

nacx:
LGTM Thanks! Mind squashing the commits before we merge the PR?

nacx:
I'll merge this in short. For completeness, could you also apply the fix to the apis in the `jclouds-labs-openstack` repo?

nacx:
Pushed to master as [d83282ca](http://git-wip-us.apache.org/repos/asf/jclouds/commit/d83282ca). Thanks!

PR #912, opened 2016-02-02
JCLOUDS-1072

rongallagher:
JCLOUDS-1072 - Add support for environment-specific run lists to Role and Role.Builder

I would like to have this commit backported to the 1.9.x branch since that's the version of jClouds that we're currently using.

* Support for manipulating environment-specific run lists has been added to the RoleBuilder class, via the envRunList and envRunListElement methods.
* The Role class includes a new member variable (envRunList) that holds the environment-specific run lists.  This variable is serialized as 'env_run_lists'.  This variable is also incorporated into the equals, hashCode and toString methods.
* The constructor for the Role class now includes a new argument (envRunList) that's mapped to the 'env_run_lists' property in the JSON representation of a Role that's generated by Chef.
* I try to adhere to the mantra "If you write a line of code, then you write a test for that line of code".  With that in mind, this pull request includes a RoleTest class that exercises all of the code that I added or modified.

One thing that I feel compelled to draw attention to is the creation of the immutable envRunList in the build() method of the Role.Builder class.  I found nothing in the Google collections framework to handle nested, immutable collections, like what is needed to maintain the map of environment-specific run lists that are collected via the envRunList and envRunListElement methods in the Role.Builder class.  So, those methods manipulate a mutable java.util.Map whose values are instances of mutable java.util.List objects.  A separate immutable map, with nested immutable lists is created in the build() method and used when instantiating the final Role object.

### DIFF
--- a/apis/chef/src/main/java/org/jclouds/chef/domain/Role.java
+++ b/apis/chef/src/main/java/org/jclouds/chef/domain/Role.java
@@ -20,8 +20,11 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static org.jclouds.chef.util.CollectionUtils.copyOfOrEmpty;
 
 import java.beans.ConstructorProperties;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 
 import org.jclouds.domain.JsonBall;
 import org.jclouds.javax.annotation.Nullable;
@@ -44,6 +47,10 @@ public class Role {
       private ImmutableMap.Builder<String, JsonBall> overrideAttributes = ImmutableMap.builder();
       private ImmutableMap.Builder<String, JsonBall> defaultAttributes = ImmutableMap.builder();
       private ImmutableList.Builder<String> runList = ImmutableList.builder();
+      // envRunList is a nested set of collections. The Immutable* classes in google collections don't appear to
+      // support this nested immutability, so the builder will utilize native collections as the envRunList is
+      // assembled. An immutable, nested map of collections will be assembled in the build() method.
+      private Map<String, List<String>> envRunList = new HashMap<String, List<String>>();
 
       public Builder name(String name) {
          this.name = checkNotNull(name, "name");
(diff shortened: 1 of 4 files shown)


apis/chef/src/main/java/org/jclouds/chef/domain/Role.java
```diff
@@ -85,8 +92,39 @@ public Builder runList(Iterable<String> runList) {
...
+      }
+
       public Role build() {
-         return new Role(name, description, defaultAttributes.build(), runList.build(), overrideAttributes.build());
+         // Assemble an immutable envRunList where each entry is an immutable list of entries.
+         ImmutableMap.Builder<String, List<String>> immutableEnvRunList = ImmutableMap.builder();
+         for (Entry<String, List<String>> e : envRunList.entrySet()) {
+            ImmutableList.Builder<String> value = ImmutableList.builder();
+            value.addAll(e.getValue());
+            immutableEnvRunList.put(e.getKey(), value.build());
```
nacx: No need for a builder and a list here. Something like the following is more concise and easier to read?
```java
immutableEnvRunList.put(e.getKey(), ImmutableList.copyOf(e.getValue()));
```
Or use the more functional approach:
```java
immutableEnv = Maps.transformValues(envRunList, new Function<List<String>, List<String>>() {
   @Override public List<String> apply(List<String> input) {
      return ImmutableList.copyOf(input);
   }
});
```

apis/chef/src/test/java/org/jclouds/chef/domain/RoleTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+      Role role1 = Role.builder().envRunListElement(env, entry).build();
+      Role role2 = Role.builder().envRunListElement(env, entry).build();
+
+      assertEquals(role1.hashCode(), role2.hashCode(), "hashCodes should be equal");
+      assertEquals(role1, role2, "role1 should equal role2");
+      assertEquals(role2, role1, "role2 should equal role1");
+   }
+
+   public void rolesWithDifferentEnvRunListAreEqual() {
```
nacx: Typo: "AreNotEqual".

nacx:
Thanks for your contribution @rongallagher! Code looks good! There are just very minor comments.
Could you also update the [create Role live test](https://github.com/jclouds/jclouds/blob/master/apis/chef/src/test/java/org/jclouds/chef/ChefApiLiveTest.java#L244-L252), for completeness?

When making the changes, just add a commit and push the branch; it will automatically show in the pull request. Thanks for your first contribution to jclouds!

apis/chef/src/main/java/org/jclouds/chef/domain/Role.java
rongallagher: I'll update the pull request with the more functional approach you provided.

apis/chef/src/test/java/org/jclouds/chef/domain/RoleTest.java
rongallagher: This will be fixed in a forthcoming update to this pull request.

apis/chef/src/test/java/org/jclouds/chef/domain/RoleTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+      Role role1 = Role.builder().envRunListElement(env, entry.toUpperCase()).build();
+      Role role2 = Role.builder().envRunListElement(env, entry.toLowerCase()).build();
+
+      assertNotEquals(role1.hashCode(), role2.hashCode(), "hashCodes should not be equal");
+      assertNotEquals(role1, role2, "role1 should not equal role2");
+      assertNotEquals(role2, role1, "role2 should not equal role1");
+   }
+
+   private void verifyRunListForEnvironment(Map<String, List<String>> envRunList, String envName,
```
nacx: Make this public static so you can call it from the live test too and avoid the code duplication?

nacx:
Squashed the commits and pushed to master as [27b3a844](http://git-wip-us.apache.org/repos/asf/jclouds/commit/27b3a844). The merge to 1.9.x is not clean, if you want it to be backported to 1.9.x, mind opening a specific PR to that branch?
Thanks for your first contribution to jclouds!

PR #913, opened 2016-02-03
Set xattr before rename to make blob create atomic

andrewgaul:
Previously concurrent operations could expose an object before put had
completed.  Note that the temporary file is still exposed to the
client.

### DIFF
--- a/apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
+++ b/apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
@@ -456,8 +456,9 @@ public class FilesystemStorageStrategyImpl implements LocalStorageStrategy {
       }
       File outputFile = getFileForBlobKey(containerName, blobKey);
       // TODO: should we use a known suffix to filter these out during list?
-      File tmpFile = getFileForBlobKey(containerName, blobKey + "-" + UUID.randomUUID());
-      Path outputPath = outputFile.toPath();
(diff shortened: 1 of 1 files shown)


andrewgaul:
@kahing Please review.

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -475,20 +476,22 @@ public String putBlob(final String containerName, final Blob blob) throws IOExce
...
-               logger.debug("xattrs not supported on %s", outputPath);
+               logger.debug("xattrs not supported on %s", tmpPath);
             }
          }
-         setBlobAccess(containerName, blobKey, BlobAccess.PRIVATE);
+
+         setBlobAccess(containerName, tmpBlobName, BlobAccess.PRIVATE);
+
+         if (!tmpFile.renameTo(outputFile)) {
+            throw new RuntimeException("Could not rename file " + tmpFile + " to " + outputFile);
```
kahing: if rename fails, tmpFile should be cleaned up

kahing:
:+1:

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
andrewgaul: Existing code cleans up `tmpFile` on line 499.

PR #914, opened 2016-02-05
clean up logic and docs for setting proxies

ahgittin:
In the absence of any special `jclouds.*proxy*` settings, jclouds should respect the normal JVM conventions for proxy detection, such as the `java.net.httpProxyHost` property.  For it to default to no proxy, unlike everything else in java, is surprising.  Previously jclouds only looked for the JVM proxy if `jclouds.use-system-proxies` was set, and then in a cumbersome way:

* `jclouds.use-system-proxies` tried to set `java.net.useSystemProxy` but that is read once at startup and subsequently ignored by the JVM so it had little effect
* so the effect of `jclouds.use-system-proxies` was effectively to tell jclouds to use the default proxy chosen by the JVM looking at the `java.net.*` properties, *not* necessarily the OS proxy referred to by `java.net.useSystemProxy`
* the *default* value of `jclouds.use-system-proxies` was taken from `java.net.useSystemProxy` and so while that succeeds in causing `java.net.useSystemProxy` to take effect, it was succeeding in a strange accidental way:  the only way for the normal `java.net.httpProxyHost` and others to take effect was to set `jclouds.use-system-proxies=true` but make sure `java.net.useSystemProxy=false`

This adds explicit control over whether the JVM default proxy is usable (true by default), and switches the precedence so that if a user specifies a `jclouds.proxy-host` then it is used -- irrespective of whether any other system proxies or jvm proxies are specified.

I think this is more what a user would expect if they do need to set special proxy settings for jclouds to use.  More importantly in most cases they can now just set JVM proxy settings and the right thing will happen.

### DIFF
--- a/core/src/main/java/org/jclouds/Constants.java
+++ b/core/src/main/java/org/jclouds/Constants.java
@@ -100,9 +100,37 @@ public final class Constants {
    /**
     * Boolean property.
     * <p/>
-    * Whether or not to use the proxy setup from the underlying operating system.
+    * Whether or not to attempt to use the proxy setup from the underlying operating system.
+    * Defaults to false. 
+    * Only considered if {@link #PROPERTY_PROXY_FROM_JVM} is false
+    * and {@link #PROPERTY_PROXY_HOST} is not supplied.
+    * Due to how Java's <code>java.net.useSystemProxies</code> is handled,
+    * this may have limited effectiveness.
     */
+   // TODO: should this be deprecated? the impl attempts to set the corresponding JVM system property
+   // but that is documented to have no effect if set after system startup;
+   // see e.g. https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html
    public static final String PROPERTY_PROXY_SYSTEM = "jclouds.use-system-proxy";
+   
+   /**
+    * Boolean property.
+    * <p/>
+    * Whether or not jclouds is permitted to use the default proxy detected by the JVM
+    * and configured there using the usual Java settings:
+    * <ul>
+    * <li> <code>java.net.useSystemProxies</code>
+    * <li> <code>java.net.httpProxyHost</code>
+    * <li> <code>java.net.httpProxyPort</code>
+    * </ul>
+    * <p/> 
+    * Defaults to true so that the Java standard way of setting proxies can be used.
+    * However if {@link #PROPERTY_PROXY_HOST} is set that will always take priority
+    * when jclouds looks for a proxy.
+    * If this property is explicitly set <code>false</code>, 
+    * then jclouds will not use a proxy irrespective of the <code>java.net.*</code> settings,
+    * unless {@link #PROPERTY_PROXY_HOST} is set or {@link #PROPERTY_PROXY_SYSTEM} is true.
+    */
+   public static final String PROPERTY_PROXY_FROM_JVM = "jclouds.use-jvm-proxy";
    
    /**
     * String property.
@@ -132,6 +160,7 @@ public final class Constants {
     * String property.
     * <p/>
     * Explicitly sets the user name credential for proxy authentication.
+    * This only applies when {@link #PROPERTY_PROXY_HOST} is supplied.
     */
    public static final String PROPERTY_PROXY_USER = "jclouds.proxy-user";
    
@@ -139,6 +168,7 @@ public final class Constants {
     * String property.
     * <p/>
     * Explicitly sets the password credential for proxy authentication.
+    * This only applies when {@link #PROPERTY_PROXY_HOST} is supplied.
     */
    public static final String PROPERTY_PROXY_PASSWORD = "jclouds.proxy-password";
 
--- a/core/src/main/java/org/jclouds/Constants.java
+++ b/core/src/main/java/org/jclouds/Constants.java
@@ -102,11 +102,11 @@ public final class Constants {
     * <p/>
     * Whether or not to attempt to use the proxy setup from the underlying operating system.
     * Defaults to false. 
-    * Only considered if {@link #PROPERTY_PROXY_FROM_JVM} is false
+    * Only considered if {@link #PROPERTY_PROXY_ENABLE_JVM_PROXY} is false
     * and {@link #PROPERTY_PROXY_HOST} is not supplied.
     * Due to how Java's <code>java.net.useSystemProxies</code> is handled,
     * this may have limited effectiveness.
-    * @deprecated in 2.0.0, replaced by {@link #PROPERTY_PROXY_FROM_JVM} does what this intended but better
+    * @deprecated in 2.0.0, replaced by {@link #PROPERTY_PROXY_ENABLE_JVM_PROXY} does what this intended but better
     */
    @Deprecated
    // deprecated because:  the impl attempts to set the corresponding JVM system property
@@ -132,7 +132,7 @@ public final class Constants {
     * then jclouds will not use a proxy irrespective of the <code>java.net.*</code> settings,
     * unless {@link #PROPERTY_PROXY_HOST} is set or {@link #PROPERTY_PROXY_SYSTEM} is true.
     */
-   public static final String PROPERTY_PROXY_FROM_JVM = "jclouds.use-jvm-proxy";
+   public static final String PROPERTY_PROXY_ENABLE_JVM_PROXY = "jclouds.enable-jvm-proxy";
    
    /**
     * String property.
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/jclouds/Constants.java
```diff
@@ -100,11 +100,39 @@
...
+    * Whether or not to attempt to use the proxy setup from the underlying operating system.
+    * Defaults to false. 
+    * Only considered if {@link #PROPERTY_PROXY_FROM_JVM} is false
+    * and {@link #PROPERTY_PROXY_HOST} is not supplied.
+    * Due to how Java's <code>java.net.useSystemProxies</code> is handled,
+    * this may have limited effectiveness.
+    */
+   // TODO: should this be deprecated? the impl attempts to set the corresponding JVM system property
+   // but that is documented to have no effect if set after system startup;
+   // see e.g. https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html
```
nacx: +1 to mark it as deprecated.

core/src/test/java/org/jclouds/proxy/ProxyForURITest.java
```diff
@@ -105,47 +112,62 @@ public void testUseProxyForSocketsSettingShouldntAffectHTTP() throws Exception {
...
+      assertNotNull(new ProxyForURI(config).apply(uri));
+   }
+
+   @Test
+   public void testJvmProxyAlwaysPreferredOverSystem() throws URISyntaxException {
+      ProxyConfig test = new MyProxyConfig(true, true, Proxy.Type.HTTP, noHostAndPort, noCreds);
+      ProxyConfig jvm = new MyProxyConfig(false, true, Proxy.Type.HTTP, noHostAndPort, noCreds);
+      URI uri = new URI("http://example.com/file");
+      assertEquals(new ProxyForURI(test).apply(uri), new ProxyForURI(jvm).apply(uri));
+   }
```
nacx: Add a test that uses jvmProxy | systemProxy and also sets the jclouds proxy host? Just to verify that the jclouds proxy takes precedence.

nacx:
Thanks @ahgittin!

ahgittin:
@nacx thx for comments, now addressed.

core/src/main/java/org/jclouds/proxy/ProxyForURI.java
```diff
@@ -65,15 +65,12 @@
...
-         return getLast(proxies);
-      } else if (config.getProxy().isPresent()) {
+      }
+      if (config.getProxy().isPresent()) {
          SocketAddress addr = new InetSocketAddress(config.getProxy().get().getHostText(), config.getProxy().get()
-               .getPort());
+            .getPort());
          Proxy proxy = new Proxy(config.getType(), addr);
-
+
```
demobox: [nit] Remove this?

core/src/test/java/org/jclouds/proxy/ProxyForURITest.java
```diff
@@ -41,12 +42,14 @@
 
    private class MyProxyConfig implements ProxyConfig {
       private boolean useSystem;
+      private boolean canUseProxyFromJvm;
       private Type type;
       private Optional<HostAndPort> proxy;
       private Optional<Credentials> credentials;
 
-      MyProxyConfig(boolean useSystem, Type type, Optional<HostAndPort> proxy, Optional<Credentials> credentials) {
+      MyProxyConfig(boolean useSystem, boolean canUseProxyFromJvm, Type type, Optional<HostAndPort> proxy, Optional<Credentials> credentials) {
```
demobox: [nit] Is there a specific reason to call this `canUse...` rather than simply `useJvm` (in analogy to `useSystem`)?

demobox:
Thanks, @ahgittin! Looks like this makes a bunch more sense, indeed.

Would it make sense to consider backporting this to 1.9.x, but then in such as way that `useSystem` continues to take precedence, so the change is backwards-compatible? Or would that be a lot of work?

/cc @nacx

core/src/main/java/org/jclouds/proxy/ProxyConfig.java
```diff
@@ -33,9 +33,15 @@
...
+    * @deprecated
     * @see org.jclouds.Constants#PROPERTY_PROXY_SYSTEM
-    */
+    */ @Deprecated
    boolean useSystem();
+
+   /**
+    * @see org.jclouds.Constants#PROPERTY_PROXY_FROM_JVM
+    */
+   boolean canUseProxyFromJvm();
```
demobox: [minor] See comment about naming below: any reason not to call use `useJvm` or `useJvm`?

core/src/main/java/org/jclouds/proxy/ProxyForURI.java
```diff
@@ -84,9 +81,20 @@ public PasswordAuthentication getPasswordAuthentication() {
...
          }
          return proxy;
-      } else {
-         return Proxy.NO_PROXY;
       }
+      if (config.canUseProxyFromJvm()) {
+         return getDefaultProxy(endpoint);
+      }
+      if (config.useSystem()) {
+         System.setProperty("java.net.useSystemProxies", "true");
```
demobox: Add a comment here to explain that this actually has little effect in practice? I.e. the same comment as in the PR explanation?

core/src/main/java/org/jclouds/proxy/ProxyForURI.java
ahgittin: should we go to the mailing list to get consensus on whether blank lines should have whitespace to match the prefix on the previous line or the following line or to remove it altogether?  ;)

core/src/test/java/org/jclouds/proxy/ProxyForURITest.java
ahgittin: it isn't whether it *will* use, it is whether it *can* use it.

core/src/main/java/org/jclouds/proxy/ProxyForURI.java
ahgittin: there is a comment on the property in `Constants` to that effect

ahgittin:
thx @demobox.  happy to see this backported if someone wants to.  should be easy, just change the default value of `useSystem`, maybe change order the properties are checked, and udpate the docs.

core/src/main/java/org/jclouds/proxy/ProxyForURI.java
demobox: > should we go to the mailing list to get consensus on whether blank lines

*grin*. I'd say simply go with what is happening between the other declarations in this file?

core/src/main/java/org/jclouds/proxy/ProxyForURI.java
demobox: > there is a comment on the property in Constants to that effect

I was thinking more along the lines of an implementer hint? Something like `// probably does nothing - see comment in Constants...`?

core/src/test/java/org/jclouds/proxy/ProxyForURITest.java
demobox: > it isn't whether it will use, it is whether it can use it.

Ah, then perhaps I'm misunderstanding the code. If this property is set, and no other properties override it, it **will** go through the default logic to pick a proxy? Or do you have to do something else to trigger that?

If setting this to/leaving it at true will cause a proxy to be configured (based on JVM properties) _even in the absence of other commands_, I would be tempted to interpret this is "**will** use the JVM proxy, if possible".

Don't want to make this a big thing, though - also OK with this if we feel it's more accurate.

@nacx Any particular thoughts from you on this?

nacx: Hmmm, this only enables the jvm proxy as a *fallback* if no jclouds proxy properties have been configured (which will be the most common case in jclouds applications given the current logic). I'd say that perhaps `enableJvmProxy` could be more clear, but I'm not the language expert here :)
Not a big deal for me in any case.

ahgittin:
Have gone with `enable` -- comments all addressed I think.

nacx:
lgtm. Thanks! Mind squashing the commits to prepare the merge?
@demobox, good to merge?

core/src/test/java/org/jclouds/proxy/ProxyForURITest.java
```diff
@@ -41,12 +42,14 @@
...
          this.useSystem = useSystem;
+         this.canUseProxyFromJvm = canUseProxyFromJvm;
```
demobox: [minor] Will clean this up to match the naming of the property during the merge, if that's OK with you @ahgittin @nacx...

demobox:
> @demobox, good to merge?

Looks good to me! I'll make one small change as part of the merge just to get the property names consistent, if that's OK with you @ahgittin @nacx?

demobox:
PS: Is there a JIRA issue for sounds? Looks like enough of a functionality change that we'd want that to be documented, if only for the release notes...

demobox:
> PS: Is there a JIRA issue for this?

Created [JCLOUDS-1075](https://issues.apache.org/jira/browse/JCLOUDS-1075)

demobox:
Superseded by #915

PR #916, opened 2016-02-09
Add support for Swift conditional copy

andrewgaul:


apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/ObjectApi.java
```diff
@@ -323,6 +351,7 @@ void copy(@PathParam("destinationObject") String destinationObject,
     *
     * @throws KeyNotFoundException if the source or destination container do not exist.
     */
+   // TODO: CopyOptions
```
andrewgaul: Need to update these two APIs as above.

andrewgaul:
@zack-shoylev Please review.

zack-shoylev:
Looks good, what about the TODOs?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/ObjectApi.java
andrewgaul: Done.

PR #917, opened 2016-02-09
Fix for issue https://issues.apache.org/jira/browse/JCLOUDS-1068.  Th…

oldpeculier:
…is commit fixes the issue where optOSReferenceCode was being selected over operatingSystem.getId(), despite getId returning a good value.  This mis-selection meant that private Softlayer images could not be retrived via jclouds.  With this change the code will now only select optOSReferenceCode if operatingSystem.getId() is empty or null.


oldpeculier:
Closed.  Failed regression testing.

PR #919, opened 2016-02-13
JCLOUDS-651: Conditional copy blob

andrewgaul:
This pull request changes the `CopyOptions` interface which is annotated as `@Beta`.

### DIFF
--- a/blobstore/src/main/java/org/jclouds/blobstore/options/CopyOptions.java
+++ b/blobstore/src/main/java/org/jclouds/blobstore/options/CopyOptions.java
@@ -17,6 +17,7 @@
 
 package org.jclouds.blobstore.options;
 
+import java.util.Date;
 import java.util.Map;
 
 import org.jclouds.io.ContentMetadata;
@@ -24,6 +25,7 @@ import org.jclouds.javax.annotation.Nullable;
 
 import com.google.auto.value.AutoValue;
 import com.google.common.annotations.Beta;
+import com.google.common.collect.ImmutableMap;
 
 @AutoValue
 @Beta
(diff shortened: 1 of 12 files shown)


andrewgaul:
@zack-shoylev Please review.

blobstore/src/main/java/org/jclouds/blobstore/options/CopyOptions.java
```diff
@@ -17,59 +17,48 @@
...
+   public abstract Date ifUnmodifiedSince();
+   @Nullable
+   public abstract String ifMatch();
+   @Nullable
+   public abstract String ifNoneMatch();
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+      public abstract Builder contentMetadata(ContentMetadata contentMetadata);
+      public abstract Builder userMetadata(Map<String, String> userMetadata);
```
zack-shoylev: You still need ImmutableMap.copyOf, so you will have to use autoBuild, like this:

https://github.com/jclouds/jclouds-labs-openstack/blob/master/openstack-heat/src/main/java/org/jclouds/openstack/heat/v1/options/UpdateStack.java

zack-shoylev:
One note, but otherwise looks good.

blobstore/src/main/java/org/jclouds/blobstore/options/CopyOptions.java
andrewgaul: Done.  Any idea how to make the getters for `Date` objects call `clone`?

zack-shoylev: I would suggest using String instead of Date :) Or use an immutable Date class or wrapper, maybe. Basically, don't use Date here if at all possible.

andrewgaul: I could not find an low effort solution and will punt on this for now.  We can revisit immutable `Date` when we move to Java 8 and can use `Instant`.

zack-shoylev: Sounds good.

PR #921, opened 2016-02-17
[JCLOUDS-1079] Make Floating IPs public in the NodeMetadata in openstack-nova provider

kwart:
Floating IPs from private address blocks were not available as public address after refreshing NodeMetadata.

### DIFF
--- a/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/ServerInRegionToNodeMetadata.java
+++ b/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/ServerInRegionToNodeMetadata.java
@@ -18,7 +18,9 @@ package org.jclouds.openstack.nova.v2_0.compute.functions;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.base.Predicates.and;
 import static com.google.common.base.Predicates.not;
+import static com.google.common.base.Predicates.or;
 import static com.google.common.collect.Iterables.filter;
 import static com.google.common.collect.Iterables.find;
 import static com.google.common.collect.Iterables.transform;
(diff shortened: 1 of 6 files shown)


apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/ServerInRegionToNodeMetadata.java
```diff
@@ -146,6 +147,11 @@ public NodeMetadata apply(ServerInRegion serverInRegion) {
 
    public static final Predicate<Address> isPrivateAddress = new Predicate<Address>() {
```
nacx: Rename the predicate to make it clean that it looks for private IPs that are not floating (or leave it as-is and create a floating ip predicate and combine both to classify the addresses). Otherwise this can cause confusion when reading the code.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/domain/Address.java
```diff
@@ -121,12 +154,14 @@ public boolean equals(Object obj) {
...
-               && Objects.equal(this.version, that.version);
+               && Objects.equal(this.version, that.version)
+               && Objects.equal(this.macAddr, that.macAddr)
+               && Objects.equal(this.type, that.type);
    }
    
    protected ToStringHelper string() {
       return Objects.toStringHelper(this)
-            .add("addr", addr).add("version", version);
+            .add("addr", addr).add("version", version).add("macAddr", macAddr).add("type", type);
```
nacx: Use `macAddr.or("")` and `type.or("")` (or another default), to have a better string representation?

nacx:
Change lgtm. Thanks @kwart! Waiting for the build...
@zack-shoylev wanna have a look at this too?

nacx:
Test failure was unrelated to this change.
Pushed to master as [e0ab5d84](http://git-wip-us.apache.org/repos/asf/jclouds/commit/e0ab5d84). Thanks @kwart !

PR #922, opened 2016-02-18
Ensures the Swift container list call requests JSON as per spec

zack-shoylev:


zack-shoylev:
https://issues.apache.org/jira/browse/JCLOUDS-1080

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/ContainerApi.java
```diff
@@ -77,6 +78,7 @@
    @Named("container:list")
    @GET
    @Fallback(EmptyFluentIterableOnNotFoundOr404.class)
+   @QueryParams(keys = {"format"}, values = {"json"})
```
andrewgaul: Don't need an array for single values.

andrewgaul:
:+1: Can we confirm that other clients also send this header?

zack-shoylev:
Other clients?

andrewgaul:
python-swiftclient or similar?

zack-shoylev:
swift on ubuntu:

DEBUG:swiftclient:REQ: curl -i https://storage101.ord1.clouddrive.com/v1/MossoCloudFS_0d629063-aa0f-4bf9-9e04-62e1fd6aaa2c?format=json -X GET -H "X-Auth-Token:

Good to get confirmation. It's in the spec and other clients use it. Still having some issues with the path test though.

zack-shoylev:
merged

PR #923, opened 2016-02-19
Swift container JSON requests and null checks

IgnacioCalvo:
1) Swift container and JSON format for requests
Regarding the following pull request by @zack-shoylev (thanks, good job) : 
https://github.com/jclouds/jclouds/pull/922
Related to this ticket : 
https://issues.apache.org/jira/browse/JCLOUDS-1080
I think that is also needed for all the @GET methods in ContainerApi.java, ObjectApi.java and StaticLargeObjectApi.java

2) Prevent NumberFormatException and NullPointerException when parsing Swift's response
We need to prevent NumberFormatException and NullPointerException as Swift will not give a value for bytesUsed, objectCount, date, etag and name.

### DIFF
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/StaticLargeObjectApi.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/StaticLargeObjectApi.java
@@ -128,6 +128,6 @@ public interface StaticLargeObjectApi {
    @Named("staticLargeObject:getManifest")
    @GET
    @Fallback(EmptyListOnNotFoundOr404.class)
-   @QueryParams(keys = "multipart-manifest", values = "get")
+   @QueryParams(keys = {"format", "multipart-manifest"}, values = {"json", "get"})
    List<Segment> getManifest(@PathParam("objectName") String objectName);
 }
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/functions/ParseContainerFromHeaders.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/functions/ParseContainerFromHeaders.java
@@ -36,11 +36,13 @@ public class ParseContainerFromHeaders implements Function<HttpResponse, Contain
 
    @Override
    public Container apply(HttpResponse from) {
+      String bytesUsed = from.getFirstHeaderOrNull(CONTAINER_BYTES_USED);
+      String objectCount = from.getFirstHeaderOrNull(CONTAINER_OBJECT_COUNT);
       Container c = 
       Container.builder()
             .name(name)
-            .bytesUsed(Long.parseLong(from.getFirstHeaderOrNull(CONTAINER_BYTES_USED)))
-            .objectCount(Long.parseLong(from.getFirstHeaderOrNull(CONTAINER_OBJECT_COUNT)))
+            .bytesUsed(bytesUsed != null ? Long.valueOf(bytesUsed) : null)
+            .objectCount(objectCount != null ? Long.valueOf(objectCount) : null)
             .anybodyRead(CONTAINER_ACL_ANYBODY_READ.equals(from.getFirstHeaderOrNull(CONTAINER_READ)))
             .metadata(EntriesWithoutMetaPrefix.INSTANCE.apply(from.getHeaders())).build();
       return c;
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/functions/ParseObjectListFromResponse.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/functions/ParseObjectListFromResponse.java
@@ -36,6 +36,7 @@ import org.jclouds.rest.InvocationContext;
 import org.jclouds.rest.internal.GeneratedHttpRequest;
 
 import com.google.common.base.Function;
+import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.google.common.hash.HashCode;
 import com.google.common.io.ByteSource;
@@ -91,12 +92,15 @@ public class ParseObjectListFromResponse implements Function<HttpResponse, Objec
                   .payload(payload(input.bytes, input.hash, "application/directory", input.expires))
                   .lastModified(new Date(0)).build();
          }
+         String name = Strings.nullToEmpty(input.name);
+         String etag = Strings.nullToEmpty(input.hash);
+         Date lastModified = input.last_modified == null ? new Date() : input.last_modified;
          return SwiftObject.builder()
                .uri(uriBuilder(containerUri).clearQuery().appendPath(input.name).build())
-               .name(input.name)
-               .etag(input.hash)
-               .payload(payload(input.bytes, input.hash, input.content_type, input.expires))
-               .lastModified(input.last_modified).build();
+               .name(name)
+               .etag(etag)
+               .payload(payload(input.bytes, etag, input.content_type, input.expires))
+               .lastModified(lastModified).build();
       }
    }
 
--- a/apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/BulkApiLiveTest.java
+++ b/apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/BulkApiLiveTest.java
@@ -69,7 +69,7 @@ public class BulkApiLiveTest extends BaseSwiftApiLiveTest<SwiftApi> {
                                                      .extractArchive(containerName, payload, "tar.gz");
          assertEquals(extractResponse.getCreated(), OBJECT_COUNT);
          assertTrue(extractResponse.getErrors().isEmpty());
-         assertEquals(api.getContainerApi(regionId).get(containerName).getObjectCount(), OBJECT_COUNT);
+         assertEquals(api.getContainerApi(regionId).get(containerName).getObjectCount(), Long.valueOf(OBJECT_COUNT));
 
          // repeat the command
          extractResponse = api.getBulkApi(regionId).extractArchive(containerName, payload, "tar.gz");
@@ -85,7 +85,7 @@ public class BulkApiLiveTest extends BaseSwiftApiLiveTest<SwiftApi> {
          assertEquals(deleteResponse.getDeleted(), OBJECT_COUNT);
          assertEquals(deleteResponse.getNotFound(), 0);
          assertTrue(deleteResponse.getErrors().isEmpty());
-         assertEquals(api.getContainerApi(regionId).get(containerName).getObjectCount(), 0);
+         assertEquals(api.getContainerApi(regionId).get(containerName).getObjectCount(), Long.valueOf(0));
       }
    }
 
--- a/apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ContainerApiMockTest.java
+++ b/apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ContainerApiMockTest.java
@@ -58,12 +58,12 @@ public class ContainerApiMockTest extends BaseOpenStackMockTest<SwiftApi> {
          assertEquals(containers, ImmutableList.of(
                Container.builder()
                      .name("test_container_1")
-                     .objectCount(2)
-                     .bytesUsed(78).build(),
+                     .objectCount(2L)
+                     .bytesUsed(78L).build(),
                Container.builder()
                      .name("test_container_2")
-                     .objectCount(1)
-                     .bytesUsed(17).build()));
+                     .objectCount(1L)
+                     .bytesUsed(17L).build()));
 
          assertEquals(server.getRequestCount(), 2);
          assertAuthentication(server);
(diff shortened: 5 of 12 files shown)


apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/StaticLargeObjectApi.java
```diff
@@ -128,6 +128,7 @@ String replaceManifest(@PathParam("objectName") String objectName,
    @Named("staticLargeObject:getManifest")
    @GET
    @Fallback(EmptyListOnNotFoundOr404.class)
+   @QueryParams(keys = "format", values = "json")
    @QueryParams(keys = "multipart-manifest", values = "get")
```
nacx: Better use just one annotation: `@QueryParams(keys = {"format", "multipart-manifest"}, values = {"json", "get"})`

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/functions/ParseContainerFromHeaders.java
```diff
@@ -36,11 +36,13 @@
 
    @Override
    public Container apply(HttpResponse from) {
+      String bytesUsed = from.getFirstHeaderOrNull(CONTAINER_BYTES_USED);
+	String objectCount = from.getFirstHeaderOrNull(CONTAINER_OBJECT_COUNT);
```
nacx: [minor] formatting

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/functions/ParseContainerFromHeaders.java
```diff
@@ -36,11 +36,13 @@
...
       Container c = 
       Container.builder()
             .name(name)
-            .bytesUsed(Long.parseLong(from.getFirstHeaderOrNull(CONTAINER_BYTES_USED)))
-            .objectCount(Long.parseLong(from.getFirstHeaderOrNull(CONTAINER_OBJECT_COUNT)))
+            .bytesUsed(bytesUsed==null?0:Long.parseLong(bytesUsed))
+            .objectCount(objectCount==null?0:Long.parseLong(objectCount))
```
nacx: Our checkstyle checks might complain about this. Use spaces: `bytesUsed == null ? 0 : `

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/functions/ParseObjectListFromResponse.java
```diff
@@ -83,20 +83,23 @@ public ObjectList apply(HttpResponse from) {
 
       @Override
       public SwiftObject apply(InternalObject input) {
+         String name = input.name==null?"":input.name;
```
nacx: Use the more idiomatic `Strings.nullToEmpty(input.name)`?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/functions/ParseContainerFromHeaders.java
nacx: When the swift server doesn't return these values, is 0 the *real* value? Would it be more correct to update the container object variables to be `Long` and accept null values?

nacx:
Thanks @IgnacioCalvo!

nacx:
The `ContainerApiMockTest`, `ObjectApiMockTest` and `StaticLargeObjectApiMockTest` tests also need to be modified to properly verify the format parameter is being sent.

IgnacioCalvo:
@nacx Hi "tocayo" :)
Thanks for your detailed and fast response. I understand good coding is necessary. I'm doing my best, as contributing to open source JClouds and Github are new to me. 
I'm trying to amend the changes and fix the build. Hopefully in a while it will be a bit better.

nacx:
> Hi "tocayo" :)

:D

Your contribution is very welcome! Just ping us if you need help!

IgnacioCalvo:
I agree that objectCount and bytesUsed should probably be Long and (instead of long). I can see that it is referenced by several unit tests and classes which I will fix soon (can't finish today, I'm afraid). I guess many existing applications may already rely on this definition. I'm wondering if this would be an issue or we are happy that they need to change their code.
Thanks again.

zack-shoylev:
Is there a particular openstack doc that describes these are not required to be set? Would be good to documents this for this PR.

As for releases, it's fairly easy to use the master snapshot in-between releases, which we try to do every 4-5 weeks.

zack-shoylev:
> I think that is also needed for all the @GET methods in ContainerApi.java, ObjectApi.java and StaticLargeObjectApi.java

I could only find the format=json in the spec for Container list as described in the original issue... I am not sure if it applies to everything - it might.

IgnacioCalvo:
From the object storage API documentation : 
http://developer.openstack.org/api-ref-objectstorage-v1.html
We can see that the format=json is specified either as a parameter or as an example at least for the following : 
Show account details and list containers
Show container details and list objects

From the spec I can't give evidence that those fields may be optional. I think the issue here is that I'm connecting to Openstack Swift which effectively it is middleware for a Ceph instance.  
http://docs.ceph.com/docs/master/radosgw/swift/
"Ceph supports a RESTful API that is compatible with the basic data access model of the Swift API."
In the documentation, we can't see a clear spec of the headers of the response :
http://docs.ceph.com/docs/master/radosgw/swift/containerops/

But looking at Ceph's tracker I can see tickets were they report the following :
http://tracker.ceph.com/issues/10666
RGW swift API: Response header of GET request for container does not contain X-Container-Object-Count, X-Container-Bytes-Used and x-trans-id headers 
http://tracker.ceph.com/issues/10684
RGW swift API: Response returned by GET request on account does not contain few headers.

I wonder when we'll get that patch. Meanwhile, I posted a request and the response header looks like this : 

```
HTTP/1.1 200 OK
Content-type: application/json; charset=utf-8
Content-Length: 851
Date: Fri, 19 Feb 2016 23:21:34 GMT
Connection: Keep-Alive
```

Those are my 2 cents.

nacx:
The code itself lgtm, but I'n not the OpenStack expert here to evaluate if the functional change to nullable values is ok. @zack-shoylev wdyt?

zack-shoylev:
The only problem is backwards compatibility, but as long as we don't backport, I think this is a good change. We have been moving in the direction of using more boxed types already, especially with AutoValue making it easier to do so.

andrewgaul:
As I understand it, our backwards compatibility goal avoids changing a public, non-beta, non-labs API without deprecating it in the previous major release.  One advantage here is that we retain *source* compatibility due to boxing although break *binary* compatibility.  Given this I prefer to merge this change.

Do we have our compatibility goals documented anywhere other than mailing lists?

nacx:
>Do we have our compatibility goals documented anywhere other than mailing lists?

We have this page (but don't know if it accurate or up to date): https://cwiki.apache.org/confluence/display/JCLOUDS/Deprecation+and+Beta+Policy

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/BulkApiLiveTest.java
```diff
@@ -69,7 +69,7 @@ public void testExtractArchive() throws Exception {
                                                      .extractArchive(containerName, payload, "tar.gz");
          assertEquals(extractResponse.getCreated(), OBJECT_COUNT);
          assertTrue(extractResponse.getErrors().isEmpty());
-         assertEquals(api.getContainerApi(regionId).get(containerName).getObjectCount(), OBJECT_COUNT);
+         assertEquals(api.getContainerApi(regionId).get(containerName).getObjectCount(), Long.valueOf(OBJECT_COUNT));
```
zack-shoylev: Is this needed (here and elsewhere)? If you let OBJECT_COUNT be autoboxed, the test spec will reflect the backwards compatibility of the change, as least in source, as was mentioned. Thoughts?

IgnacioCalvo: I´m not completely sure about what you mean. At the very beginning I went for the easy way and kept the object count as long, and make it 0 when null. nacx was probably right saying that would not be correct so I made it Long. 
In the tests, sometimes constant OBJECT_COUNT is compared with extractResponse.getCreated() , and therefore is a long, and sometimes it compared with getObjectCount() and therefore we must convert it to Long. 
I wonder if your concern is about object count long vs Long , or about how we deal with constants in tests, as they sometimes are compared with long and with Long. 
Thanks again.

zack-shoylev: I left my comment at the wrong line. Sorry! There were a few Long.valueOf() such as 

> .objectCount(Long.valueOf(2))

that I think are not necessary, as long would be converted to Long automatically in that case.

IgnacioCalvo:
Hi zack-shoylev, 
Probably I could have inferred it from the context, but I missed it. 
At the beginning I tried :
`.objectCount(2)`
but I would get an error : 
> The method objectCount(Long) in the type Container.Builder is not applicable for the arguments (int)
I didn't realize at the moment it would be enough with :  
`.objectCount(2L)`
I will ammend that now, and update it this way.

zack-shoylev:
Alright. If @andrewgaul has no other notes, let's rebase to a single commit and get ready to merge?

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ContainerApiMockTest.java
```diff
@@ -116,8 +116,8 @@ public void testContainerExists() throws Exception {
 
          Container container = api.getContainerApi("DFW").get("myContainer");
          assertEquals(container.getName(), "myContainer");
-         assertEquals(container.getObjectCount(), 42l);
-         assertEquals(container.getBytesUsed(), 323479l);
+         assertEquals(container.getObjectCount(), Long.valueOf(42l));
+         assertEquals(container.getBytesUsed(), Long.valueOf(323479l));
```
andrewgaul: Please use upper-case L to avoid confusion between lower-case l and number 1.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/BulkApiLiveTest.java
```diff
@@ -85,7 +85,7 @@ public void testBulkDelete() throws Exception {
          assertEquals(deleteResponse.getDeleted(), OBJECT_COUNT);
          assertEquals(deleteResponse.getNotFound(), 0);
          assertTrue(deleteResponse.getErrors().isEmpty());
-         assertEquals(api.getContainerApi(regionId).get(containerName).getObjectCount(), 0);
+         assertEquals(api.getContainerApi(regionId).get(containerName).getObjectCount(), Long.valueOf(0));
```
andrewgaul: Unneeded `Long.valueOf`.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ContainerApiMockTest.java
```diff
@@ -236,8 +236,8 @@ public void testGetKnowingServerMessesWithMetadataKeyCaseFormat() throws Excepti
          SwiftApi api = api(server.getUrl("/").toString(), "openstack-swift");
          Container container = api.getContainerApi("DFW").get("myContainer");
          assertEquals(container.getName(), "myContainer");
-         assertEquals(container.getObjectCount(), 42l);
-         assertEquals(container.getBytesUsed(), 323479l);
+         assertEquals(container.getObjectCount(), Long.valueOf(42l));
+         assertEquals(container.getBytesUsed(), Long.valueOf(323479l));
```
andrewgaul: Why do we need this change?

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/StaticLargeObjectApiLiveTest.java
```diff
@@ -99,7 +99,7 @@ public void testDelete() throws Exception {
          assertThat(resp.deleted()).isEqualTo(3);
          assertThat(resp.notFound()).isZero();
          assertThat(resp.errors()).isEmpty();
-         assertEquals(api.getContainerApi(regionId).get(containerName).getObjectCount(), 0);
+         assertEquals(api.getContainerApi(regionId).get(containerName).getObjectCount(), Long.valueOf(0));
```
andrewgaul: Use `1L` for implicit conversion to `Long`.

andrewgaul:
Please address nits and squash to a single commit as per @zack-shoylev suggestion.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/ContainerApiMockTest.java
IgnacioCalvo: Regarding the particular cases of assertEquals and Long.valueOf, if I go with : 

`assertEquals(container.getObjectCount(), Long.valueOf(42l));`

I'll get : 

`The method assertEquals(Object, Object) is ambiguous for the type ... `

http://stackoverflow.com/questions/1811103/java-junit-the-method-x-is-ambiguous-for-type-y

> What this error means is that you're passing a double and and Double into a method that has two different signatures: assertEquals(Object, Object) and assertEquals(double, double) both of which could be called, thanks to autoboxing.

IgnacioCalvo:
Hi all, 
My apologies if I did something wrong with the squash. I can see that for some funny reason my local and Github users got mixed, I guess that's ok. As I said, I'm new both to Git and Github.
@andrewgaul asked a few days ago why we needed those `Long.valueOf`and I replied with that reference to http://stackoverflow.com/questions/1811103/java-junit-the-method-x-is-ambiguous-for-type-y 
To be fair, I didn't chase on this because I was too busy with something else (sorry again). 
In summary, squash applied, hopefully ok, but no action taken on those `Long.valueOf` as that would lead to a compilation error.
No rush, but just to manage expectations on my side, can I ask what would the target date is for an eventual release? I see I am committing to 2.0.0-SNAPSHOT and I understand releases may still happen from 1.9.x. I came across the idea of merging this into 1.9.x too, but probably it's a bad one, as we are changing an attribute's data type and we may not want this for a micro release. 
Thanks, guys.

andrewgaul:
I squashed the commits, rebased, reworded the commit message, and pushed to master as 7d2f1e8b4d83e7ed2b520e635acdb6ef2cf561d7.  We do not plan further 1.9.x releases so let's skip backporting.  Thank you for your contribution @IgnacioCalvo!

PR #926, opened 2016-02-19
Add alias fallback info to keystone and nova extensions.

zack-shoylev:


apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/extensions/RoleAdminApi.java
```diff
@@ -48,7 +48,7 @@
  *
  */
 @Beta
-@Extension(of = ServiceType.IDENTITY, namespace = ExtensionNamespaces.OS_KSADM)
+@Extension(of = ServiceType.IDENTITY, namespace = ExtensionNamespaces.OS_KSADM, name = "OpenStack Keystone Admin", alias = "OS-KSADM")
```
nacx: Out of curiosity. Why are these values not used as a constant, like the rest?

zack-shoylev: Keystone only has two extensions, and I wasn't sure how I wanted these implemented. Considering how nova is, I will probably add another file and refactor these to constants.

zack-shoylev:
merged

appsdesh:
Hello!
Looks like your commit helped me in solving following issue
http://www.mail-archive.com/dev@jclouds.apache.org/msg06427.html

I was wondering if you know what changed in Keystone which broke jclouds 1.8 and needed your patch?

PR #928, opened 2016-02-25
Properly handler ProfitBricks service errors

nacx:
The current implementation of the error handler just propagated exceptions for the expected response codes, but ignored all other failed responses. This ended up in failed requests to be treated as successful ones, and errors appeared later and without the proper context. This PR fixes that.

Also, ProfitBricks returns a `javax.ejb.EJBException` when calling `getImage` from an account that has permissions on that image but does not own it. it is likely to be a bug on their side, but we can properly handle that, so the compute service adapter has been changed to the this into account.

/cc @devcsrj


devcsrj:
+1. Thanks for the fix @nacx ! :)

nacx:
All live tests passing with this last fix.

nacx:
Squashed and pushed to master as [72f2652d]( http://git-wip-us.apache.org/repos/asf/jclouds/commit/72f2652d).

PR #930, opened 2016-03-07
Fixes an intermittently failing swift live test.

zack-shoylev:


apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/internal/BaseSwiftApiLiveTest.java
```diff
@@ -48,6 +48,10 @@ protected BaseSwiftApiLiveTest() {
       provider = "openstack-swift";
    }
 
+   protected void awaitConsistency() {
+      Uninterruptibles.sleepUninterruptibly(5, TimeUnit.SECONDS);
```
andrewgaul: Should this use the `test.blobstore.await-consistency-timeout-seconds` property as used by `BaseBlobStoreIntegrationTest`?

andrewgaul:
@zack-shoylev Relevant, S3Proxy now models eventual consistency: https://github.com/andrewgaul/s3proxy/commit/6292d28c9c66e8f432ae5b18acbadd0c191bc05d.  Easy to add this to [SwiftProxy](https://github.com/bouncestorage/swiftproxy).

zack-shoylev:
Thanks @andrewgaul ! I will refactor this a bit, but there are some other swift tests to fix after this, I think. Adding this to swiftproxy is also a good idea - I'll try to look into it some more.

zack-shoylev:
merged

PR #931, opened 2016-03-08
fix for issue if SL machine has multiple credentials registered

ahgittin:
now just pick the best one. it matters only when we are going to log in to a machine.
the only time the problem has been observed has been with pre-existing machines set up
outwith jclouds with multiple password.

PR #932, opened 2016-03-08
fix for issue if SL machine has multiple credentials registered

ahgittin:
now just pick the best one. it matters only when we are going to log in to a machine.
the only time the problem has been observed has been with pre-existing machines set up
outwith jclouds with multiple password.

**backport of #931**


andreaturli:
merged at [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/0248931c)
thanks @ahgittin

PR #934, opened 2016-03-09
Version range specification for guava in OSGi

CMoH:
This PR changes jclouds' OSGi import directive for guava to allow version range [16.x-19.0]. This helps dependent projects mix jclouds with bundles using other guava versions within OSGi environments.

According to https://issues.apache.org/jira/browse/JCLOUDS-622 jclouds was made to use only guava classes that are compatible throughout guava versions, so this should be safe for the time being.

What are your thoughs on this? Is there better way to solve this package resolution problem?

P.S. The changes are based on the 1.9.x branch, because that's the version we use right now. I can port them to master if needed.


CMoH:
For reference, here is an analysis of guava's API compatibility across versions: http://upstream.rosalinux.ru/java/versions/guava.html

neykov:
+1, looks like this has already been discussed at https://issues.apache.org/jira/browse/JCLOUDS-534

scriptbuilder/pom.xml
```diff
@@ -83,6 +83,6 @@
         </configuration>
       </plugin>
     </plugins>
-  </build>
+  </build>-->
```
nacx: I'm not an OSGi expert, but could you please explain why this should be commented?

project/pom.xml
```diff
@@ -206,6 +206,7 @@
     <maven.site.url.base>gitsite:git@github.com/jclouds/jclouds-maven-site.git</maven.site.url.base>
     <clojure.version>1.3.0</clojure.version>
     <guava.version>16.0.1</guava.version>
+    <guava.osgi.import>com.google.common.*;version="[16.0.1,19.0.0]"</guava.osgi.import>
```
andrewgaul: Should this be half-closed notation, e.g., [16.0.1,20.0.0) ?

nacx: FTR, these are the Guava versions we are compatible with (failed builds are due to transient issues):
https://jclouds.ci.cloudbees.com/job/jclouds-guava-guice-compat/

CMoH: @andrewgaul Yeah, that is better. I'll update the PR.

scriptbuilder/pom.xml
CMoH: I added the guava version range override in `jclouds-project`, but `jclouds-scriptbuilder` didn't get them because it had these settings for `maven-bundle-plugin`. It appears to me that these are unnecessary, since they are exactly the same as in `jclouds-project` - maybe they were left here by accident?

At first I commented them out, but I think I'll simply remove them when updating the PR.

nacx:
Looks like an unrelated failure. Let's trigger the build again: rebuild please

nacx:
Just one last sanity check now that the 1.9.x branch build is back to green: rebuild, please

nacx:
Just one last sanity check now that the 1.9.x branch build is back to green: rebuild please

nacx:
Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/9a20cada) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/fc5f5e50). Thanks @CMoH!

PR #935, opened 2016-03-10
Ensures parallel tests don't step on each other in Static Large Objec…

zack-shoylev:
…t swift testing


apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/StaticLargeObjectApiLiveTest.java
```diff
@@ -43,11 +46,23 @@
 public class StaticLargeObjectApiLiveTest extends BaseSwiftApiLiveTest<SwiftApi> {
 
    private String name = getClass().getSimpleName();
-   private String containerName = getClass().getSimpleName() + "Container";
    private byte[] megOf1s;
    private byte[] megOf2s;
 
+   private static Queue<String> containerCleanup = new ConcurrentLinkedQueue<String>();
+
+   private String getContainer() {
```
andrewgaul: Instead of using a custom setup and teardown, could you just change `BaseSwiftApiLiveTest` to inherit from `BaseBlobStoreIntegrationTest` and use its methods?  `S3ClientLiveTest` uses this common code for an example.

PR #938, opened 2016-03-20
JCLOUDS-1095: Update OpenStack KeyStone Support for Zaqar

fribeiro1:


nacx:
This change *replaces* the current queuing service with the latest one, which is likely to be a backwards incompatible change. Does it make sense to also keep the marconi service type and just add zaqar? Would that make jclouds work with both versions?

fribeiro1:
I don't really think we need to keep the service type for Marconi, which wasn't included in any distribution before Liberty, for example, but the 1.0 API should continue to be supported as long as the Zaqar project maintains it.

nacx:
@zack-shoylev WDYT?

zack-shoylev:
@nacx I think it has to be "deprecated" (or documented as such) before it is removed in a major release, like everything else.

nacx:
Hmmm now that I see both PRs together, it looks like this change is needed to change the *default* value for the service type in the Marconi API, as done in https://github.com/jclouds/jclouds-labs-openstack/pull/204.
I agree to keep the current constant and mark it as deprecated, but also keep the PR that changes the default service type in Marconi. In the end code is compatible; we're just changing the endpoint that would be picked? And also that change really affects just the labs API, so it should be OK.
Dos this look like the way to go?

fribeiro1:
I think so, and just amended the commit, please review.

nacx:
Pushed to master as [92e6319c](http://git-wip-us.apache.org/repos/asf/jclouds/commit/92e6319c). Thanks @fribeiro1!

PR #942, opened 2016-03-30
JCLOUDS-1102: Rackspace returns a new structure for volume types

opavlov24:
https://issues.apache.org/jira/browse/JCLOUDS-1102


apis/openstack-cinder/src/main/java/org/jclouds/openstack/cinder/v1/domain/VolumeType.java
```diff
@@ -126,12 +125,12 @@ protected ConcreteBuilder self() {
...
       "id", "name", "created_at", "updated_at", "extra_specs"
    })
-   protected VolumeType(String id, String name, @Nullable Date created, @Nullable Date updated, Map<String, String> extraSpecs) {
+   protected VolumeType(String id, String name, @Nullable Date created, @Nullable Date updated, @Nullable Map<String, String> extraSpecs) {
       this.id = checkNotNull(id, "id");
       this.name = checkNotNull(name, "name");
       this.created = Optional.fromNullable(created);
       this.updated = Optional.fromNullable(updated);
-      this.extraSpecs = ImmutableMap.copyOf(checkNotNull(extraSpecs, "extraSpecs"));     
+      this.extraSpecs = extraSpecs == null ? new HashMap<String, String>() : ImmutableMap.copyOf(extraSpecs);
```
nacx: We want immutable collections, so better use:
```java
this.extraSpecs = extraSpecs == null ? ImmutableMap.of() : ImmutableMap.copyOf(extraSpecs);
```

nacx:
Thanks for your first contribution @opavlov24!

Could you also add a couple elements to the [volume_type_list_simple.json](https://github.com/jclouds/jclouds/blob/master/apis/openstack-cinder/src/test/resources/volume_type_list_simple.json) object, one with an empty `extra_specs` and another with a `null` one? This file is used in the [VolumeTypeApiExpectTest](https://github.com/jclouds/jclouds/blob/master/apis/openstack-cinder/src/test/java/org/jclouds/openstack/cinder/v1/features/VolumeTypeApiExpectTest.java#L41-L52) to verify the api method behaves as expected and generates the expected request and can deserialize the response. That test will tell us the patch actually works and also make sure we won't break that accidentally in the future.

nacx:
Pushed to master as [74a2a268](http://git-wip-us.apache.org/repos/asf/jclouds/commit/74a2a268). Thanks @opavlov24!

PR #943, opened 2016-04-06
Fixed OAuth authentication flow injections

nacx:
Changes how OAuth authentication flows are injected. Since different flows require different context properties, we can't eagerly create all flow classes because users will only bind to the context the properties relevant to the authentication for the current credential type.

This PR changes the current approach of declaring the properties as optional in each flow to let the OAuthModule lazily load only the right flow. This way we eliminate the need for optional injections and allow every flow to properly configure their properties.

It also provides an OAuthScopes implementation to properly designate the absence of scopes.

/cc @jmspring


nacx:
Squashed and pushed to master as [819141a6](http://git-wip-us.apache.org/repos/asf/jclouds/commit/819141a6).

PR #946, opened 2016-04-17
JCLOUDS-1101: Correct generics wildcard errors

andrewgaul:
Always incorrect but previously broken on Java 9.  Reference:

https://bugs.openjdk.java.net/browse/JDK-8075793


nacx:
+1. can we set up a Java 9 build?

PR #947, opened 2016-04-21
JCLOUDS-1105: Do not create the default security group if custom groups are specified

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1105

I'm still not sure about the impact of this change. I'd rather play a bit with this and get feedback from users. I haven't been able to test it live yet, so this PR just relies on the unit tests. I plan to do some testing later today but I'd love to get some feedback from others too.

/cc @andreaturli @aledsage @andras-nemes


andras-nemes:
Thanks for your quick reaction and feedback. I think it the code looks good.
//Andras

andreaturli:
@nacx code looks sensible +1

I will keep this PR in mind if we need to revert/change the implementation later on, ok?

Thanks!

marcocast:
Hi Guys,

any idea on when this patch will be merged? We have the same problem and we would love to have the fix as soon as possible. 

Thanks!!

nacx:
@marcocast Expect it to be merged this week. I running the live test in my backlog. I'll try to run them today.
In any case, if you can run the live tests and share the results in a gist or pastie, that would be great :) There are some that are already failing, but I just want to run the entire suite and verify this change is not breaking anything that is working now.

marcocast:
@nacx Thanks man.

nacx:
Pushed to master as [bde70457](http://git-wip-us.apache.org/repos/asf/jclouds/commit/bde70457). Thanks everyone!

PR #948, opened 2016-04-21
JCLOUDS-1104: Extension namespaces are deprecated and can be null

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1104

If the namespace of an extension is not present, then try to match them just by name and alias.
@zack-shoylev I'd pretty much appreciate your comments :)


nacx:
Pushed to master as [04f1bb2b](http://git-wip-us.apache.org/repos/asf/jclouds/commit/04f1bb2b).

PR #949, opened 2016-04-26
GCE ResetWindowsPassword

bostko:
- Function which tells to GCE to reset Windows password

- InstanceApiLiveTest for windows

- Unit test for decrypting windows password

- fixed compile errors


bostko:
This PR is already made for 1.9.x in https://github.com/jclouds/jclouds-labs-google/pull/170

I had to comment out GoogleComputeEngineServiceAdapter.java:145 to compile against master.

PR #951, opened 2016-05-03
[JCLOUDS-1108]: shorten hardwareId

andreaturli:
This PR fixes https://issues.apache.org/jira/browse/JCLOUDS-1108

### DIFF
--- a/providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
+++ b/providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
@@ -147,15 +147,15 @@ public final class GoogleComputeEngineServiceAdapter
       Scheduling scheduling = getScheduling(options);
 
       NewInstance newInstance = new NewInstance.Builder(
-              name,
-              hardwareUri, // machineType
-              network,
-              disks)
-              .description(group)
-              .tags(Tags.create(null, ImmutableList.copyOf(tags)))
-              .serviceAccounts(options.serviceAccounts())
-              .scheduling(scheduling)
-              .build();
+            name,
+            hardwareUri, // machineType
+            network,
+            disks)
+            .description(group)
+            .tags(Tags.create(null, ImmutableList.copyOf(tags)))
+            .serviceAccounts(options.serviceAccounts())
+            .scheduling(scheduling)
+            .build();
 
       // Add metadata from template and for ssh key and image id
       newInstance.metadata().putAll(options.getUserMetadata());
--- a/providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
+++ b/providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
@@ -25,23 +25,15 @@ import static org.jclouds.googlecloud.internal.ListPages.concat;
 import static org.jclouds.googlecomputeengine.compute.strategy.CreateNodesWithGroupEncodedIntoNameThenAddToSet.simplifyPorts;
 import static org.jclouds.googlecomputeengine.config.GoogleComputeEngineProperties.IMAGE_PROJECTS;
 
-import javax.inject.Inject;
-import javax.inject.Named;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
 
-import com.google.common.base.Predicate;
-import com.google.common.base.Splitter;
-import com.google.common.base.Strings;
-import com.google.common.cache.LoadingCache;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
-import com.google.common.util.concurrent.Atomics;
-import com.google.common.util.concurrent.UncheckedTimeoutException;
+import javax.inject.Inject;
+import javax.inject.Named;
+
 import org.jclouds.compute.ComputeServiceAdapter;
 import org.jclouds.compute.domain.Hardware;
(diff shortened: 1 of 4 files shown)


providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
```diff
@@ -141,15 +146,16 @@
...
-      NewInstance newInstance = new NewInstance.Builder( name,
-            template.getHardware().getUri(), // machineType
-            network,
-            disks)
-            .description(group)
-            .tags(Tags.create(null, ImmutableList.copyOf(tags)))
-            .serviceAccounts(options.serviceAccounts())
-            .scheduling(scheduling)
-            .build();
+      NewInstance newInstance = new NewInstance.Builder(
```
bostko: [minor] can you preserve identation

bostko:
Wouldn't be nice if you preserve the feature for setting hardwareUri somehow?

andreaturli:
I think it is a nice-to-have but hard to achieve. I think short hardwareId will be enough, don't you think @bostko ?

nacx:
Hmmm thinking a bit more about this, how does this improve the provider?
I mean, now the code is robust to changes in GCE, as the API already provides which profiles are available in each zone (assuming the adapter now does the right thing and returns each hardware profile for each zone).

With your change, it looks like the `listHardwareProfiles` method would be inconsistent? What would it return? It should return just a single copy of each hardware profile, but, can we guarantee they will be the same in each region (and no more and no less)?

After thinking about it, I'm not sure if it is a good idea to differ from how they API behaves just to provide a sugar when specifying the hardware ids.

aledsage:
@nacx @andreaturli My opinion is that we should offer a consistent cloud abstraction with jclouds. I don't think any of the other providers require the region to be included in the hardwareId, so ideally we shouldn't require it for GCE.

(Is that statement true? Do any other providers include the region/zone in the hardwareId?)

But I'm not yet sure how we'd achieve that while maintaining consistency between listHardwareProfiles and template options!

Some general thoughts/observations below.

---
As for some machine-types (i.e. hardwareIds) not being available in some zones, that is also true for aws-ec2 (new hardware profiles are usually introduced in just a sub-set of regions first). We are happy to use the simple name in jclouds, but then AWS is a bad example because their API doesn't let you list the machine-types!

Looking at how Google themselves present this in the CLI, they use the simple name rather than asking people to repeat the zone:
    https://cloud.google.com/sdk/gcloud/reference/compute/instances/set-machine-type

There is an interesting comparison for AMIs however: in aws-ec2 the AMI id includes the region. However, I think that is not a fair comparison because the id will always be different in different regions (or if it is the same it is purely coincidence).

---
As to how `listHardwareProfiles` should behave, I agree that's very tricky. We could return `n1-standard-1` multiple times, with each instance having a different `hardware.getLocation()`. But the hardware.getId javadoc is clear that if we did so then the id should include the zone:

    /**
     * 
     * A means to uniquely address this resource within a provider. For example, if the namespace of
     * a node or image is region based, the id will likely include both the region and the
     * provider-supplied id encoded to avoid collisions.
     * 
     * @return unique id within your account on the provider
     */

And if `hardware.getId()` includes the zone, then for consistency that is what a user should supply in the template.

An alternative would be to *not* tell the user which locations support which machine-types when they call `listHardwareProfiles`. That feels consistent with the (ugly) aws-ec2 implementation, but sub-optimal.

andreaturli:
@aledsage good points. Thanks for your comments.

Just a minor addition to what you said: if GCE CLI returns just simple names, I think we can return them and remove `uri` from `Hardware` so that `listHardwareProfiles` will be then consistent.

Makese sense @nacx, @aledsage, @bostko ?

aledsage:
@andreaturli not quite. The GCE CLI's equivalent of listHardwareProfiles gives a nice table: one column is the zone, another column is the machine-type. So it does list each machine-type multiple times.

It is when you use that machine-type name in subsequent GCE CLI calls that it accepts the simple name.

Unfortunately that does not fit cleanly with the current jclouds abstraction (where listHardwareProfiles and template options are expected to agree, and where we have javadoc description of what hardware profiles should contain.

nacx:
> I don't think any of the other providers require the region to be included in the hardwareId, so ideally we shouldn't require it for GCE.
(Is that statement true? Do any other providers include the region/zone in the hardwareId?)

I think this statement is tricky, because it assumes we (and users) know in advance how hardware IDs are encoded internally in the provider. But we should assume we don't even know they are URIs! jclouds interfaces are prepared to work fine with hardware IDs whatever format they have, even if they are URIs, strings that encode additional information, etc. We don't care. Whatever the provider API says "this is how you should uniquely identify this hardware profile", we use that. As soon as we start making our own assumptions and parsing, we are opening the door to future issues when things change in the provider (hopefully never).

My concerns are that the current code and behavior is consistent, does not make any assumption on how the cloud provider works, and meets the contract of the jclouds interfaces, but changing this to shorten the hardware profile id, compromises the above, just to provide a sugar when specifying the hardware profile id.

I understand that having URIs with the location encoded in them can look ugly, or not convenient for some use cases (and it is not jclouds the one who made the choice of how hardware profile ids are identified in GCE), but IMO that logic could be better/easily implemented in the applications that use jclouds?

I'm just putting in a balance the trade-offs and the gains of the patch.

andreaturli:
Thanks @nacx and @aledsage for your interesting inputs to the conversation.

Let me step back for a moment, I think with this patch I'm trying to make the hardwareId independent of the `project` and `zone`, as GCE URI for MachineType looks like:
`https://www.googleapis.com/compute/v1/projects/jclouds-gce/zones/us-central1-a/machineTypes/f1-micro`.

My idea is to simplify the user experience: supplying a short hardwareId, one doesn't need to specify/remember the project + zone, but I understand @nacx  when he said this approach may potentially "open the door to future issues, if the provider changes".

andreaturli:
I'm gonna close it as it doesn't need to be really useful as-is. We'll come back to it later if needed

PR #953, opened 2016-05-17
Fix NPE when node is null

SerCeMan:
Method formatStatus throws NPE if the node is null. 
Now we throw readable message instead.


nacx:
Pushed to master as [0e41b45a](http://git-wip-us.apache.org/repos/asf/jclouds/commit/0e41b45a). Thanks @SerCeMan!

PR #954, opened 2016-05-17
Support for "Placement.Tenancy" and "Placement.HostId" AWS EC2 parameters

storgashov:
Fixes [JCLOUDS-1114](https://issues.apache.org/jira/browse/JCLOUDS-1114).

This adds support for "Placement.Tenancy" and "Placement.HostId" parameters for the "RunInstances" action in AWS EC2 API.

See also AWS EC2 API docs:
- [RunInstances](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html) action
- [Placement](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Placement.html) data type


nacx:
Thanks for the patch @storgashov! Code looks good, but a couple of tests should also be added:

* Add the new options to the [AWSInstanceApiTest#testRunInstancesOptions](https://github.com/jclouds/jclouds/blob/master/providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/AWSInstanceApiTest.java#L119-L143), to verify the requests are properly generated.
* Include the new options in the [AWSEC2ComputeServiceApiMockTest#launchVPCSpotInstanceSubnetId](https://github.com/jclouds/jclouds/blob/master/providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/compute/AWSEC2ComputeServiceApiMockTest.java#L49-L87), or just copy the test and create a new one, to verify that options are properly populated from the compute service options to the generated http request.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/options/AWSRunInstancesOptions.java
```diff
@@ -61,6 +61,22 @@ public AWSRunInstancesOptions inPlacementGroup(String placementGroup) {
    }
 
    /**
+    * Specifies the tenancy of the instance within which to launch the instance(s).
+    */
+   public AWSRunInstancesOptions withTenancy(String tenancy) {
```
nacx: Looking at the docs, it looks like the only allowed values are `default | dedicated | host`. I think it is worth creating an enum so users don't have the chance to pass in the wrong values (the same enum should be used here and in the aws template options.

storgashov:
@nacx, thanks for your review! I've updated the patch according to your comments, please take a look.
Also please confirm, if the ``Tenancy`` enum name and package I've selected is ok.

nacx:
Squashed, amended commit message to include the JIRA issue, and pushed to master as [ccd1ef2b](http://git-wip-us.apache.org/repos/asf/jclouds/commit/ccd1ef2b).Thanks @storgashov!

PR #957, opened 2016-05-19
update CloudstackTemplateOptions pattern

andreaturli:


apis/cloudstack/src/main/java/org/jclouds/cloudstack/compute/options/CloudStackTemplateOptions.java
```diff
@@ -48,265 +55,208 @@
...
-      return this.getNetworks();
+   public CloudStackTemplateOptions userData(String userData) {
+      this.userData = checkNotNull(userData, "userdata").getBytes(Charsets.UTF_8);
+      return this;
    }
 
-   public CloudStackTemplateOptions setupStaticNat(boolean setupStaticNat) {
-      this.setupStaticNat = setupStaticNat;
+   public CloudStackTemplateOptions userData(URL userDataUrl) throws IOException {
+      this.userData = Resources.toString(checkNotNull(userDataUrl, "userDataUrl"), Charsets.UTF_8).getBytes(Charsets.UTF_8);
```
nacx: Why providing this method that accepts a URL? I don't see it necessary and only introduces issues: can we guarantee the classloader being used to load the resource the right one to load the file? And if jclouds code is deployed in a webapp? Where do we expect to load files from?

IMO there is no need to provide this "sugar" method. Let users read files/resources themselves and just pass the array of bytes.

PR #959, opened 2016-05-20
Handle nested braces in UriTemplates.expand

andrewgaul:
This allows replacement of JSON-like payloads without using hacks like
percent-encoding braces.


nacx:
lgtm!

PR #961, opened 2016-05-21
[JCLOUDS-1115] Include filesystem API dependency in jclouds-allblobstore when using Java 8 or higher

Gerrit-Janssen:


andrewgaul:
Thank you for your contribution @Gerrit-Janssen!

PR #968, opened 2016-06-16
Properly set the permissions in the ~/.ssh folder

nacx:
When writing the `~/.ssh/authorized_keys` file in a script that is being run as sudo, the file is created with the root owner, instead of the user defined by the node credentials. File ownership should be enforced to make sure the right owner is alwaays set.


nacx:
To make it right it should also consider the node credentials, as this `options.LoginUser` will only be present when someone overrides it in the options.
This requires more investigation, so I'll close it and reopen with a proper fix.

PR #969, opened 2016-06-16
JCLOUDS-1116: Delegate endpoint should be preferred to the caller's one

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1116

When computing the endpoint for delegate apis, if both the caller and the callee define the endpoint, then the one in the callee should take precedence.


core/src/test/java/org/jclouds/rest/internal/RestAnnotationProcessorTest.java
```diff
@@ -411,6 +418,46 @@ public HttpResponse invoke(HttpCommand command) {
...
+         @Override
+         public HttpResponse invoke(HttpCommand command) {
+            assertEquals(command.getCurrentRequest().getRequestLine(), "GET http://localhost:1111/client/1 HTTP/1.1");
+            return HttpResponse.builder().build();
+         }
+      });
+
+      try {
+         child.getInstance(Callee.class);
+         fail("Callee shouldn't be bound yet");
```
andrewgaul: Might want to use assertj `failBecauseExceptionWasNotThrown`.

andrewgaul:
:+1:

core/src/test/java/org/jclouds/rest/internal/RestAnnotationProcessorTest.java
nacx: Done!

core/src/main/java/org/jclouds/rest/internal/RestAnnotationProcessor.java
```diff
@@ -198,14 +198,19 @@ public GeneratedHttpRequest apply(Invocation invocation) {
...
             logger.trace("using endpoint %s from invocation.getArgs() for %s", endpoint, invocation);
-      } else if (caller != null) {
-         endpoint = getEndpointFor(caller);
-         if (endpoint.isPresent())
-            logger.trace("using endpoint %s from caller %s for %s", endpoint, caller, invocation);
-         else
-            endpoint = findEndpoint(invocation);
       } else {
-         endpoint = findEndpoint(invocation);
+         endpoint = getEndpointFor(invocation);
```
zack-shoylev: Document the new logic a little bit for future reference?

nacx:
Squashed, rebased, added comment and merged to master as [040df11d](http://git-wip-us.apache.org/repos/asf/jclouds/commit/040df11d). Thanks for the review!

PR #970, opened 2016-06-17
JCLOUDS-1129:Add more interfaces of the same network

Urban123:
Change the implementation to List instead of Set, so all the networks are preserved and configured in openstack, not only distinct ones.
Link to issue: https://issues.apache.org/jira/browse/JCLOUDS-1129#


nacx:
Pushed to master as [71b3a2b6](http://git-wip-us.apache.org/repos/asf/jclouds/commit/71b3a2b6). Thanks @Urban123!

PR #971, opened 2016-06-17
Properly set the permissions in the ~/.ssh folder

nacx:
When writing the ~/.ssh/authorized_keys file in a script that is being run as sudo, the file is created with the root owner, instead of the user defined by the node credentials. File ownership should be enforced to make sure the right owner is always set.

This will require changes to jclouds-labs to align existing classes to the new interface.

@ritazh Could you test this with the azurecompute-arm provider?

I've just opened this now to get early feedback. I still have to run several live tests before merging.


nacx:
I've validated the change with google-compute-engine and it does not break the live tests, so I assume it is good to go as soon as I get the confirmation for azurecompute-arm.

ritazh:
@nacx LGTM! Confirming this works as expected with azurecompute-arm provider.

nacx:
Pushed to master as [179ed3b2](http://git-wip-us.apache.org/repos/asf/jclouds/commit/179ed3b2). Thanks @ritazh!

PR #974, opened 2016-06-19
Fix Issue with flavor creation by making links nullable

arvindn05:
Fix Issue with flavor creation by making links nullable
see https://issues.apache.org/jira/browse/JCLOUDS-1126


zack-shoylev:
Looks good to me!

arvindn05:
thanks Zack!

zack-shoylev:
merged and backported

nacx:
Can we mark the JRIA issue as resolved?

arvindn05:
i think so for now. But the issue i created was to track a general issue with with new validations introduced as part of v2.1 of openstack api's.

This might be only the first instance of these new validations causing issues...i will reopen if i see issues in other apis.

Thanks for the quick review!

nacx:
Will stay tuned then. Thanks for the contribution!

PR #975, opened 2016-06-20
Changes the upload behavior to parallel, completing a TODO

zack-shoylev:


apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -582,19 +589,58 @@ public long countBlobs(String containerName, ListContainerOptions options) {
...
+
       for (Payload payload : slicer.slice(blob.getPayload(), partSize)) {
-         MultipartPart part = uploadMultipartPart(mpu, partNumber, payload);
-         parts.add(part);
          ++partNumber;
+         BlobUploader b =
+               new BlobUploader(mpu, partNumber, payload, parts);
+         executor.execute(b);
+      }
+      executor.shutdown();
```
andrewgaul: Instead of creating a new `ExecutorService` per `putBlob` request, we should take an optional one as an argument.  This would allow callers to control the degree of parallelism with multiple `putBlob` calls.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -582,19 +589,58 @@ public long countBlobs(String containerName, ListContainerOptions options) {
...
+      try {
+         executor.awaitTermination(Integer.MAX_VALUE, TimeUnit.DAYS);
+      } catch (InterruptedException e) {
+         e.printStackTrace();
+      }
+
+      return completeMultipartUpload(mpu, new ArrayList( parts ));
+   }
+
+   private class BlobUploader implements Runnable {
```
andrewgaul: Use `AutoValue` and make `static`?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -582,19 +589,58 @@ public long countBlobs(String containerName, ListContainerOptions options) {
...
+
+      BlobUploader(MultipartUpload mpu, int partNumber, Payload payload, ConcurrentLinkedQueue<MultipartPart> list) {
+         this.mpu = mpu;
+         this.partNumber = partNumber;
+         this.payload = payload;
+         this.list = list;
+      }
+
+      @Override
+      public void run() {
```
andrewgaul: If you make this a `Callable` can you use the return value instead of the explicit `list`?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -582,19 +589,58 @@ public long countBlobs(String containerName, ListContainerOptions options) {
...
             "must provide content-length to use multi-part upload");
       MultipartUploadSlicingAlgorithm algorithm = new MultipartUploadSlicingAlgorithm(
             getMinimumMultipartPartSize(), getMaximumMultipartPartSize(), getMaximumNumberOfParts());
       long partSize = algorithm.calculateChunkSize(contentLength);
       MultipartUpload mpu = initiateMultipartUpload(container, blob.getMetadata(), partSize, overrides);
       int partNumber = 1;
+
+      int numThreads = 16;
+      if (System.getProperty("jclouds.mpu.parallel.degree") != null) {
+         numThreads = Integer.getInteger(System.getProperty("jclouds.mpu.parallel.degree"));
```
andrewgaul: Also shouldn't we inject this property instead of using the system properties?

zack-shoylev: Also I am not very sure if we should be using this property at all and I am not sure about how it is documented (the website states it has a default value of 4, actually, not 16, but it is also not used elsewhere in our code).

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
zack-shoylev: OK, this makes sense, but might make it pretty complex for users. What about making the thread-pool for large blob uploads static instead? Thoughts? Do both? Not very sure what is the best design here. So many options :)

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
zack-shoylev: Making it static means uploadMultipartPart in BlobStore will have be made static, otherwise it's referencing a dynamic context from a static one. But if I do that it would mean changes in other providers and so forth. There might be other problems with that, too.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
andrewgaul: For a while we have talked about geting away from jclouds managing threads; we want the contexts to be more lightweight and default to caller context/single-thread execution.  For performance-critical code like threaded multipart upload and clear container we can either take an explicit `ExecutorService` or push it into the `ContextBuilder`.  The latter is more discoverable but feels more clunky, but in the end the user should manage the lifetime.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
zack-shoylev: As for AutoValue, I am not sure if it's worth it, and I am not sure it will work well considering payload and multipartupload are mutable... I think...

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -579,22 +590,79 @@ public long countBlobs(String containerName, ListContainerOptions options) {
...
    }
 
-   // copied from BaseBlobStore
+   // TODO: Move statics up to BaseBlobStore to reuse?
+   @com.google.inject.Inject(optional = true)
+   @Named("jclouds.mpu.parallel.degree")
+   @VisibleForTesting
+   private int numThreads = 16;
+
+   private ExecutorService jcloudsBigBlobUploadTreadPool = MoreExecutors.getExitingExecutorService(
```
andrewgaul: We should not add random `ExecutorService` to our code.  Either we reuse the deprecated `userExecutor` and `ioExecutor` or make a caller provide this via a parameter.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -579,22 +590,79 @@ public long countBlobs(String containerName, ListContainerOptions options) {
...
+      private Payload payload;
+
+      BlobUploader(MultipartUpload mpu, int partNumber, Payload payload) {
+         this.mpu = mpu;
+         this.partNumber = partNumber;
+         this.payload = payload;
+      }
+
+      @Override
+      public final MultipartPart call() {
```
andrewgaul: Remove `final` here, add to class and fields.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -579,22 +590,79 @@ public long countBlobs(String containerName, ListContainerOptions options) {
...
-         parts.add(part);
-         ++partNumber;
+         BlobUploader b =
+               new BlobUploader(mpu, partNumber++, payload);
+         parts.add(executor.submit(b));
+      }
+
+      return completeMultipartUpload(mpu, Lists.newArrayList(Iterables.transform(parts, new Function<Future<MultipartPart>, MultipartPart>() {
+         @Override
+         public MultipartPart apply(Future<MultipartPart> input) {
```
andrewgaul: Call `Futures.allAsList().get()` instead.  Possibly with `Futures.getUnchecked` to handle exceptions.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -579,22 +590,79 @@ public long countBlobs(String containerName, ListContainerOptions options) {
...
+            try {
+               return input.get();
+            } catch (InterruptedException e) {
+               e.printStackTrace();
+               throw new RuntimeException(e);
+            } catch (ExecutionException e) {
+               e.printStackTrace();
```
andrewgaul: Don't `printStackTrace`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -579,22 +590,79 @@ public long countBlobs(String containerName, ListContainerOptions options) {
...
    protected String putMultipartBlob(String container, Blob blob, PutOptions overrides) {
-      List<MultipartPart> parts = Lists.newArrayList();
+      if (overrides.getCustomExecutor() != null) {
+         return putMultipartBlob(container, blob, overrides, overrides.getCustomExecutor());
+      } else {
+         return putMultipartBlob(container, blob, overrides, jcloudsBigBlobUploadTreadPool);
+      }
+   }
+
+   // copied from BaseBlobStore
```
andrewgaul: We need similar code in `BaseBlobStore` for other providers now that we change `PutOptions`.  This will also resolve [JCLOUDS-227](https://issues.apache.org/jira/browse/JCLOUDS-227).

blobstore/src/main/java/org/jclouds/blobstore/options/PutOptions.java
```diff
@@ -36,6 +38,7 @@
 
    private BlobAccess blobAccess = BlobAccess.PRIVATE;
    private boolean multipart = false;
+   private ExecutorService customExecutor = null;
```
andrewgaul: Instead use `MoreExecutors.sameThreadExecutor` so all code can rely on this.

blobstore/src/main/java/org/jclouds/blobstore/options/PutOptions.java
```diff
@@ -44,6 +47,16 @@ public PutOptions(boolean multipart) {
...
    }
 
+   public PutOptions(boolean multipart, ExecutorService customExecutor) {
+      this.multipart = multipart;
+      this.customExecutor = customExecutor;
+   }
+
+   public PutOptions(ExecutorService customExecutor) {
+      this.multipart = true;
+      this.customExecutor = customExecutor;
```
andrewgaul: Call `Preconditions.checkNotNull`.

blobstore/src/main/java/org/jclouds/blobstore/options/PutOptions.java
```diff
@@ -44,6 +47,16 @@ public PutOptions(boolean multipart) {
       this.multipart = multipart;
    }
 
+   public PutOptions(boolean multipart, ExecutorService customExecutor) {
```
andrewgaul: Do we need this variant?

blobstore/src/main/java/org/jclouds/blobstore/options/PutOptions.java
```diff
@@ -87,6 +110,15 @@ public BlobAccess getBlobAccess() {
       return blobAccess;
    }
 
+   public ExecutorService getCustomExecutor() {
+      return customExecutor;
+   }
+
+   public PutOptions setCustomExecutor(ExecutorService customExecutor) {
+      this.customExecutor = customExecutor;
```
andrewgaul: Call `Preconditions.checkNotNull`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
zack-shoylev: Hmm yes. Do you think it would be possible to move all the thread-pool code to BaseBlobStore, or will it cause issues with some providers?

blobstore/src/main/java/org/jclouds/blobstore/options/PutOptions.java
zack-shoylev: Yeah that seems redundant.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
zack-shoylev: Using an existing executor means we can't use the property variable for number of threads, though, right? Since the variable is already documented, I think there is an expectation that such a separate executor should exist.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
andrewgaul: We should include it in `BaseBlobStore`, although `RegionScopedSwiftBlobStore` does not subclass it.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
andrewgaul: Incredibly, none of the multipart configuration variables actually *do* anything, except for this one.  We should just remove all of them.

zack-shoylev:
@andrewgaul Have another look. Also, I'd like to merge this before I start adding similar code in the abstraction layer (separate PR).

andrewgaul:
👍

zack-shoylev:
merged

PR #976, opened 2016-06-21
JCLOUDS-1125: local blobstore list MPUs

andrewgaul:


andrewgaul:
@zack-shoylev Please review.  This commit changes the way that multipart uploads are named and makes the approach more similar to S3 which has a separate namespace for MPU.

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
```diff
@@ -100,6 +100,7 @@
 
 @Singleton
 public final class LocalBlobStore implements BlobStore {
+   private static final String MULTIPART_PREFIX = ".mpus-";
```
zack-shoylev: Should this be configurable?

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
```diff
@@ -783,24 +784,29 @@ public boolean createContainerInLocation(Location location, String container, Cr
 
    @Override
    public MultipartUpload initiateMultipartUpload(String container, BlobMetadata blobMetadata, PutOptions options) {
-      return MultipartUpload.create(container, blobMetadata.getName(), UUID.randomUUID().toString(),
+      String uploadId = UUID.randomUUID().toString();
+      // create a stub blob
+      Blob blob = blobBuilder(MULTIPART_PREFIX + uploadId + "-" + blobMetadata.getName() + "-0").payload(ByteSource.empty()).build();
```
zack-shoylev: The -0 might be confusing. -stub maybe?

zack-shoylev:
Looks good. I will run some more tests on my windows machine tomorrow.

zack-shoylev:
Ran tests, looks good.

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
andrewgaul: Done.

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
andrewgaul: I don't think so; it might be nice to configure MPUs to go into another directory or similar but I don't see much need to change the prefix.

PR #977, opened 2016-06-21
Allow subclasses to override the scripts

nacx:
These changes are needed in https://github.com/jclouds/jclouds-labs/pull/291

/cc @ritazh

### DIFF
--- a/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java
+++ b/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java
@@ -290,7 +290,7 @@ public abstract class BaseComputeServiceLiveTest extends BaseComputeServiceConte
    }
 
    @Test(enabled = false)
-   public void weCanCancelTasks(NodeMetadata node) throws InterruptedException, ExecutionException {
+   protected void weCanCancelTasks(NodeMetadata node) throws InterruptedException, ExecutionException {
       ListenableFuture<ExecResponse> future = client.submitScriptOnNode(node.getId(), "sleep 300",
             nameTask("sleeper").runAsRoot(false));
       ExecResponse response = null;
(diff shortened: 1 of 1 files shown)


compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java
```diff
@@ -386,11 +386,11 @@ public void testCreateTwoNodesWithOneSpecifiedName() throws Exception {
...
    }
 
-   private Template refreshTemplate() {
+   protected Template refreshTemplate() {
       return template = addRunScriptToTemplate(buildTemplate(client.templateBuilder()));
    }
 
-   protected static Template addRunScriptToTemplate(Template template) {
+   protected Template addRunScriptToTemplate(Template template) {
       template.getOptions().runScript(Statements.newStatementList(AdminAccess.standard(), InstallJDK.fromOpenJDK()));
```
demobox: If subclasses need to override this, does it mean `AdminAccess.standard()` doesn't work there? That might be an issue to address too?

compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java
```diff
@@ -386,11 +386,11 @@ public void testCreateTwoNodesWithOneSpecifiedName() throws Exception {
       this.nodes = newTreeSet(concat(this.nodes, nodes));
    }
 
-   private Template refreshTemplate() {
+   protected Template refreshTemplate() {
```
demobox: @nacx: Similarly to the comment below, if the "standard" code this test is executing doesn't work, does it mean users will also need to execute "non-standard" commands to interact successfully with the image?

compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java
nacx: This has been spotted during the Azure ARM development. The scripts work fine, *but* when a node starts for the first time, Azure installs some extensions and that uses APT in Debian based distros. APT does not allow concurrent executions, and the jclouds script fails. This just happens because when jclouds finds the port 22 open, it immediately runs the script, but Azure is still installing the extensions.

To mitigate that, we allow providers to override how the script is build, so they can inject logic bypass these kind of things, while preserving the purpose of the test. In the current Azure PR they're using this to add a "sleep 50", but that will be replaced by [this commit I have on a branch in my fork](https://github.com/nacx/jclouds/commit/2860ec9aff5e4f009946a78fe8f64446456bbdd5), that would allow to effectively wait until the APT lock is gone with something like `ScriptBuilder.waitForLockFile("/var/lib/dpkg/lock", 100)`. I'm waiting to open the PR until I have confirmation that the commit properly fixes the APT lock issue.

So, long story short, the jclouds scripts are OK and should work in most cases. However, we've identified this particular case in which the jclouds execution has to wait a bit more depending on the status of the node, so I nade these methods protected to let the scripts be customized.

PR #979, opened 2016-07-11
Isolate the task cancellation in a test

nacx:
Extract the task cancellation to a specific test so it does not prevent other tests from running if it fails.


nacx:
@ritazh Once this is merged we should be able to remove the override of the task cancellation method from the ARM tests. It's just FYI. Let's leave them unchanged now and I'll take care of updating them once merged.

PR #982, opened 2016-07-13
JCLOUDS-1074: Guava 20 compatibility

andrewgaul:
JCLOUDS-1074: Guava 20 compatibility

* dynamically call TypeToken.isSupertypeOf with Guava 19 and later and
  TypeToken.isAssignableFrom with Guava 18 and earlier
* consume or ignore values from methods with CheckReturnValue
* replace usage of removed Iterators.emptyIterator

### DIFF
--- a/core/src/main/java/org/jclouds/internal/BaseView.java
+++ b/core/src/main/java/org/jclouds/internal/BaseView.java
@@ -25,6 +25,7 @@ import org.jclouds.Context;
 import org.jclouds.View;
 import org.jclouds.location.Provider;
 import org.jclouds.rest.ApiContext;
+import org.jclouds.util.TypeTokenUtils;
 
 import com.google.common.base.Objects;
 import com.google.common.base.Objects.ToStringHelper;
@@ -45,7 +46,7 @@ public abstract class BaseView extends ForwardingObject implements View {
    @SuppressWarnings("unchecked")
    @Override
    public <C extends Context> C unwrap(TypeToken<C> type) {
-      checkArgument(checkNotNull(type, "type").isAssignableFrom(backendType), "backend type: %s not assignable to %s", backendType, type);
+      checkArgument(TypeTokenUtils.isSupertypeOf(checkNotNull(type, "type"), backendType), "%s is not a supertype of backend type %s", type, backendType);
       return (C) backend;
    }
    
--- a/core/src/main/java/org/jclouds/util/TypeTokenUtils.java
+++ b/core/src/main/java/org/jclouds/util/TypeTokenUtils.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.util;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+
+import com.google.common.annotations.Beta;
+import com.google.common.base.Throwables;
+import com.google.common.reflect.TypeToken;
+
+@Beta
+public final class TypeTokenUtils {
+   private static final Method IS_SUPERTYPE_OF_TYPE;
+   private static final Method IS_SUPERTYPE_OF_TYPETOKEN;
+   static {
+      Method isSuperTypeOfType;
+      Method isSuperTypeOfTypeToken;
+      try {
+         // Guava 19 and later method
+         isSuperTypeOfType = TypeToken.class.getDeclaredMethod("isSupertypeOf", Type.class);
+         isSuperTypeOfTypeToken = TypeToken.class.getDeclaredMethod("isSupertypeOf", TypeToken.class);
+      } catch (NoSuchMethodException nsme) {
+         try {
+            // Guava 18 and earlier method
+            isSuperTypeOfType = TypeToken.class.getDeclaredMethod("isAssignableFrom", Type.class);
+            isSuperTypeOfTypeToken = TypeToken.class.getDeclaredMethod("isAssignableFrom", TypeToken.class);
+         } catch (NoSuchMethodException nsme2) {
+            throw Throwables.propagate(nsme2);
+         }
+      }
+      IS_SUPERTYPE_OF_TYPE = isSuperTypeOfType;
+      IS_SUPERTYPE_OF_TYPETOKEN = isSuperTypeOfTypeToken;
+   }
+
+   private TypeTokenUtils() {
+      throw new AssertionError("intentionally not implemented");
+   }
+
+   public static <C> boolean isSupertypeOf(TypeToken<C> token, Type type) {
+      try {
+         return (Boolean) IS_SUPERTYPE_OF_TYPE.invoke(token, type);
+      } catch (IllegalAccessException iae) {
+         throw Throwables.propagate(iae);
+      } catch (InvocationTargetException ite) {
+         throw Throwables.propagate(ite);
+      }
+   }
+
+   public static <C, D> boolean isSupertypeOf(TypeToken<C> token, TypeToken<D> token2) {
+      try {
+         return (Boolean) IS_SUPERTYPE_OF_TYPETOKEN.invoke(token, token2);
+      } catch (IllegalAccessException iae) {
+         throw Throwables.propagate(iae);
+      } catch (InvocationTargetException ite) {
+         throw Throwables.propagate(ite);
+      }
+   }
+}
--- a/core/src/test/java/org/jclouds/internal/BaseViewTest.java
+++ b/core/src/test/java/org/jclouds/internal/BaseViewTest.java
@@ -104,7 +104,7 @@ public class BaseViewTest {
          wine.unwrap(typeToken(PeanutButter.class));
          fail();
       } catch (IllegalArgumentException e) {
-         assertEquals(e.getMessage(), "backend type: org.jclouds.internal.BaseViewTest$Water not assignable to org.jclouds.internal.BaseViewTest$PeanutButter");
+         assertEquals(e.getMessage(), "org.jclouds.internal.BaseViewTest$PeanutButter is not a supertype of backend type org.jclouds.internal.BaseViewTest$Water");
       }
    }
 
--- a/providers/softlayer/src/test/java/org/jclouds/softlayer/features/VirtualGuestApiLiveTest.java
+++ b/providers/softlayer/src/test/java/org/jclouds/softlayer/features/VirtualGuestApiLiveTest.java
@@ -17,6 +17,7 @@
 package org.jclouds.softlayer.features;
 
 import static com.google.common.base.Preconditions.checkState;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.jclouds.util.Predicates2.retry;
 import static org.testng.Assert.assertEquals;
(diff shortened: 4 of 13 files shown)


andrewgaul:
This pull request may prove contentious since it upgrades Guava to the latest release to work around a removed method.  Do we want to go through the pain of reflection to make this more compatible?

@blop opened in response to [JCLOUDS-1074](https://issues.apache.org/jira/browse/JCLOUDS-1074)

core/pom.xml
```diff
@@ -50,6 +50,14 @@
...
     </dependency>
     <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_annotations</artifactId>
+    </dependency>
+    <dependency>
```
demobox: Have these been broken out of the Guava dep?

core/src/main/java/org/jclouds/internal/BaseView.java
```diff
@@ -45,7 +45,7 @@ protected BaseView(@Provider Context backend, @Provider TypeToken<? extends Cont
    @SuppressWarnings("unchecked")
    @Override
    public <C extends Context> C unwrap(TypeToken<C> type) {
-      checkArgument(checkNotNull(type, "type").isAssignableFrom(backendType), "backend type: %s not assignable to %s", backendType, type);
+      checkArgument(checkNotNull(type, "type").isSupertypeOf(backendType), "backend type: %s not assignable to %s", backendType, type);
```
demobox: [minor] Perhaps change the message to `type: %s is not a supertype of %s...` to match the code? Probably more comprehensible for users, too ;-)

providers/softlayer/src/test/java/org/jclouds/softlayer/features/VirtualGuestApiLiveTest.java
```diff
@@ -133,7 +134,7 @@ public void testSetTagsOnVirtualGuest() throws Exception {
       Set<TagReference> tagReferences = found.getTagReferences();
       assertNotNull(tagReferences);
       for (String tag : tags) {
-         Iterables.contains(tagReferences, tag);
+         assertThat(Iterables.contains(tagReferences, tag)).isTrue();
```
demobox: Nice catch - thanks!

core/src/main/java/org/jclouds/internal/BaseView.java
andrewgaul: Done.

providers/softlayer/src/test/java/org/jclouds/softlayer/features/VirtualGuestApiLiveTest.java
andrewgaul: This is due to Guava 20 introducing annotating new methods which error-prone catches.

core/pom.xml
andrewgaul: Sorry the jsr305 was a spurious change.  Without the error-prone annotations dependency, I see errors of the form:

```
error: cannot access CanIgnoreReturnValue
  class file for com.google.errorprone.annotations.CanIgnoreReturnValue not found
```

providers/softlayer/src/test/java/org/jclouds/softlayer/features/VirtualGuestApiLiveTest.java
andrewgaul: Inspecting this further, this test always failed due to `tagReferences` having a different type than `tag`.  I have corrected it.

nacx:
Can we setup a Guavva 20 Jenkins build to validate this branch?

demobox:
> This pull request may prove contentious since it upgrades Guava to the latest release to work around a removed method. Do we want to go through the pain of reflection to make this more compatible?

Thanks for the warning, @andrewgaul! Hm...Guava 20 isn't even released yet, and we are still compatible with 19.0, which is the latest official version. On the other hand, given that this is `master` and we are getting close to 2.0, it seems like waiting until _after_ 2.0 for this is not going to make things any easier.

* do we have any idea what the tentative release date is for Guava 20?
* is there any way we could easily find out which Guava versions people are using? Specifically, how many are looking at using 20?

Given that we are on 16 on the next branch down, forcing 19 on to users from `master` seems like a pretty big gap. Unless we have good data that lots of users are planning to use 20 (only 1 vote on the issue so far), I'm hesitant about this one as I think it will force more users into upgrading than it will help users looking to use 20.

Can we put a call out to ask users to upvote JCLOUDS-1074 if the are affected by this?

nacx:
This is something we want to bring to the dev mailing list.

The only problem with this is dropping compatibility with Guava < 19, which will probably have a big impact in users. However, if we don't do such a change in 2.0... then when?

Like @demobox, I'm a bit hesitant until we know a bit more about the Guava versions people are using now, although not that worried by the version gap between 1.9.x and 2.0. We already have the [guava/guice compatibility build](https://jclouds.ci.cloudbees.com/job/jclouds-guava-guice-compat/) and I'd say it should work for 1.9.x too. Despite Guava 16 being the default version we use, Guava 19 should be already supported, so users should be able to upgrade Guava without upgrading jclouds, when planning a future upgrade to 2.0.

We should run a Guava 19 build against 1.9.x and, if it succeeds, I'd be happy to have this merged as soon as we document a smooth upgrade plan for users willing to upgrade to 2.0.

nacx:
We also need to make sure we don't break jclouds-karaf and the CLI.

demobox:
> This is something we want to bring to the dev mailing list...we should run a Guava 19 build against 1.9.x 

Sounds like a good plan to me.

> Despite Guava 16 being the default version we use, Guava 19 should be already supported

It's not so much that we already support 19 that I'm concerned about, more that we currently allow users of our stable version to use any Guava version from 16 upwards. If we move to 19, that will force all stable users to jump up to 3 major versions (two years of development) of a commonly-used library, and we don't know how many other libraries they might be using which are **not** compatible with Guava 19? There's a decent amount of changes from 16 to 19:

* http://google.github.io/guava/releases/17.0/api/diffs/
* http://google.github.io/guava/releases/18.0/api/diffs/
* http://google.github.io/guava/releases/19.0/api/diffs/

Having said that, I agree that 2.0 would be the most obvious time to consider this. Still, I'd want to avoid a situation where we end up losing more 2.0 users because of a forced Guava jump than we gain from being Guava 20 compatible.

To be continued on the dev list, I suspect... ;-)

andrewgaul:
Based on your feedback I will go through the small pain of reflection to address Guava 20 compatibility.

demobox:
> Based on your feedback I will go through the small pain of reflection to address Guava 20 compatibility.

Thanks, Andrew G!

nacx:
Ping?

andrewgaul:
I am working on this and should complete it over the weekend.

andrewgaul:
Updated without Guava 19 dependency.

core/src/test/java/org/jclouds/internal/BaseViewTest.java
```diff
@@ -104,7 +104,7 @@ public void testPeanutButterDidntTurnIntoWine() {
          wine.unwrap(typeToken(PeanutButter.class));
          fail();
       } catch (IllegalArgumentException e) {
-         assertEquals(e.getMessage(), "backend type: org.jclouds.internal.BaseViewTest$Water not assignable to org.jclouds.internal.BaseViewTest$PeanutButter");
+         assertEquals(e.getMessage(), "backend type: org.jclouds.internal.BaseViewTest$Water is not a supertype of org.jclouds.internal.BaseViewTest$PeanutButter");
```
nacx: Does this message depend on the Guava version being used? In that case we need to change the assertion, otherwise our builds that check different versions of Guava will fail.

nacx:
Thanks @andrewgaul! Just one comment in a test.

core/src/main/java/org/jclouds/internal/BaseView.java
```diff
@@ -45,7 +46,7 @@ protected BaseView(@Provider Context backend, @Provider TypeToken<? extends Cont
    @SuppressWarnings("unchecked")
    @Override
    public <C extends Context> C unwrap(TypeToken<C> type) {
-      checkArgument(checkNotNull(type, "type").isAssignableFrom(backendType), "backend type: %s not assignable to %s", backendType, type);
+      checkArgument(TypeTokenUtils.isSupertypeOf(checkNotNull(type, "type"), backendType), "backend type: %s is not a supertype of %s", backendType, type);
```
demobox: Should the error message be the other way around? I.e. `"type %s is not a supertype of backend type %s", type, backendType'?

core/src/main/java/org/jclouds/util/TypeTokenUtils.java
```diff
@@ -0,0 +1,74 @@
...
+         } catch (NoSuchMethodException nsme2) {
+            throw Throwables.propagate(nsme2);
+         }
+      }
+      IS_SUPERTYPE_OF_TYPE = isSuperTypeOfType;
+      IS_SUPERTYPE_OF_TYPETOKEN = isSuperTypeOfTypeToken;
+   }
+
+   private TypeTokenUtils() {
+      throw new AssertionError("intentionally not implemented");
```
demobox: [minor] Or [`UnsupportedOperationException`](http://docs.oracle.com/javase/6/docs/api/java/lang/UnsupportedOperationException.html)? But if `AssertionError` is what we use in similar situations elsewhere, fine with that

core/src/main/java/org/jclouds/util/TypeTokenUtils.java
```diff
@@ -0,0 +1,74 @@
...
+   }
+
+   public static <C> boolean isSupertypeOf(TypeToken<C> token, Type type) {
```
demobox: Do we know if this is much more efficient that simply doing:
```
  return isSupertypeOf(token, TypeToken.of(type));
```
? See https://google.github.io/guava/releases/19.0/api/docs/com/google/common/reflect/TypeToken.html#of(java.lang.reflect.Type).

That would allow to get down the class to use only _one_ method?

core/src/test/java/org/jclouds/internal/BaseViewTest.java
demobox: I think this is **our** exception, thrown in `Context.unwrap` and modified in this PR above?

@andrewgaul Does that sound right?

providers/softlayer/src/test/java/org/jclouds/softlayer/features/VirtualGuestApiLiveTest.java
```diff
@@ -132,9 +133,13 @@ public void testSetTagsOnVirtualGuest() throws Exception {
...
       assertNotNull(tagReferences);
-      for (String tag : tags) {
-         Iterables.contains(tagReferences, tag);
+
+      ImmutableSet.Builder<String> actualTagsBuilder = ImmutableSet.builder();
+      for (TagReference ref : tagReferences) {
+         actualTagsBuilder.add(ref.getTag().getName());
       }
+      Set<String> actualTags = actualTagsBuilder.build();
+      assertThat(actualTags).containsAll(tags);
```
demobox: Just curious: related to this PR, or just cleanup?

core/src/test/java/org/jclouds/internal/BaseViewTest.java
andrewgaul: Yes `BaseView` generates this, albeit with the typo as remarked on below.

core/src/main/java/org/jclouds/util/TypeTokenUtils.java
andrewgaul: We use the `AssertionError` pattern throughout.

providers/softlayer/src/test/java/org/jclouds/softlayer/features/VirtualGuestApiLiveTest.java
andrewgaul: Guava 20 adds a `CheckReturnValue` annotation to several methods which we must consume when compiling with the newer library.

core/src/main/java/org/jclouds/internal/BaseView.java
andrewgaul: Done.

core/src/main/java/org/jclouds/util/TypeTokenUtils.java
andrewgaul: Your suggestion is shorter and the result is equivalent but I prefer to call through to the reflected method to preserve the intent of the workaround.

PR #978, opened 2016-07-07
Parallel upload for BaseBlobStore

zack-shoylev:
Based on https://github.com/jclouds/jclouds/pull/975

PR #983, opened 2016-07-15
Allows users to download large files efficiently and directly to disk.

zack-shoylev:
Since this has come up in the past via emails, why not provide a way to efficiently get a local copy of the blob's contents? Users can already do it themselves, but it is somewhat complex, and comes up often. Here is a suggestion with swift, but should be doable with baseblobstore as well (if contentlength and ranges are supported).


zack-shoylev:
Planning to add streaming and more tests to this PR

zack-shoylev:
@andrewgaul , this PR is the continuation of the performance improvements started in https://github.com/jclouds/jclouds/pull/978 and https://github.com/jclouds/jclouds/pull/975 but for downloads.

zack-shoylev:
@andrewgaul Do you have any notes on this? I would like to merge this before 2.0 so I can also work on getting similar functionality into base blobstore for the other providers.

zack-shoylev:
merged

Will work on getting something very similar for the rest of the providers, might refactor it more then.

PR #984, opened 2016-07-18
JCLOUDS-482: ArbitraryCpuRam base case

ivanlomba:
- ArbitraryCpuRamTemplateBuilderImpl and GoogleComputeEngineArbitraryCpuRamTemplateBuilderImpl subclasses.

- Parse utility.


compute/src/main/java/org/jclouds/compute/config/BaseComputeServiceContextModule.java
```diff
@@ -91,6 +92,7 @@ protected void configure() {
       }).to(CreateSshClientOncePortIsListeningOnNode.class);
       bind(new TypeLiteral<Function<TemplateOptions, Statement>>() {
       }).to(TemplateOptionsToStatement.class);
+      bind(TemplateBuilderImpl.class).to(ArbitraryCpuRamTemplateBuilderImpl.class);
```
nacx: This cannot be done here. This binding must be applied in the module of each provider that supports an arbitrary cpu and ram config.

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
```diff
@@ -0,0 +1,82 @@
...
+            .id(generateAutomaticId(cores, ram))
+            .ram(ram)
+            .processor(new Processor(cores, 1.0))
+            .build();
+   }
+
+   protected Hardware findHardwareWithId(Set<? extends Hardware> hardwaresToSearch) {
+      try {
+            return super.findHardwareWithId(hardwaresToSearch);
+      } catch (Exception NoSuchElementException) {
```
nacx: Variable names should start in lower case.

nacx: Also, just `catch (NoSuchElementException ex)`

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
```diff
@@ -0,0 +1,82 @@
...
+               throw throwNoSuchElementExceptionAfterLoggingHardwareIds("Invalid hardwareId", hardwaresToSearch);
+            }
+        }
+    }
+
+   protected Hardware resolveHardware(Set<? extends Hardware> hardwarel, final Iterable<? extends Image> images) {
+      try {
+         return super.resolveHardware(hardwarel, images);
+      }
+      catch (Exception NoSuchElementException) {
```
nacx: Same here.

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
```diff
@@ -0,0 +1,82 @@
...
+   protected Hardware findHardwareWithId(Set<? extends Hardware> hardwaresToSearch) {
+      try {
+            return super.findHardwareWithId(hardwaresToSearch);
+      } catch (Exception NoSuchElementException) {
+            if (isAutomaticId(hardwareId)) {
+               AutomaticHardwareIdSpec spec = parseId(hardwareId);
+               return automaticHardwareForCpuAndRam(spec.getCores(), spec.getRam());
+            }
+            else{
+               throw throwNoSuchElementExceptionAfterLoggingHardwareIds("Invalid hardwareId", hardwaresToSearch);
```
nacx: Consider just propagating the current exception here

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+    }
+
+   protected Hardware resolveHardware(Set<? extends Hardware> hardwarel, final Iterable<? extends Image> images) {
+      try {
+         return super.resolveHardware(hardwarel, images);
+      }
+      catch (Exception NoSuchElementException) {
+         if (super.minCores != 0 && super.minRam != 0) {
```
nacx: As commented in the commit, should allow users to just pass the cpu or the ram? Should we have some defaults for that case?

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
```diff
@@ -0,0 +1,82 @@
...
+            return automaticHardwareForCpuAndRam(minCores, minRam);
+            }
+         else throw throwNoSuchElementExceptionAfterLoggingHardwareIds("Invalid hardware", hardwarel);
```
nacx: Same here. Consider just propagating the current exception.

compute/src/main/java/org/jclouds/compute/domain/internal/TemplateBuilderImpl.java
```diff
@@ -131,6 +129,7 @@
    @VisibleForTesting
    protected Boolean forceCacheReload;
 
+   @VisibleForTesting
```
nacx: It is not just visible for testing; it is visible to allow inheritance, so remove this annotation.

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
```diff
@@ -0,0 +1,59 @@
...
+         throw new IllegalArgumentException(String.format("Omitted keys on hardwareId: %s", hardwareId));
+      }
+      spec.ram = Integer.parseInt(specValues.get("ram"));
+      spec.cores = Double.parseDouble(specValues.get("cores"));
+      return spec;
+   }
+
+   public static String generateAutomaticId(double cores, int ram) {
+      return String.format("automatic:cores=%s;ram=%s", cores, ram);
+   }
```
nacx: Consider overriding the `toString()` method instead of having this one.

compute/src/test/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.jclouds.domain.LocationScope;
+import org.jclouds.rest.AuthorizationException;
+import org.testng.annotations.Test;
+
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.easymock.EasyMock.*;
+
+@Test(groups = "unit", singleThreaded = true, testName = "TemplateBuilderImplTest")
```
nacx: Fix the test name.

compute/src/test/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.jclouds.domain.Location;
+import org.jclouds.domain.LocationBuilder;
+import org.jclouds.domain.LocationScope;
+import org.jclouds.rest.AuthorizationException;
+import org.testng.annotations.Test;
+
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.easymock.EasyMock.*;
```
nacx: Remove wildcard imports

compute/src/test/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+      replay(defaultTemplate, optionsProvider, templateBuilderProvider, getImageStrategy);
+      TemplateBuilderImpl templateBuilder = new ArbitraryCpuRamTemplateBuilderImpl(locations, new ImageCacheSupplier(images, 60,
+            Atomics.<AuthorizationException>newReference(), Providers.of(getImageStrategy)), hardwares,
+            Suppliers.ofInstance(region), optionsProvider, templateBuilderProvider);
+      templateBuilder.minRam(1024);
+      templateBuilder.minCores(2);
+      Hardware hardware = templateBuilder.build().getHardware();
+      assertThat(hardware.getRam()).isEqualTo(1024);
+      assertThat(hardware.getProcessors()).extracting("cores").containsExactly(2.0);
+   }
```
nacx: Add a test to verify what happens when you only set the minRam or the minCores, as it is a path that your class is explicitly handling.
Also add a couple more tests where there is actually a hardware profile present:

* One test where the existing hardware profile does not match, and an automatic one is used.
* One test where the existing hardware profile matches the given cpu and ram, to verify that the existing hardware profile is used.

compute/src/test/java/org/jclouds/compute/util/AutomaticHardwareIdSpecTest.java
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.compute.util;
+
+import org.testng.annotations.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Test(groups = "unit", testName = "IdParserTest")
```
nacx: Fix test name.

providers/google-compute-engine/src/test/java/org/jclouds/googlecomputeengine/compute/domain/internal/GoogleComputeEngineArbitraryCpuRamTemplateBuilderImplTest.java
```diff
@@ -0,0 +1,153 @@
...
+      expect(getImageStrategy.getImage(anyObject(String.class))).andReturn(null);
+      replay(defaultTemplate, optionsProvider, templateBuilderProvider, getImageStrategy);
+      TemplateBuilderImpl templateBuilder = new GoogleComputeEngineArbitraryCpuRamTemplateBuilderImpl(locations, new ImageCacheSupplier(images, 60,
+            Atomics.<AuthorizationException>newReference(), Providers.of(getImageStrategy)), hardwares,
+            Suppliers.ofInstance(region), optionsProvider, templateBuilderProvider);
+      templateBuilder.minRam(1024);
+      templateBuilder.minCores(2);
+      Hardware hardware = templateBuilder.build().getHardware();
+      assertThat(hardware.getRam()).isEqualTo(1024);
+      assertThat(hardware.getProcessors()).extracting("cores").containsExactly(2.0);
```
nacx: Verify the hardware ID too, to make sure an automatic one has been picked.

compute/src/test/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImplTest.java
nacx: Also verify the hardware ID too, to make sure an automatic one has been picked.

nacx:
Thanks @ivanlomba! Good start. Apart from the inline comments, have a look at the formatting, and re-indent the code to use a 3 space indent. Also have a look at the checkstyle violations.

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
ivanlomba: Considering that if only minRam or minCores are provided and they match with a hardware profile a VM is created according to the hardware profile, I think that we should have some defaults. I will ask on the mail list about the default values.

compute/src/main/java/org/jclouds/compute/domain/internal/TemplateBuilderImpl.java
ivanlomba: Yes, it was a misprint, I don't know how this annotation came here :(

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
nacx: Another option would be to throw a NoSuchElementException with a proper messsge. Something like "No hardware profile was found matching the given criteria. If you want to use exact values, please set the minCores and minRam values".

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
```diff
@@ -0,0 +1,84 @@
...
+   protected Hardware resolveHardware(Set<? extends Hardware> hardwarel, final Iterable<? extends Image> images) {
+      try {
+         return super.resolveHardware(hardwarel, images);
+      }
+      catch (NoSuchElementException ex) {
+         if (super.minCores != 0 && super.minRam != 0) {
+            return automaticHardwareForCpuAndRam(minCores, minRam);
+         }
+         else throw new NoSuchElementException("No hardware profile matching the given criteria was found. " +
+               "If you want to use exact values, please set the minCores and minRam values");
```
nacx: Wrap the original exception as the cause of this one, to keep the complete stacktrace.

compute/src/main/java/org/jclouds/compute/config/BaseComputeServiceContextModule.java
```diff
@@ -65,21 +66,19 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.jclouds.Constants.PROPERTY_SESSION_INTERVAL;
+import static org.jclouds.compute.config.ComputeServiceProperties.IMAGE_ID;
+import static org.jclouds.compute.config.ComputeServiceProperties.TEMPLATE;
+import static org.jclouds.compute.domain.OsFamily.UBUNTU;
 
 public abstract class BaseComputeServiceContextModule extends AbstractModule {
```
nacx: Remove changes in this class; it is only reorganizing the imports.

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
```diff
@@ -0,0 +1,70 @@
...
+   public static AutomaticHardwareIdSpec parseId(String hardwareId) {
+      AutomaticHardwareIdSpec spec = new AutomaticHardwareIdSpec();
+      String hardwareSpec = hardwareId.substring(10);
+      Map<String, String> specValues = Splitter.on(';')
+            .trimResults()
+            .omitEmptyStrings()
+            .withKeyValueSeparator('=')
+            .split(hardwareSpec);
+      if (!specValues.containsKey("ram") || !specValues.containsKey("cores")) {
+         throw new IllegalArgumentException(String.format("Omitted keys on hardwareId: %s", hardwareId));
```
nacx: Give feedback about what is missing.

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
```diff
@@ -0,0 +1,70 @@
...
+      }
+      spec.ram = Integer.parseInt(specValues.get("ram"));
+      spec.cores = Double.parseDouble(specValues.get("cores"));
+      return spec;
+   }
+
+   public static AutomaticHardwareIdSpec AutomaticHardwareIdSpecBuilder(Double cores, int ram) {
```
nacx: Why one argument is an object t and the other is a primitive type?

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
```diff
@@ -0,0 +1,70 @@
...
+      AutomaticHardwareIdSpec spec = new AutomaticHardwareIdSpec();
+      if (cores == 0 || ram == 0) {
+         throw new IllegalArgumentException(String.format("Omitted or wrong parms"));
```
nacx: Be explicit about what is exactly missing or wrong.

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
nacx: Also, method names should start with a lower case letter.

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
```diff
@@ -0,0 +1,70 @@
...
+package org.jclouds.compute.util;
+
+import com.google.common.base.Splitter;
+
+import java.util.Map;
+
+public class AutomaticHardwareIdSpec {
+
+   private Double cores;
+   private Integer ram;
```
nacx: If both fields are mandatory and an instance of this class can't be built without both being set, change the types to primitives to enforce non-null values.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/config/GoogleComputeEngineServiceContextModule.java
```diff
@@ -91,6 +91,8 @@ protected void configure() {
       bind(new TypeLiteral<ComputeServiceAdapter<Instance, MachineType, Image, Location>>() {
       }).to(GoogleComputeEngineServiceAdapter.class);
 
+      bind(ArbitraryCpuRamTemplateBuilderImpl.class).to(GoogleComputeEngineArbitraryCpuRamTemplateBuilderImpl.class);
```
nacx: This should be:
```java
bind(TemplateBuilder.class).to(GoogleComputeEngineArbitraryCpuRamTemplateBuilderImpl.class);
```

nacx:
The build is failing due to checkstyle violations

>[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.15:check (default) on project jclouds-compute: You have 2 Checkstyle violations. -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.15:check (default) on project jclouds-compute: You have 2 Checkstyle violations.

You can run a local build, and have a look at the `compute/target/checkstyle-result.xml` for the details.

compute/src/test/java/org/jclouds/compute/internal/BaseTemplateBuilderLiveTest.java
```diff
@@ -44,6 +49,9 @@
 @Test(groups = "integration,live")
 public abstract class BaseTemplateBuilderLiveTest extends BaseComputeServiceContextLiveTest {
 
+   public static final Object[][] NO_INVOCATIONS = new Object[0][0];
+   public static final Object[][] SINGLE_NO_ARG_INVOCATION = { new Object[0] };
```
nacx: This is a recurring pattern. Move these constants to the [TestUtils](https://github.com/jclouds/jclouds/blob/master/core/src/test/java/org/jclouds/utils/TestUtils.java) class, and also remove them from the filesystem provider and other classes, so all data providers using this pattern use the same constants.

nacx:
This looks pretty good @ivanlomba! Thanks!

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
```diff
@@ -0,0 +1,84 @@
...
+      } catch (NoSuchElementException ex) {
+            if (isAutomaticId(hardwareId)) {
+               AutomaticHardwareIdSpec spec = parseId(hardwareId);
+               return automaticHardwareForCpuAndRam(spec.getCores(), spec.getRam());
+            }
+            else {
+               throw ex;
+            }
+        }
+    }
```
demobox: [minor] Indenting in the above section is off?

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
```diff
@@ -0,0 +1,84 @@
...
+            .id(automaticHardwareIdSpecBuilder(cores, ram).toString())
+            .ram(ram)
+            .processor(new Processor(cores, 1.0))
+            .build();
+   }
+
+   protected Hardware findHardwareWithId(Set<? extends Hardware> hardwaresToSearch) {
+      try {
+            return super.findHardwareWithId(hardwaresToSearch);
+      } catch (NoSuchElementException ex) {
```
demobox: Is catching an exception and falling back the best way of doing this? Would the following alternative perhaps also work?

```
if (isAutomaticId(...)) {
   // do stuff for automated IDs
} else {
  // fall back to original behaviour
}
```

The immediate challenge I can see with this is that users _might_ have existing hardware IDs that match this scheme. Is there a way we can avoid this by not "overloading" the ID in some way?

compute/src/main/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImpl.java
```diff
@@ -0,0 +1,84 @@
...
+               throw ex;
+            }
+        }
+    }
+
+   protected Hardware resolveHardware(Set<? extends Hardware> hardwarel, final Iterable<? extends Image> images) {
+      try {
+         return super.resolveHardware(hardwarel, images);
+      }
+      catch (NoSuchElementException ex) {
```
demobox: Same comment as above - can we avoid "try then catch" as a strategy here?

compute/src/main/java/org/jclouds/compute/domain/internal/TemplateBuilderImpl.java
```diff
@@ -727,7 +724,7 @@ private Image loadImageWithId(Iterable<? extends Image> images) {
       return image.get();
    }
 
-   private Hardware findHardwareWithId(Set<? extends Hardware> hardwaresToSearch) {
+   protected Hardware findHardwareWithId(Set<? extends Hardware> hardwaresToSearch) {
       Hardware hardware;
       // TODO: switch to GetHardwareStrategy in version 1.5
```
demobox: Likely unrelated to this commit, but...is this comment still relevant?

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
```diff
@@ -0,0 +1,72 @@
...
+public class AutomaticHardwareIdSpec {
+
+   private double cores;
+   private int ram;
+
+   public static boolean isAutomaticId(String id) {
+      return id.startsWith("automatic:");
+   }
+
+   public static AutomaticHardwareIdSpec parseId(String hardwareId) {
```
demobox: I'm guessing there's no way to avoid encoding this a string and using an (automatically deserialized) object instead?

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
```diff
@@ -0,0 +1,72 @@
...
+            .trimResults()
+            .omitEmptyStrings()
+            .withKeyValueSeparator('=')
+            .split(hardwareSpec);
+      if (!specValues.containsKey("ram") || !specValues.containsKey("cores")) {
+         throw new IllegalArgumentException(String.format("Omitted keys on hardwareId: %s. Please set number " +
+               "of cores and ram amount.", hardwareId));
+      }
+      spec.ram = Integer.parseInt(specValues.get("ram"));
+      spec.cores = Double.parseDouble(specValues.get("cores"));
```
demobox: Do we need some exception handling here?

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
```diff
@@ -0,0 +1,72 @@
...
+      return spec;
+   }
+
+   public static AutomaticHardwareIdSpec automaticHardwareIdSpecBuilder(double cores, int ram) {
+      AutomaticHardwareIdSpec spec = new AutomaticHardwareIdSpec();
+      if (cores == 0 || ram == 0) {
```
demobox: Are negative values allowed?

compute/src/test/java/org/jclouds/compute/domain/internal/ArbitraryCpuRamTemplateBuilderImplTest.java
```diff
@@ -0,0 +1,275 @@
...
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.easymock.EasyMock.anyObject;
+import static org.easymock.EasyMock.createMock;
+import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.replay;
+
+@Test(groups = "unit", singleThreaded = true, testName = "ArbitraryCpuRamTemplateBuilderImplTest")
+public class ArbitraryCpuRamTemplateBuilderImplTest {
```
demobox: Can we extend another test and inherit the fixture?

demobox:
Thanks for all the work on this, @ivanlomba! My main question is about the need to encode the "arbitrary spec" in a string and do our our parsing, rather than somehow using an object and having jclouds do the (de-)serialization automatically.

If that's already something that's been discussed on the mailing list of elsewhere, apologies ;-)

/cc @ivanlomba

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
ivanlomba: Yes, It's only checking that the user has specified both values (they are primitives).

demobox: > It's only checking that the user has specified both values (they are primitives).

What would we get here if one of the values was *not* specified? Wouldn't we see the same default value here? Also, equality checking of floating-point values is a little tricky, because depending on the system you can get things like `a/2 - a/2 != 0` due to rounding issues.

nacx:
Squashed, rebased to the latest version of master and merged to the GSoC branch. I've also cherry-picked the commit in https://github.com/jclouds/jclouds/pull/993.

compute/src/main/java/org/jclouds/compute/util/AutomaticHardwareIdSpec.java
ivanlomba: Thanks @demobox ! :)

PR #985, opened 2016-08-02
add support for security groups

andreaturli:
if ok, I'd like to backport it to 1.9.x

### DIFF
--- a/providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
+++ b/providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
@@ -22,29 +22,17 @@ import static com.google.common.collect.Iterables.filter;
 import static com.google.common.collect.Lists.newArrayList;
 import static java.lang.String.format;
 import static org.jclouds.googlecloud.internal.ListPages.concat;
-import static org.jclouds.googlecomputeengine.compute.strategy.CreateNodesWithGroupEncodedIntoNameThenAddToSet.simplifyPorts;
 import static org.jclouds.googlecomputeengine.config.GoogleComputeEngineProperties.IMAGE_PROJECTS;
 
-import javax.inject.Inject;
-import javax.inject.Named;
 import java.net.URI;
-import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 
-import com.google.common.base.Function;
-import com.google.common.base.Predicate;
-import com.google.common.base.Splitter;
-import com.google.common.base.Strings;
-import com.google.common.cache.LoadingCache;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
-import com.google.common.util.concurrent.Atomics;
-import com.google.common.util.concurrent.UncheckedTimeoutException;
+import javax.inject.Inject;
+import javax.inject.Named;
+
 import org.jclouds.compute.ComputeServiceAdapter;
 import org.jclouds.compute.domain.Hardware;
 import org.jclouds.compute.domain.NodeMetadata;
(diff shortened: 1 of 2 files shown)


providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
```diff
@@ -142,9 +146,18 @@
...
       List<String> ports = simplifyPorts(options.getInboundPorts());
-      if (ports != null){
+      if (ports != null) {
          tags.add(naming.name(ports));
       }
+      // Add tags coming from firewalls (aka securityGroups)
+      if (!options.getGroups().isEmpty()) {
+         for (String firewallName : options.getGroups()) {
+            Firewall firewall = api.firewalls().get(firewallName);
+            validateFirewall(firewall, network);
```
nacx: This is redundant, as it has already been validated in the `CreateNodesWithGroupEncodedIntoNameThenAddToSet`.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
```diff
@@ -142,9 +146,18 @@
...
+            if (!firewall.targetTags().isEmpty()) {
+               tags.addAll(firewall.targetTags());
```
nacx: Missing closing `}`.

nacx:
Having a closer look, the strategy that validates the firewalls will call the api only once for each firewall, but then the code in the adapter will be executed once for each firewall and node being deployed, which is too many redundant calls.

Given that the adapter [already sets the configured tags](https://github.com/jclouds/jclouds/blob/master/providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java#L140), I'd change the strategy to, instead of just validating, add the tags to the templateOptions.tags, so the adapter always gets the list of tags to apply properly populated. This way we eliminate all code from the adapter that is coupled to the firewall naming generation and keep everything in one place (and we save all those redundant calls). WDYT?

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/strategy/CreateNodesWithGroupEncodedIntoNameThenAddToSet.java
```diff
@@ -186,6 +201,15 @@ private void getOrCreateFirewalls(GoogleComputeEngineTemplateOptions templateOpt
          operationDone.apply(operation);
          checkState(operation.get().httpErrorStatusCode() == null, "Could not insert firewall, operation failed %s",
                operation);
+
+         // Add tags for firewalls
+         tags.add(name);
+      }
+   }
+
+   public static void validateFirewall(Firewall firewall, URI networkUri) {
```
nacx: Does it make more sense for the signature to be `validateFirewall(Firewall firewall, Network network)`?

nacx:
Just the minor cosmetic comment about the signature of the validate firewall method. But regardless of that, LGTM!
Good to emrge @andreaturli! Thanks!

andreaturli:
thanks @nacx, can't see the cosmetic comment btw

- is it fine to backport this to 1.9.x ?

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/strategy/CreateNodesWithGroupEncodedIntoNameThenAddToSet.java
```diff
@@ -186,6 +202,16 @@ private void getOrCreateFirewalls(GoogleComputeEngineTemplateOptions templateOpt
...
          checkState(operation.get().httpErrorStatusCode() == null, "Could not insert firewall, operation failed %s",
                operation);
+
+         // Add tags for firewalls
+         tags.add(name);
+      }
+      templateOptions.tags(tags);
+   }
+
+   public static void validateFirewall(Firewall firewall, URI networkUri) {
```
nacx: Does it make more sense for the signature to be validateFirewall(Firewall firewall, Network network)?

nacx:
It does not introduce breaking changes, so +1 to backport it.
I've re-posted the comment.

andreaturli:
ah! it is not easy to backport it as in 1.9.x it was on jclouds-labs, so I need to open another PR referencing this review.

andreaturli:
rebuild please

andreaturli:
the [error](https://jclouds.ci.cloudbees.com/job/jclouds-pull-requests/2368/console) doesn't seem related to this PR

andreaturli:
Merged *only* at [master](https://git1-us-west.apache.org/repos/asf?p=jclouds.git;a=commit;h=deeedd55)

nacx:
Tests fail due to this change (the order of the instance tags has been altered). Could you fix it please?
https://jclouds.ci.cloudbees.com/job/jclouds/3457/

andreaturli:
Sorry I will asap

andreaturli:
https://jclouds.ci.cloudbees.com/job/jclouds/org.apache.jclouds.provider$google-compute-engine/3458/changes

PR #988, opened 2016-08-05
JCLOUDS-1148 JWTBearerTokenFlow.tokenCache never hits

daniel-h-rech:
This PR fixes [https://issues.apache.org/jira/browse/JCLOUDS-1148](https://issues.apache.org/jira/browse/JCLOUDS-1148)

### DIFF
--- a/apis/oauth/src/main/java/org/jclouds/oauth/v2/domain/ClientSecret.java
+++ b/apis/oauth/src/main/java/org/jclouds/oauth/v2/domain/ClientSecret.java
@@ -38,12 +38,9 @@ public abstract class ClientSecret {
     /** The scope(s) to authorize against. **/
     @Nullable public abstract String scope();
 
-    /** When does the token expire. **/
-    public abstract long expire();
-
-    @SerializedNames({ "client_id", "client_secret", "resource", "scope", "expire" })
-    public static ClientSecret create(String clientId, String clientSecret, String resource, String scope, long expire) {
-        return new AutoValue_ClientSecret(clientId, clientSecret, resource, scope, expire);
+    @SerializedNames({ "client_id", "client_secret", "resource", "scope" })
+    public static ClientSecret create(String clientId, String clientSecret, String resource, String scope) {
+        return new AutoValue_ClientSecret(clientId, clientSecret, resource, scope);
     }
 
     ClientSecret() {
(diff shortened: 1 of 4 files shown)


nacx:
Thanks @daniel-h-rech! Change LGTM. The [ClientCredentialsJWTBearerTokenFlow](https://github.com/jclouds/jclouds/blob/master/apis/oauth/src/main/java/org/jclouds/oauth/v2/filters/ClientCredentialsJWTBearerTokenFlow.java#L96-L111) and the [ClientCredentialsSecretFlow](https://github.com/jclouds/jclouds/blob/master/apis/oauth/src/main/java/org/jclouds/oauth/v2/filters/ClientCredentialsSecretFlow.java#L92-L99) use claim caches in a similar buggy way. Mind applying the same pattern to those flows too?

daniel-h-rech:
It seems in `ClientCredentialsSecretFlow` we could just remove `expire()` from `ClientSecret`, since it never really gets used.

What do you think?

nacx:
The `ClientSecret` class is also used in the `ClientCredentialsJWTBearerTokenFlow`, and the expires seems to be used there. Instead of removing it, would it make sense to set it to null in the client secret flow?
@jmspring WDYT?

daniel-h-rech:
I see no reference to `ClientSecret` in `ClientCredentialsJWTBearerTokenFlow`, only in `ClientCredentialsSecretFlow`.

`ClientCredentialsSecretFlow` is actually sending the secret key every time, so setting an expiration time seems redundant.

ClientCredentialsJWTBearerTokenFlow actually uses `ClientCredentialsClaims` and `ClientCredentialsAuthArgs`.

jmspring:
Hey, just seeing this.  Let me take a look and comment a bit later today.

-j

nacx:
>ClientCredentialsJWTBearerTokenFlow actually uses ClientCredentialsClaims and ClientCredentialsAuthArgs.

Yes @daniel-h-rech. What I mean is that it also [uses random values to build the object being cached](
https://github.com/jclouds/jclouds/blob/master/apis/oauth/src/main/java/org/jclouds/oauth/v2/filters/ClientCredentialsJWTBearerTokenFlow.java#L100-L102), so it will present the same issue when loading the elements back from the cache. Every get will be a cache miss, so it is worth applying the same pattern you proposed?

daniel-h-rech:
I fixed the remaining flows.

nacx:
I'll do it. Thanks! I'll run a quick the on azure and squash and merge.

nacx:
Live tests are OK in GCE and Azure. I squashed the commits and amended the commit message to reflect that the change is not particular to the JWTBearerTokenFlow.
Merged to master as [913fdeb1](http://git-wip-us.apache.org/repos/asf/jclouds/commit/913fdeb1). Thanks @daniel-h-rech!

PR #989, opened 2016-08-07
Cleanup imports in google-compute-engine project.

nelsonjr:


andreaturli:
Thx @nelsonjr 

could you please avoid to reorder the imports and remove only the unused? It's a bit hard to review this PR otherwise.

nelsonjr:
Can we do in 2 steps then? :-) How about 2 separate commits, one for removing unused and another to reorder?

nelsonjr:
Actually there is only 1 instance of unused import:
org.jclouds.googlecomputeengine.domain.NewTargetInstance.Builder;

All the order changes are merely moving them according to the rest of the project. Would that be okay?

nacx:
Reordering it is very dependent on the IDE defaults, which are different in Eclipse and IDEA (the two most used by our contributors), so eventually we'll see imports moved back again. That's why we ask contributors to keep the order as-is, since we don't have an import order policy.

This said, I personally don't care much about import ordering, so feel free to merge/discard if that facilitates further contributions :)

nelsonjr:
Okay, in that case let's abandon this PR.

PR #990, opened 2016-08-07
Add support for organization internal projects.

nelsonjr:
When using IAM a company can have internal-only projects, which are
referenced in the form <domain>:<project_id>. IAM account is in the
form: account@project_id.company_domain.iam.gserviceaccount.com.


nacx:
Thanks @nelsonjr! Could you add the corresponding unit tests to the [ClientEmailTest](https://github.com/nelsonjr/jclouds/blob/master/common/googlecloud/src/test/java/org/jclouds/googlecloud/config/ClientEmailTest.java) class?

nelsonjr:
Test case added. Please take a look at your convenience.

nacx:
Pushed to master as [52477d5e](http://git-wip-us.apache.org/repos/asf/jclouds/commit/52477d5e). Thanks @nelsonjr!

PR #993, opened 2016-08-09
Always take into account the configured template builder spec in live tests

nacx:
@ivanlomba Can you cherry-pick this commit into your branch, compile it, and try running the live tests for GCE? Run all the tests, not just a single one to make sure all maven configuration for the live profile is properly applied.


ivanlomba:
Thanks : )

compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java
```diff
@@ -524,6 +524,14 @@ protected void checkNodes(Iterable<? extends NodeMetadata> nodes, String group,
       }
    }
 
+   protected TemplateBuilder templateBuilder() {
+      TemplateBuilder templateBuilder =  client.templateBuilder();
```
demobox: [formatting] Double space here

demobox:
@nacx @ivanlomba Just to confirm: this commit is only about the *tests*, correct?

nacx:
Yes. The base class for the compute service live tests was ignoring the "template" test property in some tests, so even when configured in the pom.xml or as system property, a default template was used instead of the configured one.

This lead to some tests failing to connect via SSH if the template builder picked a windows machine, etc.
I still want to run the tests in a couple more providers before merging it.

demobox:
> Yes. The base class for the compute service live tests was ignoring the "template" test property in some tests, so even when configured in the pom.xml or as system property, a default template was used instead of the configured one.

Ah, OK - thanks for explaining!

nacx:
I've run the tests in several providers and the fix works fine.
Amended the formatting thing, rebased and merged to master as [449c6d8d](http://git-wip-us.apache.org/repos/asf/jclouds/commit/449c6d8d). Thanks @ivanlomba and @demobox!

PR #995, opened 2016-08-11
Do not try to connect to ssh if already connected

nacx:
The `refreshAndRunAction` is called by the `future.cancel()` method when cancelling the future associated to a running script (typically obtained by calling the `compute.submitScriptOnNode` method.

I've observed that when the ssh connection is already established, this method fails to connect again and throws an exception. This was also causing random failures in the `BaseComputeServiceLiveTest.testWeCanCancelTasks`.

@devcsrj This will potentially fix [JCLOUDS-1058](https://issues.apache.org/jira/browse/JCLOUDS-1058). Could you build this branch and try running the tests?

### DIFF
--- a/compute/src/main/java/org/jclouds/compute/callables/SudoAwareInitManager.java
+++ b/compute/src/main/java/org/jclouds/compute/callables/SudoAwareInitManager.java
@@ -71,7 +71,9 @@ public class SudoAwareInitManager {
    public ExecResponse refreshAndRunAction(String action) {
       checkState(ssh != null, "please call init() before invoking call");
       try {
-         ssh.connect();
+         if (!ssh.isConnected()) {
+            ssh.connect();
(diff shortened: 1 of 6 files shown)


devcsrj:
Oh, this workaround might do the trick.

P.S.: My trial account in PB got expired though. I'll see what I could do.

nacx:
```bash
Starting test testWeCanCancelTasks(org.jclouds.profitbricks.compute.ProfitBricksComputeServiceLiveTest)
[TestNG] Test testWeCanCancelTasks(org.jclouds.profitbricks.compute.ProfitBricksComputeServiceLiveTest) succeeded: 238582ms
```

Yay! :)

zack-shoylev:
This might fix some issues I have seen as well. Thanks @nacx

PR #999, opened 2016-08-25
Allow overriding getUserDefinedFileAttributeView

freedge:
Today the filesystem blobstore provider use the extended attributes of the
filesystem to store metadata. This is not always desirable as some filesystems
(such as NFSv3) do not support extended metadata.  The current source code does
not allow to change this easily. getUserDefinedFileAttributeView could easily
be overriden for this purpose and allow extensions that customize where the
metadata should be stored.

See also JCLOUDS-658


zack-shoylev:
+1

zack-shoylev:
merged

andrewgaul:
Sorry I missed this earlier, but could you explain this change?  Many systems including Mac OS X do not support extended attributes; how does it fail without this support?

PR #1000, opened 2016-08-28
JCLOUDS-1160: Upgrade to Gson 2.7

andrewgaul:
Changelog:

https://github.com/google/gson/blob/master/CHANGELOG.md

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -80,7 +80,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.3.1</version>
+      <version>2.7</version>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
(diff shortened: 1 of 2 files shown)


demobox:
Thanks, Andrew G! I had only checked the changelog between 2.6.2 and 2.7 - are there any transitive deps that might be affected by the jump from 2.3.1?

/cc @nacx

nacx:
I don't think so. AFAIK we don't have dependencies that rely on gson apart form jclouds-core. I'd say the change is safe.

All mock and expect tests passing give us a lot of confidence. There are several providers (such as Chef) that use complex adapters and all tests are still passing. There are also the google auto tests, that check the sensible corner cases, so as long as all tests pass, I'd say this is good to merge.

demobox:
> I'd say this is good to merge

Good enough for me - +1 from me too. Thanks for the additional details, Ignasi!

andrewgaul:
gson does not have any runtime dependencies:

```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ gson ---
[INFO] com.google.code.gson:gson:jar:2.7.1-SNAPSHOT
[INFO] \- junit:junit:jar:4.12:test
[INFO]    \- org.hamcrest:hamcrest-core:jar:1.3:test
```

demobox:
> gson does not have any runtime dependencies:

Thanks for confirming, Andrew G!

PR #1004, opened 2016-09-01
JCLOUDS-1166/JCLOUDS-1160: Downgrade to Gson 2.5

nacx:
This change reverts https://github.com/jclouds/jclouds/commit/da4614e72ce17999afbe54ae32879f27a59d8c5c and upgrades from Gson 2.3.1 to 2.5, which is the last version that exports the internal packages, used in jclouds-core, in the OSGi bundle.

I've run a jclouds-karaf build with this version and it seems to work.

/cc @demobox @andrewgaul Let's merge this to fix the build and let's figure out how we want to proceed if we want to actually upgrade to 2.7? We are still pending confirmation if Spring Boot 1.4.0 actually requires version 2.7 (a quick look to its dependencies marks Gson as `optional`).


core/pom.xml
```diff
@@ -80,7 +80,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.7</version>
+      <version>2.5</version>
```
zack-shoylev: Maybe add a comment about why we are doing this here? In case it takes a while for us to move to 2.7

andrewgaul:
👍 Stop the bleeding.

demobox:
> I've run a jclouds-karaf build with this version and it seems to work.

Great, thanks!

> All checks have failed

Looks like a transient test failure: https://jclouds.ci.cloudbees.com/job/jclouds-pull-requests/org.apache.jclouds$jclouds-core/2407/testReport/org.jclouds.http.handlers/BackoffLimitedRetryHandlerTest/testExponentialBackoffDelaySmallInterval1/

Let's try this again...

nacx:
Merged to master as [4c927636](http://git-wip-us.apache.org/repos/asf/jclouds/commit/4c927636). And included the reasoning behind this change in the commit message as per @zack-shoylev's suggestion.

PR #1005, opened 2016-09-05
Fix autovalue class loading in karaf

duncangrant:
When Class.forName is called for a class in a different bundle it will
fail as the default karaf class loader won't load classes from other
bundles.

I have fixed this by using the classloader of the original
(non-autovalue) type and assuming it will be in the same bundle as the
autovalue type (I think this is a reasonable assumtion).

So far the only place where I've actually seen this being an issue is
when using the jclouds-labs-google provider within karaf. It fails
when serialising the Firewall.Rule class within a FirewallOptions
object.


duncangrant:
https://issues.apache.org/jira/browse/JCLOUDS-1171

nacx:
Thanks @duncangrant! A quick search shows that this is the only place where `Class.forName` is used, and other uses of `getClassLoader()` seem legit. I'll let the CI build complete and merge the PR.

duncangrant:
@nacx Thanks.  Should I create a PR to port this to 1.9.x?

nacx:
>Should I create a PR to port this to 1.9.x?

Not needed. I'll take care of merging and cherry-picking  :)

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/7aad5997) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/c79b2355). Thanks @duncangrant!

PR #1006, opened 2016-09-07
JCLOUDS-1122: Support subnetworks definitions in Google Compute.

nelsonjr:


providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
```diff
@@ -136,11 +136,17 @@
       URI network = URI.create(networks.next());
       assert !networks.hasNext() : "Error: Options should specify only one network";
 
+      Iterator<String> subnetworks = options.getSubnetworks().iterator();
+
+      URI subnetwork = subnetworks.hasNext() ? URI.create(subnetworks.next()) : null;
+      assert !subnetworks.hasNext() : "Error: Options should specify only one subnetwork";
```
nacx: Since this is an option specific to google (unlike "networks" that is in the jclouds portable interface), make it just an URI instead of a collection, to avoid unnecessary configuration issues.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/loaders/SubnetworkLoader.java
```diff
@@ -0,0 +1,57 @@
...
+
+   @Inject
+   SubnetworkLoader(Resources resources) {
+      this.resources = resources;
+   }
+
+   @Override
+   public Subnetwork load(URI key) throws ExecutionException {
+      try {
+         return resources.subnetwork(key);
```
nacx: Is this such a freqüent/expensive operation that requires a cache?

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/loaders/SubnetworkLoader.java
```diff
@@ -0,0 +1,57 @@
...
+      } catch (Exception e) {
+         throw new ExecutionException(message(key, e), e);
+      }
+   }
+
+   public static String message(URI key, Exception e) {
```
nacx: Make this private

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/options/GoogleComputeEngineTemplateOptions.java
```diff
@@ -32,6 +35,7 @@
    private List<ServiceAccount> serviceAccounts;
    private String bootDiskType;
    private boolean preemptible = false;
+   private Set<String> subnetworks;
```
nacx: Change to just `URI subnetwork`, as mentioned above?

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/domain/Network.java
```diff
@@ -38,24 +46,39 @@
...
    @Nullable public abstract String gatewayIPv4();
 
-   @SerializedNames({ "id", "creationTimestamp", "selfLink", "name", "description", "IPv4Range", "gatewayIPv4" })
+   @Nullable public abstract List<URI> subnetworks();
+
+   public static Network create(String id, Date creationTimestamp, URI selfLink, String name, String description, String rangeIPv4,
+                                String gatewayIPv4) {
+      return new AutoValue_Network(id, creationTimestamp, selfLink, name, NetworkType.LegacyNetwork, description,
+              rangeIPv4, gatewayIPv4, null);
+   }
```
nacx: To avoid errors in serialisation, it is better to provide just one factory method. Let's remove this one and keep just the new one.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/domain/NewInstance.java
```diff
@@ -39,16 +39,27 @@
...
+
       @SerializedNames({ "network", "accessConfigs" })
       static NetworkInterface create(URI network, List<AccessConfig> accessConfigs) {
-         return new AutoValue_NewInstance_NetworkInterface(network, accessConfigs);
+         return new AutoValue_NewInstance_NetworkInterface(network, null, accessConfigs);
+      }
+
+      @SerializedNames({ "network", "subnetwork", "accessConfigs" })
+      static NetworkInterface create(URI network, URI subnetwork, List<AccessConfig> accessConfigs) {
+         return new AutoValue_NewInstance_NetworkInterface(network, subnetwork, accessConfigs);
```
nacx: Same here. Keep just one static factory method.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/features/NetworkApi.java
```diff
@@ -126,7 +124,7 @@ Operation createInIPv4Range(@PayloadParam("name") String networkName,
    @Transform(NetworkPages.class)
    Iterator<ListPage<Network>> list(ListOptions options);
 
-   static final class NetworkPages extends BaseToIteratorOfListPage<Network, NetworkPages> {
+   final class NetworkPages extends BaseToIteratorOfListPage<Network, NetworkPages> {
```
nacx: What is the purpose of removing the static modifier?

providers/google-compute-engine/src/test/java/org/jclouds/googlecomputeengine/features/SubnetworkApiMockTest.java
```diff
@@ -0,0 +1,95 @@
...
+      assertEquals(subnetworkApi().list().next(), new ParseSubnetworkListTest().expected(url("/projects")));
+      assertSent(server, "GET", "/projects/party/regions/someregion/subnetworks");
+   }
+
+   public void list_empty() throws Exception {
+      server.enqueue(jsonResponse("/list_empty.json"));
+
+      assertFalse(subnetworkApi().list().hasNext());
+      assertSent(server, "GET", "/projects/party/regions/someregion/subnetworks");
+   }
```
nacx: Add mock tests for the `listPage` methods.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/loaders/SubnetworkLoader.java
```diff
@@ -0,0 +1,57 @@
...
+      return String.format("could not find image for disk %s: %s", key.toString(), e.getMessage());
+   }
+}
```
nacx: Looking in more detail, it looks like this cache is not used. Better remove it.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/functions/Resources.java
```diff
@@ -82,6 +83,10 @@
    @Path("/stop")
    Operation stopInstance(@EndpointParam URI selfLink);
 
+   @Named("Subnetworks:get")
+   @GET
+   @Fallback(NullOnNotFoundOr404.class) @Nullable Subnetwork subnetwork(@EndpointParam URI selfLink);
```
nacx: This method is not used. Remove it?

nacx:
Thanks @nelsonjr! This looks great. Just a couple comments apart from the inline ones:

* Add mock tests for the methods added to the aggregated list api.
* Add live tests for the subnetworks api and the new methods in the aggregated list api. These ones are specially important, since they warn us when things change in the provider.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/GoogleComputeEngineServiceAdapter.java
nelsonjr: I made a collection like networks because like networks it can be specified more than once.

I added the blocker to match networks behavior though, but I'll create a separate PR to remove both and bring multiple interfaces (direct or subnetworked) to the module.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/loaders/SubnetworkLoader.java
nelsonjr: Like networks it requires an API call to gather its details and may drive customers over quota/cost. I'm modeling subnetworks just like networks, because in practical sense they are.

(I almost made subnetworks deriving from network :-) but the different settings deterred me from doing that, and extracting an interface at this point would generate a breaking change that it is not worth)

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/options/GoogleComputeEngineTemplateOptions.java
nelsonjr: [explained in previous comment] TL;DR: subnetworks = networks, multiple.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/domain/Network.java
nelsonjr: I thought this would cause problems if the customer was creating the network in their code using this constructor. That's why I moved the automatic to the more generic but I kept the original one in place.

nelsonjr:
@nacx: How do I configure my local machine to run the live tests? I found some documentation that alludes to adding to the m2 settings, but I do not know what that means :-) That's why I did not write the live tests (yet). I was going to do another PR with that, but if you can help me with setting that up, so I can test it, I shall do it right away.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/domain/Network.java
nacx: This will be merged and released to 2.0, so we can expect some breaking changes. Trivial ones such as this one shouldn't worry us.
An option to make this more update-friendly for future releases would be to provide a builder with AutoValue.Builder. In any case, I think we can just break this in 2.0 and keep jsut one factory method (plus the builder if you want). Same applies for other similar comments on this.

nacx:
> How do I configure my local machine to run the live tests

The better options is to download your credentials as a JSON file. Once you have them, you can run all live tests as follows:

    mvn clean install -Plive -Dtest.google-cloud.json-key=/path/to/your/credentials.json

Or if you want to run just one single test class:

    mvn integration-test -Plive -Dtest.google-cloud.json-key=/path/to/your/credentials.json -Dtest=TheClassTestToRun

nacx:
They take some time, so I'd recommend that you just run the class where you add the new tests :)

nelsonjr:
@nacx: argh!! the notifications for your reply got lost in my inbox and I did not see them. sorry for the delay! i'll get to it asap.

nacx:
@utkarsh-devops This PR has been quiet for some time. Feel free to step in and help with the remaining bits!

nacx:
Superseded by https://github.com/jclouds/jclouds/pull/1106.

PR #1007, opened 2016-09-07
Some handles were not properly closed

zack-shoylev:


apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -700,6 +700,7 @@ public void downloadBlob(String container, String name, File destination, Execut
 
          Futures.getUnchecked(Futures.allAsList(results));
 
+         raf.close();
```
nacx: This is redundant; it will reach the `finally` block immediately.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -700,6 +700,7 @@ public void downloadBlob(String container, String name, File destination, Execut
...
       } catch (IOException e) {
          // cleanup, attempt to delete large file
          if (raf != null) {
```
nacx: Same here. If you close it in the finally block, there is no need to close it here too.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -709,6 +710,12 @@ public void downloadBlob(String container, String name, File destination, Execut
          }
          destination.delete();
          throw new RuntimeException(e);
+      } finally {
+         if (raf != null) {
+            try {
+               raf.close();
+            } catch (IOException e1) {}
```
nacx: Use `Closeables2.closeQuietly` instead?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -701,14 +702,11 @@ public void downloadBlob(String container, String name, File destination, Execut
          Futures.getUnchecked(Futures.allAsList(results));
 
       } catch (IOException e) {
-         // cleanup, attempt to delete large file
-         if (raf != null) {
-            try {
-               raf.close();
-            } catch (IOException e1) {}
-         }
+         Closeables2.closeQuietly(raf);
```
nacx: Remove this and let it be closed in the finally block

zack-shoylev:
Thanks @nacx ! There's probably a few more things to patch up in this new code for the parallel stuff, I'll be working on it some more soon.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -819,10 +817,7 @@ public void run() {
                }
             }
             // Finished writing results to stream
-            try {
-               output.close();
-            } catch (IOException e) {
-            }
+            Closeables2.closeQuietly(output);
```
nacx: I've just seen a bunch of "System.out" in this class. I know it is not the scope of this PR, but it would be great to replace them by a logger!

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
zack-shoylev: That won't work, it's trying to delete the file for cleanup; it can't do that if it' open. Which kinda sucks.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
nacx: Also, would it make sense to enclose the entire `for` block in a try/finally block, to make sur everything is always closed, without having to do it explicitly in the middle of the code and concrete catch blocks?

zack-shoylev: Good idea

zack-shoylev:
merged

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
zack-shoylev: > enclose the entire for block in a try/finally block

I merged the fixes for now, I will see if I can do a somewhat more complex refactoring in a separate PR later. I wanted to spend some time looking at reusing this code for BaseBlobstore first, before I spend too much time on it.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
andrewgaul: You can do what @nacx suggests with slightly more complicated resource management:

```java
RandomAccessFile raf = null;
File tmpFile = null;
try {
   tempFile = new File(...);
   raf = new RandomAccessFile(tempFile);
   raf.write(...);
   raf.close();
   raf = null;
   tempFile.renameTo(destinationFile);
   tempFile = null;
} finally {
   Closeables2.closeQuietly(raf);
   if (tempFile != null) {
      tempFile.delete();
   }
}
```

This ensures that `destinationFile` is not overwritten unless the download succeeds.

zack-shoylev: Good point, I'll submit a quick PR for it and one other small fix.

zack-shoylev: https://github.com/jclouds/jclouds/pull/1010

PR #1009, opened 2016-09-12
Add keystone authentication with an existing token

olivierlemasle:
This pull request adds the possibility to use JClouds for OpenStack with a pre-existing Keystone token.
Cf. http://developer.openstack.org/api-ref/identity/v2/?expanded=authenticate-detail

To authenticate with a tenant name and a token, you can do:

```
Properties config = new Properties();
config.setProperty(KeystoneProperties.CREDENTIAL_TYPE, CredentialTypes.TOKEN_CREDENTIALS);
ContextBuilder.newBuilder("openstack-neutron")
                .overrides(config)
                .credentials("tenantname:", "hereisthetoken")
...
```

To authenticate with a tenant ID and a token, you can do:

```
Properties config = new Properties();
config.setProperty(KeystoneProperties.CREDENTIAL_TYPE, CredentialTypes.TOKEN_CREDENTIALS);
config.setProperty(KeystoneProperties.TENANT_ID, "tenantid");
ContextBuilder.newBuilder("openstack-neutron")
                .overrides(config)
                .credentials("", "hereisthetoken")
...
```

NB1: auto-value is added to the project `openstack-keystone` because this is JCloud's good practice (https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=49578594#API/ProviderWritingPractices-(Start)UsingAuto-Valueforallvaluetypes)

NB2: `@CredentialType` is made `@Inherited` because the AutoValue concrete class must have the annotation.

### DIFF
--- a/apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/config/KeystoneAuthenticationModule.java
+++ b/apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/config/KeystoneAuthenticationModule.java
@@ -48,6 +48,7 @@ import org.jclouds.openstack.keystone.v2_0.domain.Access;
 import org.jclouds.openstack.keystone.v2_0.domain.Endpoint;
 import org.jclouds.openstack.keystone.v2_0.functions.AuthenticateApiAccessKeyCredentials;
 import org.jclouds.openstack.keystone.v2_0.functions.AuthenticatePasswordCredentials;
+import org.jclouds.openstack.keystone.v2_0.functions.AuthenticateTokenCredentials;
 import org.jclouds.openstack.keystone.v2_0.handlers.RetryOnRenew;
 import org.jclouds.openstack.keystone.v2_0.suppliers.LocationIdToURIFromAccessForTypeAndVersion;
 import org.jclouds.openstack.keystone.v2_0.suppliers.RegionIdToAdminURIFromAccessForTypeAndVersion;
@@ -219,6 +220,7 @@ public class KeystoneAuthenticationModule extends AbstractModule {
       Builder<Function<Credentials, Access>> fns = ImmutableSet.<Function<Credentials, Access>> builder();
       fns.add(i.getInstance(AuthenticatePasswordCredentials.class));
       fns.add(i.getInstance(AuthenticateApiAccessKeyCredentials.class));
+      fns.add(i.getInstance(AuthenticateTokenCredentials.class));
       return CredentialTypes.indexByCredentialType(fns.build());
    }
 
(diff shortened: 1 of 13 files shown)


olivierlemasle:
Hi @zack-shoylev 

Thanks for your comments.

Yes, I'm aware that the retry mechanisms will not work if the authentication is performed with a token. However, in some cases, this authentication type can still be useful, and if the token is invalid, an error should be raised.

Regarding the tests, what kind of tests were you expecting?
Should it be unit tests?
Should it be live functional tests, and in that case how can I get a working token?

zack-shoylev:
@olivierlemasle 

Yes, the code will still be quite useful for a few situations! I mostly wanted to add a comment to the code to ensure the limitations are documented. As for the tests, I think at least unit tests would be nice. For live tests, you can use devstack to test locally, but let's get the unit tests done first.

nacx:
> Yes, I'm aware that the retry mechanisms will not work if the authentication is performed with a token. 

In that case, we should better configure [this binding](https://github.com/olivierlemasle/jclouds/blob/master/apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/config/KeystoneAuthenticationModule.java#L193) **only if** the credential type is **not** `token`?

olivierlemasle:
@nacx You're right, and I've added a commit to address that. It still binds to `retryOnRenew` (we still have to use it for HTTP 408 errors), but I've modified this RetryHandler to skip retrying in the case of a token-based authentication.

@zack-shoylev I've added some basic unit tests. Please tell me if you had some specific tests on mind. Of course, I've also tested this with a live OpenStack installation (using OpenStack Kilo).

apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v2_0/config/KeystoneAuthenticationModule.java
```diff
@@ -234,6 +234,13 @@ public String get() {
       return authenticationMethods.get(credentialType);
    }
 
+   @Provides
+   @Singleton
+   @Named("credentialsRetryable")
+   protected final boolean credentialsRetryable(Function<Credentials, Access> getAccess) {
```
nacx: Better expose the `CredentialType` object instead of just a boolean

zack-shoylev: So the code could directly check the type of credentials provided instead of relying on the boolean? Makes sense.

nacx: Hmm... Nope, the code should call `credentialType.retryable()`, but it makes more sense to object a CredentialType object than a boolean named "something". The code is easier to read and understand if injecting the object (and easier to extend in the future).

olivierlemasle: @nacx You're right. I've updated the pull request following your advice (injecting `CredentialType`).

nacx:
Thanks! Changes LGTM. Just pending to add the rests for the `KeystonApiExpectTest` and its base class as @zack-shoylev suggested.

andrewgaul:
@olivierlemasle do we have a path forward on this pull request?

gaul:
Closed due to inactivity.

PR #1011, opened 2016-09-21
Better predicate signature for node operations

nacx:
This changes the signature of the compute methods to have better predicates for nodes. Without this, most methods could only use a `Predicate<ComputeMetadata>` or a `Predicate<NodeMetadata>`, which was pretty inconvenient, because generic predicates such as `NodePredicates.all()` or concrete ones like `NodePredicates.inGroup("")` could not be used in all methods.

This change should be backwards compatible regarding client code and will allow to use all node and generic predicates in all node methods.

/cc @andreaturli


nacx:
Pushed to master as [d98348d5](http://git-wip-us.apache.org/repos/asf/jclouds/commit/d98348d5).

PR #1012, opened 2016-09-22
Remove hardcoded content size for personality in Openstack Nova 

arvindn05:
Fix issue JCLOUDS-1175 by removing hardcoded content sizes in Openstack
Nova API


arvindn05:
@nacx Please review when you get a chance
https://issues.apache.org/jira/browse/JCLOUDS-1175

nacx:
Pushed to master as [34b54ad1](http://git-wip-us.apache.org/repos/asf/jclouds/commit/34b54ad1). Thanks @arvindn05!

zack-shoylev:
This is a good change. We should be moving toward server-side validation in general.

arvindn05:
in that case, should we also remove the check for maximum file path size being 255 bytes?
I left that in because that is not configurable in openstack.
checkArgument(
               path.getBytes().length < 255,
               String.format("maximum length of path is 255 bytes.  Path specified %s is %d bytes", path,
                     path.getBytes().length));

zack-shoylev:
I wouldn't be opposed, though I don't think refactoring old code to remove such checks is strictly necessary or high priority (unless actually needed).

PR #1014, opened 2016-09-26
rename oauth properties

andreaturli:
- oauth.{endpoint.identity, credential} to jclouds.oauth.{endpoint, identity, credential}

 @nacx what do you think? I think it is more homogenous this way


nacx:
:-1: jclouds alrady reads properties in the form: `<provider/api id>.<property name>`, so you can configure properties (endpoints, for example) for multiple providers:

* `aws-ec2.endpoint`
* `openstack-nova.endpoint`
* `openstack-nova.identity`

If no provider specific property is found, jclouds fallsback to the `jclouds` prefix. In this case `oauth` is the API ID, so it respects the property naming convention.

andreaturli:
thanks @nacx but I don't follow completely.

For example, to run tests in `jclouds-labs/azurecompute-arm` I need to specify:
```
mvn clean install -Plive -Dtest.azurecompute-arm.identity="identity" -Dtest.azurecompute-arm.credential="password" -Dtest.azurecompute-arm.endpoint="https://management.azure.com/subscriptions/subscription-id" -Dtest.oauth.endpoint="endpoint"
```
Is there a way to pass the last property as `-Dtest.jclouds.oauth.endpoint` ?

nacx:
Why do you want to pass it like that? I mean, you are providing properties of two different jclouds providers/apis: "azurecompute-arm" and "oauth". And you are passing both using the jclouds convention `<provider/api id>.endpoint`. Why do you want to change that add a prefix to one of the properties?

andreaturli:
because I thought it was jclouds's convention, thinking about stuff like
```
public static final String AUTO_ALLOCATE_FLOATING_IPS = "jclouds.openstack-nova.auto-create-floating-ips";
```
so for simmetry I'd have expected that for oauth things as well, no?

nacx:
Well, I'd say the wrong properties are the ones you point in this last comment, as they double-prefix the property. When jclouds creates the context, it tries to resolve the properties that are prefixed with the "provider id", and if not, it tries to resolve them with the "jclouds" prefix. You can see how this is done [here](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/ContextBuilder.java#L414).

This allows users to configure the same properties with different values for different providers, which is desirable if you run one single app that consumes several apis; the "identity" property should be different for each connection. Apart from the override properties, jclouds loads all properties prefixed with "jclouds" or the "provider id" that are configured as systme properties (see [here](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/ContextBuilder.java#L373)), so you can have global configuration variables for all providers.

So, in this case, "oauth" is the prefix, designating the endpoint the oauth api uses. It would not be correct to double-prefix it as "jclouds.oauth".

andreaturli:
thx @nacx for the explanation! I'll close the pr as not needed

PR #1016, opened 2016-10-05
JCLOUDS-945: fix local blobstore marker handling

andrewgaul:
Previously using prefix markers would not correctly find the next key.
Add test for this behavior.


andrewgaul:
@kahing Please review.

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
```diff
@@ -276,21 +276,11 @@ public StorageMetadata apply(String key) {
          if (options.getMarker() != null) {
             final String finalMarker = options.getMarker();
             String delimiter = storageStrategy.getSeparator();
-            Optional<StorageMetadata> lastMarkerMetadata;
-            if (finalMarker.endsWith(delimiter)) {
-               lastMarkerMetadata = tryFind(contents, new Predicate<StorageMetadata>() {
```
andrewgaul: I suspect this curious logic is a leftover from the deprecated directory handling.

PR #1017, opened 2016-10-07
JCLOUDS-1185 Allow rangeIPv4 to be null

stuarthendren:
The default network of recently created projects have a null value for rangeIPv4
And all regions as subnets, which do have values for the rangeIPv4.

Fixes JCLOUDS-1158


nacx:
Pushed to master as [26c060a0](http://git-wip-us.apache.org/repos/asf/jclouds/commit/26c060a0). Thanks @stuarthendren!

PR #1018, opened 2016-10-11
JCLOUDS-1179 This also addresses the same problem with other providers

zack-shoylev:


zack-shoylev:
https://issues.apache.org/jira/browse/JCLOUDS-1179

nacx:
LGTM but before merging please amend the commit message and write a proper explanation and add the relevant links, for reference.

zack-shoylev:
merged

PR #1021, opened 2016-10-13
Use file.separator in tests instead of slash

zack-shoylev:


blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseContainerIntegrationTest.java
```diff
@@ -584,12 +584,12 @@ public void testListMarkerPrefix() throws Exception {
       final String container = getContainerName();
       try {
          blobStore.createContainerInLocation(null, container);
-         blobStore.putBlob(container, blobStore.blobBuilder("a/a").payload("").build());
-         blobStore.putBlob(container, blobStore.blobBuilder("b/b").payload("").build());
-         ListContainerOptions options = new ListContainerOptions().afterMarker("b/").recursive();
+         blobStore.putBlob(container, blobStore.blobBuilder("a" + File.separator + "a").payload("").build());
```
andrewgaul: This is backwards -- the Filesystem blobstore should map the path separator and not the tests.

zack-shoylev: Valid point. I'll look into it some more tomorrow.

zack-shoylev:
https://github.com/jclouds/jclouds/pull/1024

PR #1022, opened 2016-10-19
Log information according to the auth precedence

nacx:
This basically changes the `toString()` method of the SSH connection classes so it matches the order in which credentials are taken into account: first public key, then password.
Otherwise, if the user set both, the public key was used to authenticate, but a "password based auth" was printed to the logs, which was very confusing.


nacx:
Merged to master as [edbb2c0e](http://git-wip-us.apache.org/repos/asf/jclouds/commit/edbb2c0e).

PR #1024, opened 2016-10-21
Ensure that jclouds filesystem provider consistently uses /

zack-shoylev:


apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -745,28 +743,26 @@ protected String buildPathStartingFromBaseDir(String... pathTokens) {
...
+    * @param path
     * @return
     */
-   private static String normalize(String pathToBeNormalized) {
-      if (null != pathToBeNormalized && pathToBeNormalized.contains(BACK_SLASH)) {
-         if (!BACK_SLASH.equals(File.separator)) {
-            return pathToBeNormalized.replace(BACK_SLASH, File.separator);
-         }
+   private static String normalize(String path) {
+      if (null != path) {
```
andrewgaul: Make this conditional for Windows?

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -745,28 +743,26 @@ protected String buildPathStartingFromBaseDir(String... pathTokens) {
...
-      if (null != pathToDenormalize && pathToDenormalize.contains("/")) {
-         if (BACK_SLASH.equals(File.separator)) {
-              return pathToDenormalize.replace("/", BACK_SLASH);
-         }
+   /**
+    * Convert path to jclouds standard (/)
+    */
+   private static String denormalize(String path) {
+      if (null != path) {
+         return path.replace("\\", "/");
```
andrewgaul: Same as above.

apis/filesystem/src/test/java/org/jclouds/filesystem/FilesystemBlobStoreTest.java
```diff
@@ -173,9 +173,9 @@ public void testList_NoOptionSingleContainer() throws IOException {
         checkForContainerContent(CONTAINER_NAME, null);
 
         // creates blobs in first container
-        Set<String> blobsExpected = TestUtils.createBlobsInContainer(CONTAINER_NAME, "bbb" + File.separator + "ccc"
-                + File.separator + "ddd" + File.separator + "1234.jpg", "4rrr.jpg", "rrr" + File.separator + "sss"
-                + File.separator + "788.jpg", "xdc" + File.separator + "wert.kpg");
+        Set<String> blobsExpected = TestUtils.createBlobsInContainer(CONTAINER_NAME, "bbb" + "/" + "ccc"
+                + "/" + "ddd" + "/" + "1234.jpg", "4rrr.jpg", "rrr" + "/" + "sss"
+                + "/" + "788.jpg", "xdc" + "/" + "wert.kpg");
```
andrewgaul: Collapse the no-longer needed `+`.

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -786,10 +782,11 @@ private String removeFileSeparatorFromBorders(String pathToBeCleaned, boolean on
 
       // search for separator chars
       if (!onlyTrailing) {
-         if (pathToBeCleaned.substring(0, 1).equals(File.separator))
+         if (pathToBeCleaned.substring(0, 1).equals("/") || pathToBeCleaned.substring(0, 1).equals("\\"))
```
andrewgaul: Call `charAt` instead?

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -786,10 +782,11 @@ private String removeFileSeparatorFromBorders(String pathToBeCleaned, boolean on
...
             beginIndex = 1;
       }
-      if (pathToBeCleaned.substring(pathToBeCleaned.length() - 1).equals(File.separator))
+      if (pathToBeCleaned.substring(pathToBeCleaned.length() - 1).equals("/") ||
```
andrewgaul: Same as above.

andrewgaul:
Overall looks good, this fixes the source-code incompatibility for users on Windows.

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
zack-shoylev: This will also normalize the path on a Linux system, so I would rather leave it as is

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
zack-shoylev: I rather keep these rules OS-agnostic. Does this make sense?

zack-shoylev:
merged

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
andrewgaul: Windows has other restricted characters:

http://stackoverflow.com/a/31976060/2800111

We should not strive to make the filesystem blobstore work as similar on Linux as Windows by crippling the former.  Rather we should make the filesystem blobstore as closely match real blobstores as the environment allows.

zack-shoylev: But then it is not cross-platform. I mean, it's fairly easy to put the exception in, but a real fix will probably be somewhat more complex (i.e. hex-encoding the blob name). But I do get we want to keep this a filesystem provider, so it makes sense to have it follow the filesystem closely.

PR #1025, opened 2016-10-23
JCLOUDS-1187: Do not load Json payloads to memory when serializing objects

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1187


core/src/main/java/org/jclouds/json/internal/GsonWrapper.java
```diff
@@ -46,6 +51,27 @@ public GsonWrapper(Gson gson) {
...
    }
+   
+   @SuppressWarnings("unchecked")
+   @Override
+   public <T> T fromJson(InputStream json, Type type) {
+      Reader reader = new InputStreamReader(json);
+      try {
+         return (T) gson.fromJson(reader, type);
+      } finally {
+         closeQuietly(reader);
```
andrewgaul: Should the caller do the resource management instead of the callee?

nacx: Yes, much better. Done!

nacx:
Merged to master

PR #1028, opened 2016-10-26
Change types of variables in docker ImageSummary

dudaerich:
The int type is not sufficient for size and virtualSize
parameters, if the image is bigger than 2gb.

This fixes https://issues.apache.org/jira/browse/JCLOUDS-1190


nacx:
Merged to master as [dff16876](http://git-wip-us.apache.org/repos/asf/jclouds/commit/dff16876). Thanks @dudaerich!

PR #1029, opened 2016-10-26
filesystem: Modify the backslash only on Windows

zack-shoylev:


apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -750,7 +750,10 @@ protected String buildPathStartingFromBaseDir(String... pathTokens) {
     */
    private static String normalize(String path) {
       if (null != path) {
-         return path.replace("/", File.separator).replace("\\", File.separator);
+         if( isWindows() ) {
```
andrewgaul: Checkstyle violation.

zack-shoylev:
merged

PR #1033, opened 2016-11-03
org/jclouds/scriptbuilder/functionloader/osgi/BundleFunctionLoader.java:

wltjr:
Switched from Properties to HashMap to fix compilation issues. Where
Properties is not of type Dictionary<String, ?>. This passes the right
type to bundleContext.registerService. Also corrects the first argument,
String -> String[], same issue wrong type passed.

This was need to allow it to build, I created a patch for the 
[Gentoo ebuild](https://github.com/Obsidian-StudiosInc/os-xtoo/tree/master/dev-java/jclouds-scriptbuilder)

### DIFF
--- a/scriptbuilder/src/main/java/org/jclouds/scriptbuilder/functionloader/osgi/BundleFunctionLoader.java
+++ b/scriptbuilder/src/main/java/org/jclouds/scriptbuilder/functionloader/osgi/BundleFunctionLoader.java
@@ -18,8 +18,9 @@ package org.jclouds.scriptbuilder.functionloader.osgi;
 
 import java.io.IOException;
 import java.net.URL;
+import java.util.Dictionary;
 import java.util.Enumeration;
-import java.util.Properties;
+import java.util.HashMap;
 
 import org.jclouds.scriptbuilder.domain.OsFamily;
 import org.jclouds.scriptbuilder.domain.ShellToken;
@@ -98,9 +99,10 @@ public class BundleFunctionLoader implements FunctionLoader {
    }
 
    private void registerFunction(String functions) {
-      Properties properties = new Properties();
-      properties.put("function", functions);
-      registration = bundleContext.registerService(FunctionLoader.class.getName(), this, properties);
+      String[] className = { FunctionLoader.class.getName() };
+      Dictionary dictionary = new HashMap<String, Object>();
+      dictionary.put("function", functions);
+      registration = bundleContext.registerService(className, this, dictionary);
    }
 
 }


wltjr:
Here is the error output without the patch

src/main/java/org/jclouds/scriptbuilder/functionloader/osgi/BundleFunctionLoader.java:103: error: no suitable method found for registerService(String,BundleFunctionLoader,Properties)
      registration = bundleContext.registerService(FunctionLoader.class.getName(), this, properties);
                                  ^
    method BundleContext.registerService(String[],Object,Dictionary<String,?>) is not applicable
      (argument mismatch; String cannot be converted to String[])
    method BundleContext.registerService(String,Object,Dictionary<String,?>) is not applicable
      (argument mismatch; Properties cannot be converted to Dictionary<String,?>)
    method BundleContext.<S#1>registerService(Class<S#1>,S#1,Dictionary<String,?>) is not applicable
      (cannot infer type-variable(s) S#1
        (argument mismatch; String cannot be converted to Class<S#1>))
    method BundleContext.<S#2>registerService(Class<S#2>,ServiceFactory<S#2>,Dictionary<String,?>) is not applicable
      (cannot infer type-variable(s) S#2
        (argument mismatch; String cannot be converted to Class<S#2>))
  where S#1,S#2 are type-variables:
    S#1 extends Object declared in method <S#1>registerService(Class<S#1>,S#1,Dictionary<String,?>)
    S#2 extends Object declared in method <S#2>registerService(Class<S#2>,ServiceFactory<S#2>,Dictionary<String,?>)

wltjr:
I switched to HashMap and forced pushed. Still seems to have same error of suggesting HashMap. I can correct if needed.

nacx:
Ok. The issue seems to appear because you depend on [osgi.core v6](https://github.com/Obsidian-StudiosInc/os-xtoo/blob/master/dev-java/jclouds-scriptbuilder/jclouds-scriptbuilder-1.9.2-r1.ebuild#L25), while the code assumes [version 4.2.0](https://github.com/jclouds/jclouds/blob/master/scriptbuilder/pom.xml#L62).

In version 6, the `BundleContex.registerService` signature changes the type of the lat parameter from `Dictionary` (raw type) to `Dictionary<String, ?>`, and since the `Properties` object extends `HasMap<Object, Object>` it no longer matches the signature.

The build fails with your changes (see [this](https://jclouds.ci.cloudbees.com/job/jclouds-pull-requests/2473/org.apache.jclouds$jclouds-scriptbuilder/console) for details), so changes are needed.

* Can you just make the ebuild depend on osgi core 4.2.0?
* It would be great to extract the osgi core version to jclouds/project/pom.xml so we can easily run a build with a custom version of the osgi framework. This would allow us to create a build compatibility matrix and add it to our CI.

nacx:
@wltjr I've opened https://github.com/jclouds/jclouds/pull/1034 and added some fixes to your work (it includes your commit and I've preserved the authorship). I think we can close this one now?

wltjr:
I thought the error was the same with osgi 4, 5, and 6. I can retest but you all likely know better there. OSGi is slotted on Gentoo so I can use any version, 4, 5 or 6. Though idea is always to use the latest unless need/reason to use lesser. Also in other cases can help further upstream by providing patches for newer stuff such as I did. I ran into a similar issue with jclouds-core and guava 20. I will see about making a patch for that as well.

wltjr:
Double checked with OSGi 4.3.1 and same issue there, as with 5, and 6. Definitely same sig in [4.3.1](https://osgi.org/javadoc/r4v43/core/index.html) as in 5 and 6. It changed in 4.x versions, as [4.2](https://osgi.org/javadoc/r4v42/index.html) has the old one and 4.3 new. You would think they would reserve such changes for 5, but guess not. Anyway this makes it good from 4.3.1 up to 6, so should not be an issue again till 7.x in theory. Unless they make a change in new versions of 4-6. Strange stuff...

Thanks for the modification and commit, no worries on preserving authorship but thanks for that! 👍 
More to come!

nacx:
>Thanks for the modification and commit, no worries on preserving authorship but thanks for that! 👍 
More to come!

Thanks to you for your patches! Contributions are very welcome!

PR #1034, opened 2016-11-03
Make jclouds compatible with newer OSGi framework versions

nacx:
This is based on https://github.com/jclouds/jclouds/pull/1033.

This PR fixes the build of the previous one and makes the OSGi framework version configurable, so we can run builds for different environments. The scope of the OSGi dependencies is `provided`, so we can expect target platforms to have different versions than the ones used to build jclouds.

With this change, the build succeeds with the base version, and also with the latest versions:

    mvn clean install -Dosgi.version=6.0.0 -Dosgi.compendium.version=5.0.0

### DIFF
--- a/resources/modernizer_exclusions.txt
+++ b/resources/modernizer_exclusions.txt
@@ -20,3 +20,4 @@ com/google/common/primitives/Longs.compare:(JJ)I
 com/google/inject/Inject
 com/google/inject/Provider
 java/lang/StringBuffer."<init>":()V
+java/util/Hashtable."<init>":(I)V
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -89,13 +89,11 @@
     <dependency>
       <groupId>org.osgi</groupId>
       <artifactId>org.osgi.core</artifactId>
-      <version>4.2.0</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.osgi</groupId>
       <artifactId>org.osgi.compendium</artifactId>
-      <version>4.2.0</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
--- a/core/src/test/java/org/jclouds/osgi/BundlesTest.java
+++ b/core/src/test/java/org/jclouds/osgi/BundlesTest.java
@@ -34,14 +34,15 @@ import com.google.common.collect.ImmutableSet;
 
 public class BundlesTest {
 
+   @SuppressWarnings("rawtypes")
    @Test
    public void testInstantiateAvailableClassesWhenAllAssignable() throws ClassNotFoundException {
       Bundle bundle = createMock(Bundle.class);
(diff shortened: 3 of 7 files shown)


resources/modernizer_exclusions.txt
```diff
@@ -20,3 +20,4 @@ com/google/common/primitives/Longs.compare:(JJ)I
 com/google/inject/Inject
 com/google/inject/Provider
 java/lang/StringBuffer."<init>":()V
+java/util/Hashtable."<init>":(I)V
```
nacx: @andrewgaul Note that the signatures of the OSGi framework require a `Dictionary<String, ?>`, and the properties object is no longer compatible, as it extends `Hashtable<Object, Object>`. We need to pass a proper dictionary there, thus this exclusion (`HashMap` does not extend `Dictionary`).

nacx: No, I meant `HashMap`. The modernizer plugin forbids the use of Hashtable in favor of HashMap, but HashMap does not extend Dictionary. Thus it can't be used as an argument to the BundleContext.registerService method. That's why I just excluded it from modernizer, as we now actually need a Hashtable instance. See [this comment](https://github.com/jclouds/jclouds/pull/1033#issuecomment-258078992) for further details.

nacx:
Changes are made to accommodate the tests to the signatures of those methods in newer OSGi framework versions..

Basically, they change from Dictionary (raw) to Dictionary<String, ?>, and in the case of the `loadClass`method, it returns a Class (raw) object in 4.2.0 but a Class<?> in 6.0.0.

The change sin this PR make the osgi versions configurable (I've already created a jenkins build that uses the latest osgi 6.0.0) and safely accommodates the method signatures.

resources/modernizer_exclusions.txt
demobox: > The modernizer plugin forbids the use of Hashtable in favor of HashMap, but HashMap does not extend Dictionary

Ah, like so. Would it be possible/make sense to apply this exclusion to a more limited scope, or do we need to create `Dictionary` objects also in jclouds core for this?

Thanks for explaining, @nacx!

nacx: Unfortunately no :( The modernizer plugin just reads the exclusions but you cannot configure them "just for one package or class".
We could override the configuration just for the scriptbuilder project, but if at some point we need to use the registerService in core, we'll have the same issue. I think the exclusion makes sense there, globally, to ensure we can use the OSGi 6 framework method signatures where needed.

andrewgaul: Unfortunate limitation of Modernizer:

https://github.com/andrewgaul/modernizer-maven-plugin/issues/3

demobox: > I think the exclusion makes sense there, globally, to ensure we can use the OSGi 6 framework method signatures where needed.

Makes sense - thanks for the additional explanations, @nacx and @andrewgaul!

PR #1035, opened 2016-11-07
Use temporary file for Swift test inputs

andrewgaul:
This avoids creating junk in the source directory.

### DIFF
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
@@ -679,7 +679,7 @@ public class RegionScopedSwiftBlobStore implements BlobStore {
 
       ListeningExecutorService listeningExecutor = MoreExecutors.listeningDecorator(executor);
       RandomAccessFile raf = null;
-      File tempFile = new File(destination.getName() + "." + UUID.randomUUID());
+      File tempFile = new File(destination + "." + UUID.randomUUID());
       try {
          long contentLength = api
                .getObjectApi(regionId, container)
--- a/apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStoreParallelLiveTest.java
+++ b/apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStoreParallelLiveTest.java
@@ -56,7 +56,7 @@ import com.google.common.util.concurrent.MoreExecutors;
 @Test(groups = "live", singleThreaded = true)
 public class RegionScopedSwiftBlobStoreParallelLiveTest extends BaseBlobStoreIntegrationTest {
 
-   private final File BIG_FILE = new File("random.dat");
+   private final File BIG_FILE;
    private final long SIZE = 10 * 1000 * 1000;
    private BlobStore blobStore;
    private String ETAG;
@@ -71,6 +71,11 @@ public class RegionScopedSwiftBlobStoreParallelLiveTest extends BaseBlobStoreInt
 
    public RegionScopedSwiftBlobStoreParallelLiveTest() {
       provider = "openstack-swift";
+      try {
+         BIG_FILE = File.createTempFile("random", "dat");
+      } catch (IOException ioe) {
+         throw new RuntimeException(ioe);
+      }
    }
 
    // Override as needed for the right region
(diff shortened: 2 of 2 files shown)


andrewgaul:
@zack-shoylev Please review.

demobox:
+1 - looks good to me. Thanks for the cleanup, @andrewgaul!

Had a quick look at `.gitignore` - looks like there are no changes needed/possible there as a result of this.

andrewgaul:
I discovered a more serious issue in `RegionScopedSwiftBlobStore.downloadBlob` which creates a temporary file with a relative path.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -679,7 +679,7 @@ public void downloadBlob(String container, String name, File destination, Execut
 
       ListeningExecutorService listeningExecutor = MoreExecutors.listeningDecorator(executor);
       RandomAccessFile raf = null;
-      File tempFile = new File(destination.getName() + "." + UUID.randomUUID());
+      File tempFile = new File(destination + "." + UUID.randomUUID());
```
demobox: Is `destination` here an absolute known/assumed to be an absolute path?

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStoreParallelLiveTest.java
```diff
@@ -121,7 +126,7 @@ public void uploadMultipartBlob() {
 
    @Test(dependsOnMethods = "uploadMultipartBlob", singleThreaded = true)
    public void downloadParallelBlob() throws IOException {
-      final File downloadedFile = new File(BIG_FILE.getName() + ".downloaded");
+      final File downloadedFile = new File(BIG_FILE + ".downloaded");
```
demobox: I was going to suggest making this name random across test runs, but given the way `BIG_FILE` is created, that should already be the case? Just wanting to make sure I'm not missing something here.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
andrewgaul: Given "/tmp/foo" `getName` returns "foo", ensuring the file will be relative to wherever the user launched the jclouds application.  With this change jclouds will create the temporary file in the same directory as the user specified.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStoreParallelLiveTest.java
andrewgaul: Again, `getName` returns only the file and not the complete path.

demobox: Yes, sorry, my comment wasn't clear: I was wondering whether there's a risk here - even with an absolute path - that two test runs would write the same file. But I don't think that's the case.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
demobox: > With this change jclouds will create the temporary file in the same directory as the user specified.

Clear - thanks. My question was more: do we know that `destination` here will indeed be an absolute path?

andrewgaul: `destination` need not be absolute -- we want to make the temporary file *relative* to `destination`, not whatever path the application launched from.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStoreParallelLiveTest.java
andrewgaul: Not the case because each test method gets its own `BIG_FILE`.  Further these tests run single-threaded due to the annotation and the dependency.

demobox:
+1 - looks good to me. Thanks for the additional explanations, @andrewgaul!

zack-shoylev:
+1

PR #1036, opened 2016-11-08
Fix for Java 1.8 and Guice 4.1

wltjr:
core/src/main/java/org/jclouds/rest/internal/TransformerForRequest.java:
Fix for Java 1.8 and Guice 4.1. Not sure if this is correct, but does
compile. Committing to see if it passes tests to confirm fix or not.

Attempted fix for [JCLOUDS-1194](https://issues.apache.org/jira/browse/JCLOUDS-1194)


core/src/main/java/org/jclouds/rest/internal/TransformerForRequest.java
```diff
@@ -82,7 +82,7 @@
    public Function<HttpResponse, ?> apply(HttpRequest in) {
       GeneratedHttpRequest request = GeneratedHttpRequest.class.cast(in);
       Function<HttpResponse, ?> transformer;
-      Class<? extends HandlerWithResult<?>> handler = getSaxResponseParserClassOrNull(request.getInvocation()
+      Class<? extends HandlerWithResult> handler = getSaxResponseParserClassOrNull(request.getInvocation()
```
demobox: Does this generate a warning of some kind?

wltjr:
I did not get any during compile

```bash
>>> Preparing source in /tmp/portage/dev-java/jclouds-core-9999/work/jclouds-core-9999/core ...
 * Applying jclouds-core-0-guice_java_1.8.patch ...                      [ ok ]
Searching for bundled jars:
None found.
Searching for bundled classes (no output if none found):
Search done.
>>> Source prepared.
>>> Configuring source in /tmp/portage/dev-java/jclouds-core-9999/work/jclouds-core-9999/core ...
>>> Source configured.
>>> Compiling source in /tmp/portage/dev-java/jclouds-core-9999/work/jclouds-core-9999/core ...
 * Compiling ...
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
>>> Source compiled.
>>> Test phase [not enabled]: dev-java/jclouds-core-9999
```

wltjr:
Sure thing, I was unsure of the fix as it did seem odd, but seems to work till it doesn't.

nacx:
Just getting back here. @wltjr which JDK version are you using? The build seems already OK in OpenJDK8 and Guice 4.1.0 without this patch:
https://jclouds.ci.cloudbees.com/job/jclouds-compat/GUAVA_VERSION=16.0.1,GUICE_VERSION=4.1.0,jdk=OpenJDK%208%20(latest)/40/

wltjr:
I was using Oracle JDK 1.8. I can check again to see if it is still there. Source/target 1.8

core/src/main/java/org/jclouds/rest/internal/TransformerForRequest.java
andrewgaul: Removing this should generate a warning, not an error.

wltjr:
Going to close this, I think the patch is still needed on Gentoo for some reason. But if not needed else where its moot.

nacx:
@wltjr Could you confirm that the issue is still there with the Oracle JDK 1.8? Which concrete version are you using, 1.8.0_102 as mentioned in the JIRA issue?
I wouldn't like to close this issue without understanding why it fails in your environment. It looks like an issue in the type inference on your JDK version.

Instead of removing the generic from the variable declaration, could you try declaring a variable so the type system has more information to infer the types it needs? Something like:

```java
Class<? extends HandlerWithResult<?>> handler = getSaxResponseParserClassOrNull(request.getInvocation().getInvokable());
if (handler != null) {
   HandlerWithResult<?> h = injector.getInstance(handler);
   transformer = parserFactory.create(h);
}
```

This way we explicitly set the type for the `injector.getInstance` result. Let's see if this helps your compiler. If it works I think it is a better fix than removing the generic. Could you give this a try?

wltjr:
Yes it still fails without patch with latest 1.8 JDK from Oracle. I can try via JDK 9. I will see about trying out the modifications you suggested and seeing if that has any effect. The environment on Gentoo is a bit different. It maybe related to my [guice ebuild](https://github.com/Obsidian-StudiosInc/os-xtoo/blob/master/dev-java/guice/guice-4.1.ebuild), how I am building it on Gentoo. I use some newer stuff.

```
GENTOO_VM=oracle-jdk-bin-1.8  CLASSPATH="" JAVA_HOME="/opt/oracle-jdk-bin-1.8.0.144"
JAVACFLAGS="-source 1.8 -target 1.8" COMPILER=""

src/main/java/org/jclouds/rest/internal/TransformerForRequest.java:88: error: method create in interface Factory cannot be applied to given types;
         transformer = parserFactory.create(injector.getInstance(handler));
                                    ^
  required: HandlerWithResult<T#1>
  found: CAP#1
  reason: inference variable T#2 has incompatible bounds
    equality constraints: CAP#2
    upper bounds: HandlerWithResult<CAP#3>,Object
  where T#1,T#2 are type-variables:
    T#1 extends Object declared in method <T#1>create(HandlerWithResult<T#1>)
    T#2 extends Object declared in method <T#2>getInstance(Class<T#2>)
  where CAP#1,CAP#2,CAP#3 are fresh type-variables:
    CAP#1 extends HandlerWithResult<?> from capture of ? extends HandlerWithResult<?>
    CAP#2 extends HandlerWithResult<?> from capture of ? extends HandlerWithResult<?>
    CAP#3 extends Object from capture of ?
```

wltjr:
Really strange, your suggested test modifications make it compile... Just the same as my patch. Though I think your approach is more correct. Maybe it gets messed up with it being compounded like that. By breaking it out and re-assigning it maybe makes it the proper type.

I think your suggested test/change should maybe committed. I can push it as an update to my PR or you can modify either way. Its your change that allows it to build.

nacx:
>I think your suggested test/change should maybe committed. I can push it as an update to my PR or you can modify either way. Its your change that allows it to build.

Good to see it worked! It would be great if you could reopen the PR and submit the changes. Let's get this fixed! :)

And thanks for all your testing! It is great to have feedback from downstream package maintainers. Thanks!

wltjr:
Sure thing, and since I forced pushed I must open a new PR. I will reference this one. Still need to see about contributing sometime. Got sucked into C coding :)

PR #1037, opened 2016-11-09
Include part size in listMultipartUpload result

andrewgaul:
Clients use this during `completeMultipartUpload`.


andrewgaul:
This causes integration test failures since `uploadMultipartPart` does not correctly return `partSize`

zack-shoylev:
+1

PR #1038, opened 2016-11-14
Missing cleanup was causing testReplaceManifest to fail

zack-shoylev:


apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/StaticLargeObjectApiLiveTest.java
```diff
@@ -117,6 +118,7 @@ public void testDeleteSinglePartObjectWithMultiPartDelete() throws Exception {
          assertThat(resp.deleted()).isZero();
          assertThat(resp.notFound()).isZero();
          assertThat(resp.errors()).hasSize(1);
+         api.getObjectApi(regionId, containerName).delete(objectName);
```
demobox: @zack-shoylev Is it the intention that this only be called if the assertions succeed?

zack-shoylev: Yeah, otherwise cleanup takes care of it.

demobox: > Yeah, otherwise cleanup takes care of it.

Ah, OK. In that case, I'm probably confused: if cleanup takes care of it, why do we explicitly need to delete it here? Is that more efficient? Or do downstream tests fail otherwise? Or..?

zack-shoylev:
merged

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/StaticLargeObjectApiLiveTest.java
demobox: > but unfortunately other tests here (testreplacemanifest) will fail if it is not cleaned up on success.

Thanks for clarifying, Zack!

PR #1013, opened 2016-09-23
1182: Added Seoul and Mumbai regions to AWS

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1182

PR #1040, opened 2016-11-17
JCLOUDS-1182: Added Seoul and Mumbai regions to AWS

andreaturli:
Backports https://github.com/jclouds/jclouds/pull/1013


andreaturli:
rebuild please

PR #1041, opened 2016-11-19
Updates the default template so live tests pass again.

zack-shoylev:


apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/ImageToOperatingSystem.java
```diff
@@ -80,9 +80,9 @@ public OperatingSystem apply(final Image from) {
             is64Bit = matcher.group(2).equals("x64");
          }
       } else {
-         if (imageName.contains("Red Hat EL")) {
+         if (imageName.contains("Red Hat")) {
```
zack-shoylev: Not completely sure about this one, though.

demobox: @zack-shoylev Do we happen to know why `EL` was chosen in the first place?

nacx: It would be good to know (and I'd like to :)), but it looks like a safe change anyway?

demobox: If we do find out the reason (other than "this one seemed to work" ;-)), perhaps add a comment to explain?

zack-shoylev: Well, the imageName string changed from "EL" to "Enterprise Linux" at some point, which is why the match was failing, but I don't know the original reasoning for being that specific. Seems safe, though.

zack-shoylev:
merged

andreaturli:
backport to [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/c2a4b8af) as well

PR #1042, opened 2016-11-26
Fixes JCLOUDS-1206

felfert:
See https://issues.apache.org/jira/browse/JCLOUDS-1206


felfert:
@nacx I looked into the tag api documentation and I think, it does **not** fit the jclouds semantic. In jclouds, those are `Map<String,String>`, but in DigitalOcean they have Map<String,Object> where Object is currently always a droplet. (Most likely, it's an Integer-Id, just like with their key API).

I'm going to change make the proposed change in the PR, but while we are here:

The private Json member is not needed anymore, but I did not remove it, because I am not shure about the implications of changing the signature of the constructor in line 75. Perhaps **you** can comment on that.

Also, Just out of curiosity:
The others (meaning: at least openstack-nova and aws-ec2) have a completely different API in jclouds: They implement a separate method in their TemplateOptions specializations: [userData(byte[] userData)](https://jclouds.apache.org/reference/javadoc/1.9.x/org/jclouds/openstack/nova/v2_0/compute/options/NovaTemplateOptions.html#userData(byte[])). So: why is it that the digitaloceans2 provider follows a different approach here? Resulting code in the jenkins plugin:
```java
if (!isNullOrEmpty(userData)) {
    try {
        if (options instanceof DigitalOcean2TemplateOptions) {
            options.userMetadata("user_data", userData);
        } else {
            Method userDataMethod = options.getClass().getMethod("userData", new byte[0].getClass());
            LOGGER.finest("Setting userData to " + userData);
            userDataMethod.invoke(options, userData.getBytes(StandardCharsets.UTF_8));
        }
    } catch (Exception e) {
        LOGGER.log(Level.WARNING, "userData is not supported by provider options class " + options.getClass().getName(), e);
    }
}
```

felfert:
Later: Just realized, that the tags are a Set<String> and associated with a node so: Of course that makes sense :)

nacx:
>The others (meaning: at least openstack-nova and aws-ec2) have a completely different API in jclouds: They implement a separate method in their TemplateOptions specializations: userData(byte[] userData). So: why is it that the digitaloceans2 provider follows a different approach here? Resulting code in the jenkins plugin

Right. I think the use of user metadata here is wrong. It is meant for providers like OpenStack that support key value pairs in the metadata, but not for this. Mind changing it and add the `userData` field to the DigitalOceanTemplateOptions, and also [this comment](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/test/java/org/jclouds/digitalocean2/compute/DigitalOcean2ComputeServiceLiveTest.java#L63) to properly indicate user metadata (in the form of arbitrary key value pairs) is not supported? Thanks!

nacx:
FTR, I've opened [JCLOUDS-1211](https://issues.apache.org/jira/browse/JCLOUDS-1211) to track the tags support.

felfert:
> Right. I think the use of user metadata here is wrong. It is meant for providers like OpenStack that support key value pairs in the metadata, but not for this. Mind changing it and add the userData field to the DigitalOceanTemplateOptions, and also this comment to properly indicate user metadata (in the form of arbitrary key value pairs) is not supported? Thanks!

Done. I intentionally left the old "wrong" variant via the userMetaData map in place in order to not break existing code.

providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/compute/DigitalOcean2ComputeServiceAdapter.java
```diff
@@ -96,16 +106,15 @@
...
+      // In DigitalOcean, user_data is a SINGLE string, NOT a map!
+      // Encoding tags or anything else than user_data in here breaks their functionality.
+      if (null != templateOptions.getUserData()) {
+         setUserDataIfSupported(template, options, new String(templateOptions.getUserData()));
+      }
+      // Backwards compatible variant, getting userData from userMetaData map.
+      Map<String, String> metadataAndTags = templateOptions.getUserMetadata();
+      if (null != metadataAndTags.get("user_data")) {
+         setUserDataIfSupported(template, options, metadataAndTags.get("user_data"));
       }
```
nacx: Set this in an `else` block for the first conditional to only call the backwards compatible logic if the new method has not been configured?

nacx:
I've squashed and pushed your changes to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/92044e3d) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/3ae2213a), and also added a commit in both branches to address my last comments. Thanks @felfert!

PR #1044, opened 2016-12-19
Change BaseSwiftApiLiveTest to inherit from BaseBlobStoreIntegrationT…

zack-shoylev:
…est and use its methods

Based on an old suggestion from https://github.com/jclouds/jclouds/pull/935


zack-shoylev:
merged

PR #1046, opened 2017-01-03
JCLOUDS-1220 : Managing the header name in the TempAuth (Identity Protocol v1)

Etienne-Carriere:
State of jclouds
The openstack swift "official" client (in python) manage this v1 protocol (http://docs.openstack.org/developer/python-swiftclient/swiftclient.html) so even if we don't have a specification, we will use the code of the official client for the implementation.

In jclouds, there is currently a sort-of v1 identity protocol in the openstack-swift module: in apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/config/SwiftAuthenticationModule.java, there is a tempAuthCredentials which is almost the identity v1 protocol except that the name of the headers was not the same :
= X-Storage-User vs X-Auth-User
= X-Storage-Path vs X-Auth-Key

Pull request :
= Keep the current behaviour as default
= Add 2 parameters to change the header name through variables in the Properties put in the Builder like that :

Properties overrides = new Properties();
overrides.setProperty("jclouds.keystone.credential-type", "tempAuthCredentials");
overrides.setProperty("jclouds.swift.tempAuth.headerUser", "X-Auth-User");
overrides.setProperty("jclouds.swift.tempAuth.headerPass", "X-Auth-Pass");
swiftApi = ContextBuilder.newBuilder(provider)
.endpoint(args[1])
.credentials(identity, credential)
.modules(modules)
.overrides(overrides)
.buildApi(SwiftApi.class);

### DIFF
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/binders/TempAuthBinder.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/binders/TempAuthBinder.java
@@ -33,16 +33,14 @@ import com.google.inject.name.Named;
  *
  */
 public final class TempAuthBinder implements Binder{
-   @Inject
-   @Named(TEMP_AUTH_HEADER_USER)
-   private String identityHeaderNameUser;
+   private final String identityHeaderNameUser;
+   private final String identityHeaderNamePass;
 
    @Inject
-   @Named(TEMP_AUTH_HEADER_PASS)
-   private String identityHeaderNamePass;
-
-
-
+   TempAuthBinder(@Named(TEMP_AUTH_HEADER_USER) String identityHeaderNameUser, @Named(TEMP_AUTH_HEADER_PASS) String identityHeaderNamePass) {
+      this.identityHeaderNameUser = identityHeaderNameUser;
+      this.identityHeaderNamePass = identityHeaderNamePass;
+   }
 
    @Override 
    public <R extends HttpRequest> R bindToRequest(R request, Object input) {
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/binders/TempAuthBinder.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/binders/TempAuthBinder.java
@@ -18,23 +18,38 @@ package org.jclouds.openstack.swift.v1.binders;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static org.jclouds.openstack.swift.v1.reference.TempAuthHeaders.TEMP_AUTH_HEADER_USER;
+import static org.jclouds.openstack.swift.v1.reference.TempAuthHeaders.TEMP_AUTH_HEADER_PASS;
 
 import org.jclouds.domain.Credentials;
 import org.jclouds.http.HttpRequest;
-import org.jclouds.openstack.swift.v1.config.SwiftAuthenticationModule;
 import org.jclouds.rest.Binder;
+
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
 
 /**
  * Binder to the tempAuthAuthentication
  *
  */
-public   final class TempAuthBinder implements Binder{
+public final class TempAuthBinder implements Binder{
+   @Inject
+   @Named(TEMP_AUTH_HEADER_USER)
+   private String identityHeaderNameUser;
+
+   @Inject
+   @Named(TEMP_AUTH_HEADER_PASS)
+   private String identityHeaderNamePass;
+
+
+
+
    @Override 
    public <R extends HttpRequest> R bindToRequest(R request, Object input) {
       checkNotNull(request, "request");
       checkArgument(input instanceof Credentials, "input must be a non-null org.jclouds.domain.Credentials");
-      return (R) request.toBuilder().replaceHeader(SwiftAuthenticationModule.getIdentityHeaderName(), ((Credentials) input).identity)
-	      		.replaceHeader(SwiftAuthenticationModule.getIdentityHeaderPass(), ((Credentials) input).credential).build();
+      return (R) request.toBuilder().replaceHeader(identityHeaderNameUser, ((Credentials) input).identity)
+	      		.replaceHeader(identityHeaderNamePass, ((Credentials) input).credential).build();
    }
 }
 
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/binders/TempAuthBinder.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/binders/TempAuthBinder.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.openstack.swift.v1.binders;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import org.jclouds.domain.Credentials;
+import org.jclouds.http.HttpRequest;
+import org.jclouds.openstack.swift.v1.config.SwiftAuthenticationModule;
+import org.jclouds.rest.Binder;
+
+/**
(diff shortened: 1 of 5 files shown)


apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/config/SwiftAuthenticationModule.java
```diff
@@ -62,22 +64,40 @@
...
+   public static final String TEMP_AUTH_HEADER_USER = "jclouds.swift.tempAuth.headerUser";
+   public static final String TEMP_AUTH_HEADER_PASS = "jclouds.swift.tempAuth.headerPass";
+
+   @com.google.inject.Inject(optional = true)
+   @Named(TEMP_AUTH_HEADER_USER)
+   private static String identityHeaderNameUser = "X-Storage-User";
+
+   @com.google.inject.Inject(optional = true)
+   @Named(TEMP_AUTH_HEADER_PASS)
+   private static String identityHeaderNamePass = "X-Storage-Pass";
```
nacx: Move these variables to the `TempAuthBinder ` class.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/binders/TempAuthBinder.java
```diff
@@ -0,0 +1,40 @@
...
+import org.jclouds.domain.Credentials;
+import org.jclouds.http.HttpRequest;
+import org.jclouds.openstack.swift.v1.config.SwiftAuthenticationModule;
+import org.jclouds.rest.Binder;
+
+/**
+ * Binder to the tempAuthAuthentication
+ *
+ */
+public   final class TempAuthBinder implements Binder{
```
nacx: [style] Remove the extra spaces and add a space before the last `{`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/config/SwiftAuthenticationModule.java
```diff
@@ -62,22 +64,40 @@
...
+
    private static final String STORAGE_URL = "X-Storage-Url";
 
    @Override
    protected void configure() {
       super.configure();
       bindHttpApi(binder(), AuthenticationApi.class);
       bindHttpApi(binder(), TempAuthApi.class);
+      requestStaticInjection(SwiftAuthenticationModule.class);
```
nacx: Once the injected variables are not here, this should not be needed.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/config/SwiftAuthenticationModule.java
```diff
@@ -62,22 +64,40 @@
...
                          .put("tempAuthCredentials", i.getInstance(TempAuth.class)).build();
    }
+   
+   public static String getIdentityHeaderName(){
+       return identityHeaderNameUser;
+   }
+
+   public static String getIdentityHeaderPass(){
+       return identityHeaderNamePass;
+   }
```
nacx: These getters should no longer be needed.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/config/SwiftAuthenticationModule.java
```diff
@@ -153,7 +173,7 @@ private URI getURI(String headerValue) {
       public AdaptTempAuthResponseToAccess setContext(HttpRequest request) {
          String host = request.getEndpoint().getHost();
          this.host = host;
-         this.username = request.getFirstHeaderOrNull(STORAGE_USER);
+	 this.username = request.getFirstHeaderOrNull(SwiftAuthenticationModule.getIdentityHeaderName());
```
nacx: You can inject the header variable in this class too, to avoid dealing with static methods.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/TempAuthMockTest.java
```diff
@@ -45,13 +46,24 @@
...
    public void testTempAuthRequest() throws Exception {
-      tempAuthServer.enqueue(new MockResponse().setResponseCode(204)
+     Properties overrides = null;
+     // with default values
+     test(overrides);
+     overrides = new Properties();
+     overrides.setProperty(SwiftAuthenticationModule.TEMP_AUTH_HEADER_USER , "X-Auth-User");
+     overrides.setProperty(SwiftAuthenticationModule.TEMP_AUTH_HEADER_PASS , "X-Auth-Pass");
+     // with specific Header Name values
+     test(overrides);
```
nacx: Better move this to another test named `testTempAuthRequestWithCustomHeaders` or similar, to properly qualify and isolate the test that verifies this change.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/TempAuthMockTest.java
```diff
@@ -71,8 +83,10 @@ public void testTempAuthRequest() throws Exception {
       assertEquals(listContainers.getHeader("X-Auth-Token"), "token");
    }
 
-   private SwiftApi api(String authUrl) throws IOException {
-      Properties overrides = new Properties();
+   private SwiftApi api(String authUrl, Properties overrides) throws IOException {
+      if (overrides == null){
+         overrides = new Properties();
+      }
```
nacx: Instead of passing null, remove this check and call this method with a new Properties object.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/TempAuthMockTest.java
```diff
@@ -60,8 +72,8 @@ public void testTempAuthRequest() throws Exception {
 
       RecordedRequest auth = tempAuthServer.takeRequest();
       assertEquals(auth.getMethod(), "GET");
-      assertEquals(auth.getHeader("X-Storage-User"), "user");
-      assertEquals(auth.getHeader("X-Storage-Pass"), "password");
+      assertEquals(auth.getHeader(SwiftAuthenticationModule.getIdentityHeaderName()), "user");
+      assertEquals(auth.getHeader(SwiftAuthenticationModule.getIdentityHeaderPass()), "password");
```
nacx: Once there are two tests, it's better to keep the expected header name hardcoded, to properly verify that the expected configuration is applied.

nacx:
Also, you can easily avoid the need for optional injection (so you can declare the variables final and use constructor injection) by setting the default values for the headers in the [SwiftApiMetadata default properties](https://github.com/jclouds/jclouds/blob/master/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/SwiftApiMetadata.java#L59-L66). That would be also a cleaner approach.

Etienne-Carriere:
Compared to your proposal, I added a constant class in reference to regroup the constants.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/binders/TempAuthBinder.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ */
+public final class TempAuthBinder implements Binder{
+   @Inject
+   @Named(TEMP_AUTH_HEADER_USER)
+   private String identityHeaderNameUser;
+
+   @Inject
+   @Named(TEMP_AUTH_HEADER_PASS)
+   private String identityHeaderNamePass;
```
nacx: Declare the variables final and use [constructor injection](https://github.com/google/guice/wiki/Injections#constructor-injection).

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/config/SwiftAuthenticationModule.java
```diff
@@ -98,12 +97,16 @@ protected void configure() {
...
-      Access auth(@HeaderParam(STORAGE_USER) String user, @HeaderParam(STORAGE_PASS) String key);
+      Access auth(@BinderParam(TempAuthBinder.class) Credentials credentials);
    }
 
    static final class AdaptTempAuthResponseToAccess
          implements Function<HttpResponse, Access>, InvocationContext<AdaptTempAuthResponseToAccess> {
 
+      @Inject
+      @Named(TEMP_AUTH_HEADER_USER)
+      private String identityHeaderNameUser;
```
nacx: Add it to the constructor and declare it final to use constructor injection too.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/config/SwiftAuthenticationModule.java
```diff
@@ -153,7 +156,7 @@ private URI getURI(String headerValue) {
       public AdaptTempAuthResponseToAccess setContext(HttpRequest request) {
          String host = request.getEndpoint().getHost();
          this.host = host;
-         this.username = request.getFirstHeaderOrNull(STORAGE_USER);
+	 this.username = request.getFirstHeaderOrNull(identityHeaderNameUser);
```
nacx: [minor] Fix indentation

nacx:
Squashed the commits and amended the message to include the issue id.
Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/ac2f746e) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/01d74969). Thanks @Etienne-Carriere!

PR #1047, opened 2017-01-03
Allow ipv6 address blocks

zack-shoylev:
https://issues.apache.org/jira/browse/JCLOUDS-1100

### DIFF
--- a/core/src/main/java/org/jclouds/util/Strings2.java
+++ b/core/src/main/java/org/jclouds/util/Strings2.java
@@ -41,6 +41,7 @@ import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.collect.Multimap;
 import com.google.common.io.CharStreams;
+import com.google.common.net.InetAddresses;
 import com.google.common.primitives.Chars;
 
 public class Strings2 {
@@ -67,7 +68,7 @@ public class Strings2 {
          throw new IllegalStateException("error creating pattern: " + in, e);
       }
    }
-   
+
    private static final LoadingCache<Character, String> CHAR_TO_ENCODED = CacheBuilder.newBuilder()
          .<Character, String> build(new CacheLoader<Character, String>() {
             @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jclouds/util/Strings2.java
```diff
@@ -79,19 +79,36 @@ public String load(Character plain) throws ExecutionException {
...
+         + "((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]d|1dd|[1-9]?d)(.(25[0-5]|2[0-4]d|1dd|[1-9]?d)){3}))|:))|"
+         + "(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|"
+         + "((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]d|1dd|[1-9]?d)(.(25[0-5]|2[0-4]d|1dd|[1-9]?d)){3}))|:))|"
+         + "(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|"
+         + "((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]d|1dd|[1-9]?d)(.(25[0-5]|2[0-4]d|1dd|[1-9]?d)){3}))|:))|"
+         + "(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|"
+         + "((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]d|1dd|[1-9]?d)(.(25[0-5]|2[0-4]d|1dd|[1-9]?d)){3}))|:))|"
+         + "(:(((:[0-9A-Fa-f]{1,4}){1,7})|"
+         + "((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]d|1dd|[1-9]?d)"
+         + "(.(25[0-5]|2[0-4]d|1dd|[1-9]?d)){3}))|:)))(%.+)?s*(\\/([0-9]|[1-9][0-9]|1[0-1][0-9]|12[0-8]))?$");
```
nacx: Can we better use Guava's [InetAddresses#isInetAddress](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/net/InetAddresses.html#isInetAddress(java.lang.String)) or the `forString` methods? This looks too error-prone and difficult to read/fix/maintain. Perhaps we could manually split the mask, and validate the address part and then the mask?

core/src/test/java/org/jclouds/util/Strings2Test.java
```diff
@@ -42,9 +42,10 @@ public void testUrlEncodeDecodeShouldGiveTheSameString() {
    public void testIsCidrFormat() {
       assert Strings2.isCidrFormat("1.2.3.4/5");
       assert Strings2.isCidrFormat("0.0.0.0/0");
+      assert Strings2.isCidrFormat("fdab:43a4:b3a2:f140::/64");
```
nacx: We should add more tests to verify the IPv6 format

zack-shoylev:
@nacx Is this what you had in mind? I am asking because the logic needed to avoid the regex was a little bit more complex than just splitting.

nacx:
Yes, I see, but it is still more readable/maintainable than the regexp I think.
Thanks for the change!

zack-shoylev:
merged and backported to 2.0.x

PR #1049, opened 2017-01-05
WindowsEveryonePrincipal : update regex for FR

mildis:
"Everyone" in a french Windows is "Tout le monde".
Thus, the line.split(" ") returns only "Tout" and putBlob() throws an exception.
The modified regex searches for the first two consecutive white chars to return the name.
Windows's whoami returns a bunch of 0x20 between the name and the type.

It's an update to JCLOUDS-1015 and can be released with the next 2.0 fix.


zack-shoylev:
Ah, very good find.

+1

mildis:
@zack-shoylev Hope it could be included in the next 2.0.x fix release.

mildis:
Created https://issues.apache.org/jira/browse/JCLOUDS-1222 to match this PR.

zack-shoylev:
merged and backported to 2.0.x

PR #1052, opened 2017-01-12
GetBucketLocation should use path-style requests.

timuralp:
When making a `GetBucketLocation` request, Amazon may route the request
to the bucket region. When making it with v4 signer, the request may
fail because of the region mismatch. Concretely, a request to
test.s3.amazonaws.com may resolve to s3-us-west-2-w.amazonaws.com. The
request itself is prepared for the us-east-1 region (s3.amazonaws.com
endpoint), but then fails when the DNS resolution points to a
us-west-2 endpoint.

Bucket-in-path works around this for the `GetBucketLocation` requests.
That means that every `GetBucketLocation` request will be of the form:
https://s3.amazonaws.com/{bucket}?location. This ensures that jclouds
requests will not be subjected to Amazon's routing/DNS pointers.

Fixes: JCLOUDS-1213

### DIFF
--- a/apis/s3/src/main/java/org/jclouds/s3/S3Client.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/S3Client.java
@@ -533,10 +533,10 @@ public interface S3Client extends Closeable {
    @Named("GetBucketLocation")
    @GET
    @QueryParams(keys = "location")
-   @Path("/")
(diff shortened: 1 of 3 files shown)


andrewgaul:
I could not verify the original symptoms but the fix makes sense and passes integration tests against AWS S3 and DreamObjects.

PR #1055, opened 2017-01-25
Correct regex in ExpandProperties

andrewgaul:
Previously Android choked on this incomplete regex.


andrewgaul:
@nacx Please review.

nacx:
Uops! Dis you get errors because of this?

andrewgaul:
Yeah jclouds does not work with Android for a few different reasons.  I gave up when Guice threw some codegen errors.

PR #1058, opened 2017-02-04
update import packages for netty

andreaturli:


drivers/netty/pom.xml
```diff
@@ -35,7 +35,7 @@
     <jclouds.osgi.export>org.jclouds.netty*;version="${project.version}"</jclouds.osgi.export>
     <jclouds.osgi.import>
       org.jclouds*;version="${project.version}",
-      org.jboss.netty*;version="[3.3,4)",
+      io.netty*;version="[3.3,4)",
```
nacx: Why this change? AFAIK with the [version of netty we use](https://github.com/jclouds/jclouds/blob/master/drivers/netty/pom.xml#L57-L61) the packages we need are already [org.jboss.netty*](https://github.com/jclouds/jclouds/blob/master/drivers/netty/src/main/java/org/jclouds/netty/io/ChunkedFileInputStream.java#L24-L25).

andreaturli:
I'm creating a `feature`for netty driver in jclouds-karaf 
(`<bundle dependency='true'>mvn:io.netty/netty/3.5.9.Final</bundle>`) and it doesn't seem to export `org.jboss.netty*` - changing the import solves the problem

nacx:
But that dependency is [exporting the org.jboss.netty](https://github.com/netty/netty/blob/netty-3.5.9.Final/pom.xml#L414) package, so there's something I don't get here. Where is this new package coming from? It doesn't look like the jclouds driver is using it.

andreaturli:
I've re-tested with the current version and it seems to work. Closing the PR, though. Sorry for the noise, @nacx especially 

OSGi is interesting ...

nacx:
Hmmm instead of closing, let's re-open it fixing the version? It looks like the current range (until version 4) could bring versions of the dependency incompatible with the current import.

andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/4ae2d208) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/663123a9)

thanks @nacx for reviewing it!

nacx:
I think this [breaks jclouds-karaf](https://jclouds.ci.cloudbees.com/job/jclouds-karaf/2083/):

```bash
Caused by: org.osgi.framework.BundleException:
    Unresolved constraint in bundle jclouds-netty [95]:
        Unable to resolve 95.0: missing requirement [95.0] osgi.wiring.package; (&(osgi.wiring.package=org.jboss.netty.buffer)(version>=3.5.9.Final))
```

Do we need to update [the netty bundle version](https://github.com/jclouds/jclouds-karaf/blob/master/pom.xml#L214) too?

PR #1059, opened 2017-02-06
Fix O(n^2) response time for "list-security-groups" on openstack-nova.

geomacy:
Proposed fix for https://issues.apache.org/jira/browse/JCLOUDS-1235.

This change takes the approach of storing the information about the
overall list of groups within the `SecurityGroupInRegion` when it is
created, so that any subsequent conversion operation has access to all
the groups in the same region as the one to be converted.

It also collapses the functionality of `NovaSecurityGroupToSecurityGroup`,
`SecurityGroupRuleToIpPermission` and `FindSecurityGroupWithNameAndReturnTrue`
all into `NovaSecurityGroupInRegionToSecurityGroup`. These three classes could arguably
be removed or at least deprecated; I haven’t done either in this PR yet and would ask you for guidance on the best approach here.

Another point for review, note the comment in NovaSecurityGroupInRegionToSecurityGroup.java:110 about the workaround to the issue https://issues.apache.org/jira/browse/JCLOUDS-1234. Is this the best approach in this circumstance?

P.S. Thanks to @andreaturli for guidance on this.


apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
```diff
@@ -135,8 +136,9 @@ public NovaSecurityGroupExtension(NovaApi api,
       }
 
       Set<String> groupNames = instance.getSecurityGroupNames();
-      Set<? extends SecurityGroupInRegion> rawGroups =
-              sgApi.get().list().filter(nameIn(groupNames)).transform(groupToGroupInRegion(region)).toSet();
+       final FluentIterable<org.jclouds.openstack.nova.v2_0.domain.SecurityGroup> allGroups = sgApi.get().list();
+       Set<? extends SecurityGroupInRegion> rawGroups =
+              allGroups.filter(nameIn(groupNames)).transform(groupToGroupInRegion(allGroups, region)).toSet();
```
andreaturli: [minor] do you need allGroups variable or the fluentIterable can transform and filter and finally transform again?

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
```diff
@@ -153,7 +155,8 @@ public SecurityGroup getSecurityGroupById(String id) {
          return null;
       }
 
-      SecurityGroupInRegion rawGroup = new SecurityGroupInRegion(sgApi.get().get(groupId), region);
+       final FluentIterable<org.jclouds.openstack.nova.v2_0.domain.SecurityGroup> allGroups = sgApi.get().list();
```
andreaturli: Remember jclouds uses 3-spaces indentation

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/NovaSecurityGroupInRegionToSecurityGroup.java
```diff
@@ -30,43 +33,103 @@
...
+                        "group": {
+                           "tenant_id": "a0ade3ca76784719845363979dc1014e",
+                           "name": "jclouds-qa-scheduler-docker-entity"
+                        },
+              Rather than pick one group at random and risk using the wrong group, here we fall back to the
+              least-worst option(?) and refuse to add any rule.
 
-      return builder.build();
-   }
+              See https://issues.apache.org/jira/browse/JCLOUDS-1234.
```
andreaturli: Great commenti!

apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtensionLiveTest.java
```diff
@@ -29,4 +39,18 @@ public NovaSecurityGroupExtensionLiveTest() {
...
+        skipIfSecurityGroupsNotSupported();
+
+        ComputeService computeService = view.getComputeService();
+        Optional<SecurityGroupExtension> securityGroupExtension = computeService.getSecurityGroupExtension();
+        assertTrue(securityGroupExtension.isPresent(), "security extension was not present");
+
+        Set<SecurityGroup> groups = securityGroupExtension.get().listSecurityGroups();
+        System.out.println(groups.size());
+        for (SecurityGroup group : groups) {
+            System.out.println(group);
```
andreaturli: Use assertions instead of sout, please

andreaturli:
Exemplar PR description and jira tracking, btw -- impressive work @geomacy

nacx:
Thanks @geomacy! The details and PR are much appreciated!

I haven't read in detail the implementation, but wanted to early share a design idea. Instead of propagating the list of current security groups to each piece of code that might need it, what about:

* Create a `Supplier<List<SecurityGroup>>` that provides the list of security groups, and make that injectable.
* Also configure a `@Memoized Supplier<List<SecurityGroup>>` binding, that just decorates the previous one and caches it for the duration of the [session](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/Constants.java#L93-L98). This is a common way of configuring and exposing caches and suppliers in jclouds.

This way we could have a lazy and cached list of security groups, and classes can choose to get injected the "live" supplier or the memoized one. The memoized version could also expose a couple accessors to add/remove security groups, so they can be called from the security group extension or compute service adapter.

Thoughts?

geomacy:
hi @nacx, thanks for the suggestions.  A couple of thoughts/questions:

- Would the `Supplier<List<SecurityGroup>>` be a singleton similar to [FindSecurityGroupWithNameAndReturnTrue](https://github.com/jclouds/jclouds/blob/rel/jclouds-2.0.0/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/predicates/FindSecurityGroupWithNameAndReturnTrue.java#L52-L55) - I assume it would have to inject the `NovaApi` in the same way?
- I'm not familiar with the memoizing as used in jclouds, but I did think about some sort of memo/cache for this; however, I wondered if that would be right for this scenario.  does it cope with the need to invalidate the memo when mutating operations are performed? e.g. if I add a group A, then another, B, then add an ingress rule to B permitting access from A, each of these three operations modifies the set of security groups that is defined, so each will require a fresh fetch of the list of groups.  If the list is cached per ‘session’ (not sure what that is) then the sequence above won’t work.  
- I guess we could add accessors to support keeping the memo up-to-date on the fly as operations were carried out, but (a) I have bad experiences in the past with trying to keep caches in-sync with their back-ends, and (b) in any case I don't think that applies here, as in my example, the back end cloud will generate the ID for group A when it gets created, so the only way to get its value into the memo is to do the fetch again, so the memo isn't saving you an operation.  For this reason I thought it might be safest to stick with doing a single `list` operation prior to each of the actions above.  

I shall try to read up in the code on how the memoizers are used, and maybe we can work out together what to do.

andreaturli:
Fair points @geomacy I'm in favor of keeping the implementation as-is which already optimize a lot!

if further optimizations will be required we could always revisit it again using @nacx suggestions as starting point. Wdyt?

nacx:
+1 to keep the current approach. Thanks @geomacy! I'll have a detailed look at the PR in a while :)

geomacy:
Thanks for the comments @andreaturli and @nacx, - will add a couple of responses above, and update the PR with corresponding changes.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
geomacy: We do need the variable, as it is used both as the root of the iteration and as a parameter to it (so that all groups have knowledge of all other groups).

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
geomacy: Dang, thought I had got all of those! :-)

geomacy:
hi @andreaturli , @nacx , I have updates the PR with changes as suggested in the comments above, improving the additional test and now deleting unused classes.

andreaturli:
@geomacy can you please squash (and rebase if needed) the 3 commits into one?

Also do you need this commit ported to `2.0.x` as well?

geomacy:
hi @andreaturli, changes squashed.  Yes, it would be good to have this ported to 2.0.x as well if that is possible.

andreaturli:
thanks @geomacy, merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/717b75a3) and back-ported to [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/c4245acd)

PR #1063, opened 2017-02-17
Return lastModified and size when listing S3 parts

andrewgaul:


andrewgaul:
Unfortunately this introduces a new method since we cannot change the `listMultipartParts` return type.

apis/s3/src/main/java/org/jclouds/s3/domain/ListMultipartUploadResponse.java
```diff
@@ -0,0 +1,42 @@
...
+   public abstract int partNumber();
+   public abstract Date lastModified();
+   public abstract String eTag();
+   public abstract long size();
+
+   public static ListMultipartUploadResponse create(int partNumber, Date lastModified, String eTag, long size) {
+      checkArgument(partNumber > 0, "partNumber must be greater than zero, was: " + partNumber);
+      checkNotNull(eTag, "eTag");
+      lastModified = (Date) checkNotNull(lastModified, "lastModified").clone();
+      checkArgument(size >= 0, "size must be positive, was: " + size);
```
nacx: [minor] Use `checkArgument(size >= 0, "size must be positive, was: %s", size);`, to avoid the implicit stringbuilder if the precondition is met.

apis/s3/src/main/java/org/jclouds/s3/xml/PartIdsFromHttpResponseFull.java
```diff
@@ -0,0 +1,79 @@
...
+
+   private int partNumber;
+   private Date lastModfied;
+   private String eTag;
+   private long size;
+
+   private final ImmutableMap.Builder<Integer, ListMultipartUploadResponse> parts = ImmutableMap.builder();
+
+   @Inject
+   public PartIdsFromHttpResponseFull(DateService dateParser) {
```
nacx: [minor] According to Guice best practices, it's better to reduce the visibility of the injection constructors. Let's leave it without the `public` modifier, in default scope.

andrewgaul: Done.

apis/s3/src/main/java/org/jclouds/s3/domain/ListMultipartUploadResponse.java
andrewgaul: Done twice.  Some versions of error-prone warn about this but the check was reparented into the Google-only Klippy.

andrewgaul:
Pushed to master as ec932321bde5662d8ecbef7cf5428a6017b07dd0 and 2.0.x as d2041667d770bcb865a7bfb451b5910dd9084d91.

PR #1073, opened 2017-03-06
Take into account the template properties each time the builder is retrieved

nacx:
When getting the template for the node, always check if there is a custom property for it.


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a7ccf657) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/c6d8f804).

PR #1076, opened 2017-03-14
Fix listing nodes on gce when there is an instance created from a snapshot

tommywo:
Rebased from https://github.com/jclouds/jclouds/pull/1075


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/fdaa3fd7) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/f5a5d869). Thanks @tommywo!

PR #1077, opened 2017-03-17
Fix error creating list of volume bindings for Docker

grkvlt:
Currently the code only sets the last in the list of the passed in volume bindings, by continually overwriting the bindings with a single element list. This fixes that and sets all of them in a single list.

PR #1078, opened 2017-03-17
Fix error creating list of volume bindings for Docker

grkvlt:
Copy of #1077 for version 2.0.x


nacx:
Pushed to 2.0.x as [07935978](http://git-wip-us.apache.org/repos/asf/jclouds/commit/07935978). Thanks @grkvlt!
BTW, there's no need to open a PR to each branch (unless there are conflicts). Just say if you want to have it backported and we'll do! :)

PR #1086, opened 2017-04-12
[JCLOUDS-1128] Added page and pagesize fields to ListProjectOptions.

mcdonnell-john:
Cloudstack's listProjects request supports additional fields for page
and pagesize.  These were missing but have been added in as part of this
commit.

As per the Cloudstack documentation https://cloudstack.apache.org/api/apidocs-4.8/domain_admin/listProjects.html this call can be pagenated, and so should contain the page and pagesize fields.


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/24f961ea) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/499e61b8). Thanks!

PR #1087, opened 2017-04-20
Fix the Jetty start statement used in compute live tests

nacx:
During tests, the Jetty server was started manually running the jar file, and the PID file was not generated. This caused the stop statement to silently fail and Jetty kept running. This change also removes a graceful wait used in live tests that does not make sense in stubbed integration tests.


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/bf3a78d0) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/8c2e7435).

PR #1088, opened 2017-04-21
Fix BaseComputeServiceLiveTest configuration

nacx:
This commit fixes the configuration of the base compute service test class. That class initialized entities taken from the injector in several methods: setupContext and initializeContext. The former was called before all tests, and the latter was called several times during tests, to check the behaviour of jclouds when using multiple contexts.

The issue was that the first method initialized the OpenSocketFinder from the first context, and that context is closed everytime the initializeContext method is invoked, leaving the OpenSocketFinder instance unusable, since closing its context closes the executor services associated to it.

This moves the initialization of all entities that depend on the injector to the initializeContext method, so they are always properly initialized, and also aligns the annotation of the setupContext method with the one on the base class.


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/13cf2cba) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/6100b7c3).

PR #1089, opened 2017-04-21
Fix default executor rejection policies

nacx:
The jclouds `DynamicThreadPoolExecutor` implements a default rejection policy to enqueue the tasks even if the queue is full; in that case, we just wait for space and add the task, instead of aborting the execution.

The current implementation does not take into account that the `reject` method from a `ThreadPool` can be invoked not only when the queue is full, but when the executor is already shutdown. In that case, we shouldn't enqueue the task but propagate an exception. Otherwise, the task remains in the queue with no chance of being executed, and any `Future` that is returned for that execution will block forever when calling its `get()` method.

Regardless of the intention, convenience and the real need for this "wait-if-full" rejection policy implementation (we could discus it in the dev mailing list if it is really needed or we could remove it, but this looks like a huge impact change), this change fixes an existing race condition in the rejection policy.

/cc @andrewgaul @demobox @andreaturli


andreaturli:
I've tested jclouds-labs#382 with this patch and it helps!

andrewgaul:
👍 

Outside the scope of this PR, we hoped to remove executors in [JCLOUDS-153](https://issues.apache.org/jira/browse/JCLOUDS-153).  Is this still a goal?

nacx:
I don't know if that should be still a goal. I mean, that was brought onto the table when un-asynching the APIs to simplify jclouds. Now the use of executors is very isolated (concurrent socket finder, and a few more utilities), and I wouldn't say they introduce much overhead or complexity to jclouds. They are actually pretty useful in some operations and I think it makes sense to keep them (although I'd love to see other's opinion on this).

What we have done in several places where the executors were used to perform concurrent operations (such as the ChefService), is to refactor the interfaces and add an additional ExecutorService parameter so users can pass their own. That approach is incomplete IMO, since we already have a way to let users inject their own user executor in jclouds: just passing an ExecutorServiceModule [with their executor service](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/concurrent/config/ExecutorServiceModule.java#L101-L107) when creating the context. Having this, anyone can use jclouds with their application-managed user executor, but jclouds can provide a default one. I think this makes sense and I don't see a *real* need to remove them. WDYT?

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/e5063aea) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/05f64b20).

PR #1090, opened 2017-04-28
JClouds 1281- DLO  Swift Implementation

archupsg03:
Have implemented Swift Dynamic Large Object upload strategy. Requires changes in 
1. Manifest upload with X-Object-Manifest header
2. Remove blobs, in order to remove the manifest and segments


apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -188,11 +197,17 @@ public boolean createContainerInLocation(Location location, String container) {
...
-         return api.getContainerApi(regionId).create(container, ANYBODY_READ);
+         resultContainerCreate = api.getContainerApi(regionId).create(container, ANYBODY_READ);
       }
-      return api.getContainerApi(regionId).create(container, BASIC_CONTAINER);
+      resultContainerCreate = api.getContainerApi(regionId).create(container, BASIC_CONTAINER);
+      if (resultContainerCreate) {
+         Container val = api.getContainerApi(regionId).get(container);
+         containerCache.put(container, Optional.fromNullable(val));
+      }
+      return resultContainerCreate;
```
andrewgaul: How is creating a container related to dynamic large objects?  If this is a separate change, submit a separate pull request.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/SwiftApi.java
```diff
@@ -17,6 +17,7 @@
 package org.jclouds.openstack.swift.v1;
 
 import java.io.Closeable;
+
```
andrewgaul: Please sweep through the entire pull request and remove these whitespace changes.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -387,13 +402,26 @@ public Blob getBlob(String container, String name, GetOptions options) {
...
-      // Multipart objects have a manifest which points to subobjects.  Normally
+      // Multipart objects have a manifest which points to subobjects. Normally
       // deleting a object only deletes the manifest, leaving the subobjects.
       // We first try a multipart delete and if that fails since the object is
       // not an MPU we fall back to single-part delete.
-      DeleteStaticLargeObjectResponse response = api.getStaticLargeObjectApi(regionId, container).delete(name);
-      if (!response.status().equals("200 OK")) {
+
+      String objManifest = getManifestInfo(container, name);
+      String sloInfo = getSLOInfo(container, name);
```
andrewgaul: Previously this code issued one or two DELETEs depending if the name was an SLO or a regular object.  With this commit we issue always issue three RPCs, getWithoutBody, getWithoutBody, and delete with some flags.  How can you optimize this?  Also this fails if the object does not exist.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -519,10 +557,19 @@ public String completeMultipartUpload(MultipartUpload mpu, List<MultipartPart> p
          String path = mpu.containerName() + "/" + getMPUPartName(mpu, part.partNumber());
          builder.add(Segment.builder().path(path).etag(part.partETag()).sizeBytes(part.partSize()).build());
       }
-
       return api.getStaticLargeObjectApi(regionId, mpu.containerName()).replaceManifest(mpu.blobName(),
             builder.build(), mpu.blobMetadata().getUserMetadata(), getContentMetadataForManifest(mpu.blobMetadata().getContentMetadata()));
    }
+   
+   public String completeMultipartUploadDLO(MultipartUpload mpu, List<MultipartPart> parts) {
```
andrewgaul: Remove - we should not expose dynamic large objects in the portable abstraction!  Users can call the provider-level APIs for this functionality.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -908,4 +957,115 @@ public void run() {
...
+    * @param container
+    * @param objectName
+    * @return true if object is SLO uploaded else null
+    */
+   private String getSLOInfo(String container, String objectName) {
+      ObjectApi objectApi = api.getObjectApi(regionId, container);
+      Multimap<String, String> objInfo = objectApi.getWithoutBody(objectName).getHeaders();
+      if (objInfo.containsKey("X-Static-Large-Object")) {
+         Collection<String> objectManifest = objInfo.get("X-Static-Large-Object");
+         return String.valueOf(objectManifest.toArray()[0]);
```
andrewgaul: Consider the simpler:

```java
ObjectApi objectApi = api.getObjectApi(regionId, container);
Multimap<String, String> objInfo = objectApi.getWithoutBody(objectName).getHeaders();
Collection<String> objectManifest = objInfo.get("X-Static-Large-Object");
return Iterables.getFirst(objectManifest, null);
```

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -908,4 +957,115 @@ public void run() {
...
+      }
+      return null;
+   }
+   
+   
+   /**
+    * @param containerAndPrefix
+    */
+   private void removeObjectsWithPrefix(String containerAndPrefix) {
```
andrewgaul: Add `static`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -908,4 +957,115 @@ public void run() {
...
+      return parts;
+   }
+    
+   /**
+    * Retrieves a list of existing storage
+    * @param container
+    * @param options
+    * @return maximum of 10,000 object names per request.
+    */
+   private PageSet<? extends StorageMetadata> list(final String container,
```
andrewgaul: Why do you reimplement `ObjectApi.list`?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/options/OpenStackSwiftPutOptions.java
```diff
@@ -0,0 +1,109 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.openstack.swift.v1.options;
+
+import com.google.common.util.concurrent.ListeningExecutorService;
+import org.jclouds.blobstore.options.PutOptions;
+
+public class OpenStackSwiftPutOptions extends PutOptions {
```
andrewgaul: Remove this entire thing and make users call either the dynamic or static APIs.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamiclargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,143 @@
...
+   public void setup() {
+      super.setup();
+      for (String regionId : regions) {
+         boolean created = getApi().getContainerApi(regionId).create(defaultContainerName);
+         if (!created) {
+            deleteAllObjectsInContainer(regionId, defaultContainerName);
+         }
+      }
+
+      megOf1s = new byte[1024 * 1024];
```
andrewgaul: Call `TestUtils.randomByteSource()` instead.

blobstore/src/main/java/org/jclouds/blobstore/options/PutOptions.java
```diff
@@ -39,11 +39,11 @@
...
 
    private BlobAccess blobAccess = BlobAccess.PRIVATE;
-   private boolean multipart = false;
-   private boolean useCustomExecutor = false;
+   protected boolean multipart = false;
+   protected boolean useCustomExecutor = false;
 
    // TODO: This exposes ListeningExecutorService to the user, instead of a regular ExecutorService
-   private ListeningExecutorService customExecutor = MoreExecutors.sameThreadExecutor();
+   protected ListeningExecutorService customExecutor = MoreExecutors.sameThreadExecutor();
```
andrewgaul: Revert this after removing `OpenStackSwiftPutOptions`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -230,10 +245,10 @@ public void setContainerAccess(String name, ContainerAccess access) {
       ObjectApi objectApi = api.getObjectApi(regionId, container);
       ObjectList objects = objectApi.list(toListContainerOptions.apply(options));
       if (objects == null) {
-         containerCache.put(container, Optional.<Container> absent());
+         //containerCache.put(container, Optional.<Container> absent());
          return new PageSetImpl<StorageMetadata>(ImmutableList.<StorageMetadata> of(), null);
       } else {
-         containerCache.put(container, Optional.of(objects.getContainer()));
+         //containerCache.put(container, Optional.of(objects.getContainer()));
```
andrewgaul: Why comment this out?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,83 @@
...
+    *           name/key.
+    *
+    * @return {@link SwiftObject#getEtag()} of the object, which is the MD5
+    *         checksum of the concatenated ETag values of the {@code segments}.
+    */
+   @Named("dynamicLargeObject:replaceManifest")
+   @PUT
+   @ResponseParser(ETagHeader.class)
+   @Headers(keys = "X-Object-Manifest", values = "{containerName}/{objectName}/")
+   String replaceManifest(@PathParam("containerName") String containerName, @PathParam("objectName") String objectName,
```
andrewgaul: Annotate this method @Deprecated and document that users should prefer SLO.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStoreParallelLiveTest.java
```diff
@@ -123,6 +124,50 @@ public void uploadMultipartBlob() {
...
+            .payload(new FilePayload(bigFile))
+            .build();
+      // configure the blobstore to use multipart uploading of the file
+      OpenStackSwiftPutOptions opt = new OpenStackSwiftPutOptions();
+      opt.multipart(executor);
+      opt.DLO(true);
+      String eTag = blobStore.putBlob(CONTAINER, blob, opt);
+      // assertEquals(eTag, etag);
+      // The etag returned by Swift is not the md5 of the Blob uploaded
+      // It is the md5 of the concatenated segment md5s
```
andrewgaul: Why not verify the ETag?

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamiclargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+      String etag1s = objectApi.put(name + "/1", newByteSourcePayload(ByteSource.wrap(megOf1s)));
+      awaitConsistency();
+      assertMegabyteAndETagMatches(regionId, containerName, name + "/1", etag1s);
+
+      String etag2s = objectApi.put(name + "/2", newByteSourcePayload(ByteSource.wrap(megOf2s)));
+      awaitConsistency();
+      assertMegabyteAndETagMatches(regionId, containerName, name + "/2", etag2s);
+
+      List<Segment> segments = ImmutableList.<Segment> builder()
```
andrewgaul: Prefer `ImmutableList.of`.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamiclargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,143 @@
...
+      ObjectList objects = getApi().getObjectApi(regionId, containerName).list(new ListContainerOptions());
+      if (objects == null) {
+         return;
+      }
+      List<String> pathsToDelete = Lists.transform(objects, new Function<SwiftObject, String>() {
+         public String apply(SwiftObject input) {
+            return containerName + "/" + input.getName();
+         }
+      });
+      if (!pathsToDelete.isEmpty()) {
```
andrewgaul: Unneeded conditional.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
andrewgaul: Remove this function and make the single caller explicitly call `splitContainerAndKey` and `removeObjectsWithPrefix(String, String)`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -908,4 +957,115 @@ public void run() {
...
+      do {
+         org.jclouds.openstack.swift.v1.options.ListContainerOptions listContainerOptions = prefix(prefix);
+         if (nextMarker != null) {
+            listContainerOptions = listContainerOptions.marker(nextMarker);
+         }
+         PageSet<? extends StorageMetadata> chunks = list(container, listContainerOptions);
+         Iterator<? extends StorageMetadata> iter = chunks.iterator();
+
+         while (iter.hasNext()) {
+            objectApi.delete(iter.next().getName());
```
andrewgaul: Call `BulkApi.buckDelete`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
```diff
@@ -908,4 +957,115 @@ public void run() {
...
+
+      removeObjectsWithPrefix(container, prefix);
+   }
+
+    
+   /**
+    * @param containerAndKey
+    * @return array of size 2 with container and objectPrefix
+    */
+   private String[] splitContainerAndKey(String containerAndKey) {
```
andrewgaul: Add `static`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
archupsg03: Deleting DLO objects requires, the details of the segments.

BlobStore's List command takes BlobStore's ListContainerOptions as input.
But incase of remove blob we need Swift's ListContainerOptions. That the reasonf i copied and made few changes

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStoreParallelLiveTest.java
archupsg03: Etag comparison was already commented in the previous function

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamiclargeObjectApiLiveTest.java
archupsg03: These are just copied from SLO testcase, and appropriate changes has been made for DLO. These things remain the same.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
archupsg03: This is a separate fix. Will submit another PR

archupsg03:
For DLO upload, etag will the md5 of the manifest and not the content of the file thats getting uploaded.
Thats the reason etag mismatch happens in ParallelLiveTest

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/options/OpenStackSwiftPutOptions.java
archupsg03: Am not getting this point :(

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
archupsg03: Not getting this point too

andrewgaul: jclouds consists of two levels, the lower-level provider APIs, e.g., `DynamicLargeObjectApi`, and the higher-level portable abstraction, e.g., `RegionScopedSwiftBlobStore`.  The former exposes all the functionality of a given provider, e.g, SLO, DLO, etc., while the latter only implements a portable layer which is common between providers.  Thus `completeMultipartUploadDLO` should not exist and if callers want this functionality they can use `RegionScopedSwiftBlobStore`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
andrewgaul: `RegionScopedSwiftBlobStore` can call `ObjectApi.list`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/options/OpenStackSwiftPutOptions.java
andrewgaul: Once you address https://github.com/jclouds/jclouds/pull/1090#discussion_r114451055 this code becomes unnecessary.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamiclargeObjectApiLiveTest.java
andrewgaul: If the inconsistency bothers you, please submit a fix for the existing SLO tests as well.

andrewgaul:
> For DLO upload, etag will the md5 of the manifest and not the md5 of the concatenated etags.
> Thats the reason etag mismatch happens in ParallelLiveTest. So etag checking for DLO objects requires
> manifest to be constructed locally and compute its md5 and then compare with the etag from Swift

We cannot introduce failing tests so how do you plan to address this?

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamiclargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+import com.google.common.base.Function;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
+import com.google.common.io.ByteSource;
+import com.google.common.util.concurrent.Uninterruptibles;
+
+@Test(groups = "live", testName = "DynamicLargeObjectApiLiveTest", singleThreaded = true)
+public class DynamiclargeObjectApiLiveTest extends BaseSwiftApiLiveTest {
```
andrewgaul: Rename class to `DynamicLargeObjectApiLiveTest` with a capital "L".

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
archupsg03: Theres problem in using objectApi.List. I am not able to get the markers. 

How does objecApi.List handles once it lists the first 10k objects.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/options/OpenStackSwiftPutOptions.java
archupsg03: Say if the user prefers to call putMultipart Blob with DLO, doesn't want to use DLO API. How can such cases be handled ?

andrewgaul: The portable abstraction does not expose all functionality.  The caller must use the DLO API to access DLO.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
andrewgaul: `ObjectApi.list(ListContainerOptions)` Javadoc says:

> Lists up to 10,000 objects. To control a large list of containers beyond 10,000 objects, use the `marker` and `endMarker` parameters in the `ListContainerOptions` class.

andrewgaul:
@archupsg03 The more I think about this, let's work on the portable abstraction code in a separate pull request.  Please focus this on only the provider, i.e., remove `RegionScopedSwiftBlobStore` changes and only add `RegionScopedSwiftBlobStore`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/RegionScopedSwiftBlobStore.java
archupsg03: Does DLO, bulk deletes  are also expected to delete only the manifest file ?

andrewgaul: Yes.  DLO generally is poorly thought and and supported.  Again, scope this commit down to just the provider changes and submit the portable abstraction changes as a separate pull request.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/options/OpenStackSwiftPutOptions.java
archupsg03: I am almost done with rest of the changes, except the separation of lower and higher API.
My understanding from the comment #1090, is that DynamicLargeObjetApi should have methods such as initiateMultiPartUpload, uploadMultiPart, completeMultiPart etc., Is this correct ? If not
 Can you just give an example, say what methods are expected to in DynamiCLargeObjectApi file  ?

andrewgaul: No, you still don't understand the difference between the provider API and the portable abstraction.  Go compare something like B2 with its [MultipartApi](https://github.com/jclouds/jclouds/blob/master/providers/b2/src/main/java/org/jclouds/b2/features/MultipartApi.java) against [B2BlobStore](https://github.com/jclouds/jclouds/blob/master/providers/b2/src/main/java/org/jclouds/b2/blobstore/B2BlobStore.java).  The former implements the REST API while the latter conforms to the `BlobStore` interface.  This pull request should not contain any of the latter.

archupsg03:
No changes are required in RegionScopedBlobStore. Working on the DynamicLargeObjectAPI. Will submit PR for that. Closing this PR.

PR #1091, opened 2017-04-28
JCLOUDS-1269; JCLOUDS-1120: ec2 subnet/securityGroups in non-default vpc

aledsage:
Builds on @andreaturli's https://github.com/andreaturli/jclouds/commits/feature/aws-ec2-subnet, to fix:
* https://issues.apache.org/jira/browse/JCLOUDS-1269: Jclouds doesn't get the VPC when subnetId parameter is provided
* https://issues.apache.org/jira/browse/JCLOUDS-1120: Security groups don't work with non-default VPCs

The first two commits are identical to Andrea's branch (they have just been rebased against master, hence having different sha1 ids).


providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/compute/loaders/AWSEC2CreateSecurityGroupIfNeeded.java
```diff
@@ -102,7 +104,17 @@ private String createSecurityGroupInRegion(String region, String name, String vp
                                .build());
             }
 
-            String myOwnerId = Iterables.get(securityApi.describeSecurityGroupsInRegion(region, name), 0).getOwnerId();
+            // Use filter (as per `SecurityGroupPresent`, in securityGroupEventualConsistencyDelay)
+            Set<SecurityGroup> securityGroups = securityApi.describeSecurityGroupsInRegionWithFilter(region,
+                  ImmutableMultimap.of("group-name", name));
```
neykov: You could use the `id` of the security group with a `group-id` filter here. Same for `securityGroupEventualConsistencyDelay`, by passing `new RegionAndName(region, id)`. The ID is unique for the region, while the name can be the same between VPCs - will guarantee that you don't get multiple results here.

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/AWSSubnetApiLiveTest.java
```diff
@@ -16,12 +16,67 @@
...
+   @BeforeClass(groups = { "integration", "live" })
+   public void setupContext() {
+      super.setupContext();
+      api = view.unwrapApi(AWSEC2Api.class);
+   }
+
+   @Test
+   public void testCreateSubnetInRegion() {
+//      subnet = subnetApi().createSubnetInRegion();
+//      assertNotNull(subnet);
```
neykov: Why is the test code commented out?

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/features/AWSSubnetApi.java
```diff
@@ -0,0 +1,117 @@
...
+    * @param cidrBlock The network range for the subnet, in CIDR notation. For example, 10.0.0.0/24.
+    * @param options
+    * @return AWS Subnet
+    */
+   @Named("CreateSubnet")
+   @POST
+   @Path("/")
+   @XMLResponseParser(SubnetHandler.class)
+   @FormParams(keys = ACTION, values = "CreateSubnet")
+   Subnet createSubnetInRegion(
```
geomacy: you will need to add `DeleteSubnet` as well - this class will not be very usable if it can only create subnets

apis/ec2/src/main/java/org/jclouds/ec2/compute/EC2ComputeService.java
```diff
@@ -220,10 +223,16 @@ void deleteSecurityGroup(String region, String group) {
       checkNotNull(emptyToNull(group), "group must be defined");
       String groupName = namingConvention.create().sharedNameForGroup(group);
 
-      if (!client.getSecurityGroupApi().get().describeSecurityGroupsInRegion(region, groupName).isEmpty()) {
+      Multimap<String, String> securityGroupFilterByName = ImmutableMultimap.of("group-name", groupName);
+      Set<SecurityGroup> securityGroupsToDelete = client.getSecurityGroupApi().get()
+              .describeSecurityGroupsInRegionWithFilter(region, securityGroupFilterByName);
+      if (securityGroupsToDelete.size() > 1) {
+         logger.warn("When trying to delete security group {} found more than one matching the name. Will delete all - {}.",
+                 group, securityGroupsToDelete);
```
nacx: Don't assume logback. jclouds logger uses `%s` for variable substitution.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/compute/strategy/CreateKeyPairPlacementAndSecurityGroupsAsNeededAndReturnRunOptions.java
```diff
@@ -202,8 +203,8 @@ protected void addSecurityGroups(String region, String group, Template template,
    }
    
    @VisibleForTesting
-   String vpcIdForSubnet(String subnetId) {
-      Optional<Subnet> subnet = awsEC2Api.getSubnetApi().get().filter(new SubnetFilterBuilder().subnetId(subnetId).build()).first();
+   String vpcIdForSubnet(String region, String subnetId) {
+      Optional<Subnet> subnet = Iterables.tryFind(awsEC2Api.getAWSSubnetApi().get().describeSubnetsInRegion(region, subnetId), Predicates.<Subnet>notNull());
```
nacx: [minor] Better `getAWSSubnetApiForRegion` ?

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/AWSSubnetApiMockTest.java
```diff
@@ -0,0 +1,88 @@
...
+            new CreateSubnetOptions().dryRun().availabilityZone("us-east-1a"));
+      assertEquals(result.getVpcId(), "vpc-1a2b3c4d");
+      assertEquals(result.getCidrBlock(), "10.0.1.0/24");
+      assertEquals(result.getAvailabilityZone(), "us-east-1a");
+      assertEquals(result.getSubnetId(), "subnet-9d4a7b6c");
+      assertEquals(result.getSubnetState().value(), "pending");
+      
+      assertPosted(DEFAULT_REGION, "Action=DescribeRegions");
+      assertPosted(DEFAULT_REGION, "Action=CreateSubnet&VpcId=vpc-1a2b3c4d&CidrBlock=10.0.1.0/24&DryRun=true&AvailabilityZone=us-east-1a");
+   }
```
nacx: Add the tests for the `describe` methods as well, and add the tests that exercise the defined fallbacks too.

apis/ec2/src/main/java/org/jclouds/ec2/compute/EC2ComputeService.java
```diff
@@ -220,10 +223,16 @@ void deleteSecurityGroup(String region, String group) {
       checkNotNull(emptyToNull(group), "group must be defined");
       String groupName = namingConvention.create().sharedNameForGroup(group);
 
-      if (!client.getSecurityGroupApi().get().describeSecurityGroupsInRegion(region, groupName).isEmpty()) {
+      Multimap<String, String> securityGroupFilterByName = ImmutableMultimap.of("group-name", groupName);
+      Set<SecurityGroup> securityGroupsToDelete = client.getSecurityGroupApi().get()
+              .describeSecurityGroupsInRegionWithFilter(region, securityGroupFilterByName);
```
geomacy: @neykov will your comment about security group name filtering [here](https://github.com/jclouds/jclouds/pull/1091/files#r113952626) also apply in this code?

geomacy: I guess this is already including the region value so maybe this is fine as-is?

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/options/CreateSubnetOptions.java
```diff
@@ -0,0 +1,83 @@
...
+ * operation. <h2>
+ * Usage</h2> The recommended way to instantiate a CreateSubnetOptions
+ * object is to statically import CreateSubnetOptions.Builder.* and
+ * invoke a static creation method followed by an instance mutator (if needed):
+ * <p/>
+ * <code>
+ * import static org.jclouds.aws.ec2.options.CreateSubnetOptions.Builder.*
+ * <p/>
+ * AWSEC2Api connection = // get connection
+ * group = connection.getAWSSubnetApi().createSubnet(vpcId("123125").noReboot());
```
geomacy: Is this right?  `createSubnetInRegion` is there but not `createSubnet`, and this class doesn't contain `vpcId()` or `noReboot()` methods?

aledsage:
Thanks @neykov @nacx @geomacy I've incorporated those review comments, except for `Better getAWSSubnetApiForRegion ?`

That's a really interesting one - when I picked this up from @andreaturli, I'd forgotten that method existed. We could switch to using `awsEc2Api.getAWSSubnetApiForRegion().filter(...)`, rather than `awsEc2Api.getSubnetAWSSubnetApi().describeSubnetsInRegionWithFilter(region, ...)`. If we did that, we could even delete the new method `describeSubnetsInRegionWithFilter`; and rename `createSubnetInRegion` to `createSubnet` (same for `deleteSubnetInRegion`).

However, I see that many of the other classes in `org.jclouds.aws.ec2.features` use the `xyzInRegion(region, ...)` naming convention. It therefore feels more consistent (and thus better) to keep @andreaturli's method names.

In fact, I'd say that it's the pre-existing `org.jclouds.ec2.features.SubnetApi` that are inconsistent with the names `list()` and `filter()` - most things in `ec2` use the `...InRegion` suffix for the method name.

geomacy:
+1 looks good to me

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/AWSSubnetApiMockTest.java
```diff
@@ -82,6 +118,82 @@ public void createSubnetInRegion_options() throws Exception {
...
+      assertEquals(result.getAvailableIpAddressCount(), 251);
+      
+      assertPosted(DEFAULT_REGION, "Action=DescribeRegions");
+      assertPosted(DEFAULT_REGION, "Action=DescribeSubnets&Filter.1.Name=vpc-id&Filter.1.Value.1=vpc-1a2b3c4d&Filter.2.Name=availabilityZone&Filter.2.Value.1=us-east-1a");
+   }
+
+   public void describeSubnetInRegionWhenEmpty() throws Exception {
+      enqueueRegions(DEFAULT_REGION);
+      enqueue(DEFAULT_REGION,
+            new MockResponse().setBody(describeSubnetsEmptyResponse));
```
nacx: These "empty" tests trigger the same jclouds logic than the other ones. Instead of enqueuing an OK response with an empty list, just enqueue a 404 response without a body to trigger the `@Fallback` defined in the api methods. The expected results should be the same.

nacx:
@aledsage agree. There's a lot of legacy code in those APIs. In other providers there is just the `getApiForRegion(region)` variant, which uses the default region if the parameter is `null`. For consistency I think it is OK to leave that as-is, otherwise the right thing to do (out of the scope of this PR IMO) would be to update all methods to use the proposed approach.

aledsage:
@nacx (cc @geomacy @neykov @andreaturli) thanks for spotting the missing tests.I've added more expect/mock tests, including for various existing methods that did not already have mock/expect tests.

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/AWSSecurityGroupApiMockTest.java
```diff
@@ -0,0 +1,270 @@
...
+      assertEquals(result.getOwnerId(), "123456789012");
+      assertEquals(result.getDescription(), "Web Servers");
+
+      assertPosted(DEFAULT_REGION, "Action=DescribeRegions");
+      assertPosted(DEFAULT_REGION, "Action=DescribeSecurityGroups");
+   }
+
+   public void describeSecurityGroupsGiving404() throws Exception {
+      enqueueRegions(DEFAULT_REGION);
+      enqueue(DEFAULT_REGION, new MockResponse().setResponseCode(404));
```
andreaturli: [minor] there is also `enqueueXml` in case you missed it, which makes the test slightly more readable IMHO

aledsage: Ah, thanks for the pointer @andreaturli - I didn't know about that.

I'll leave it as-is here if that's ok with you. The response strings are defined in the class `AWSSecurityGroupApiMockTest`, rather than external resource files (following the pattern you used in your commit for `AWSSubnetApiMockTest`). Therefore `enqueueXml` wouldn't work unless we moved each response into its own resource file.

andreaturli: sounds good, @aledsage

andreaturli:
@nacx unless you have other comments, I think we can merge that. Ok?

geomacy:
+1 to merging this but I have been trying this out and I think we will need to extend it for practical purposes; if you want to create a VPC and subnet and then deploy a machine on to it, you also need to jump through a few other hoops apart from creating the subnet:
- modify the subnet attributes to permit auto-assign public IP ("ModifySubnetAttribute")
- create an Internet Gateway on the VPC ("CreateInternetGateway")
- get and then modify the routing table of the subnet to add a public (0.0.0.0/0) route through the newly added gateway ("CreateRoute" and friends)

ygy:
@geomacy Very good suggestion. I think what you have described is a known limitation that can be addressed in a subsequent PR.

andreaturli:
@aledsage mind squash the commits into one so that I can merge it easily (as both master and 2.0.x may want this, right?)

aledsage:
@andreaturli done - squashed into one commit.

And yes please, I'd like this in 2.0.x as well!

andreaturli:
thanks @aledsage! merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a9cc10c4) and backported to [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a9cc10c4)

PR #1092, opened 2017-05-04
Fix logging in nova: don’t assume logback

aledsage:
@nacx pointed out a similar mistake I'd made in my PR at https://github.com/jclouds/jclouds/pull/1091#discussion_r114330959

So I grep'ed the repo, and found this one (which assumes logback, rather than the jclouds logger format of `%s` for variable substitution).


nacx:
The CI slave failed, but the build is OK. Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/fe134cf6) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/df5427f7). Thanks, @aledsage!

PR #1094, opened 2017-05-08
fix for deleting an invalid file

ttulka:
If the file for delete is invalid (typically a wrong filename) an IOException will be thrown.

### DIFF
--- a/apis/filesystem/src/test/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImplTest.java
+++ b/apis/filesystem/src/test/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImplTest.java
@@ -29,6 +29,7 @@ import static org.testng.Assert.fail;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.nio.file.InvalidPathException;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


ttulka:
In the version 2.x is a different behaviour then in 1.9.x -> when a file is invalid InvalidPathException is thrown instead of an expected IOException

andrewgaul:
@ttulka Please add a unit test to `FilesystemStorageStrategyImplTest` which exercises this behavior.

ttulka:
The build was aborded because of a timeout, the build self and all the test ran sucessfully.
Could you please fix this and let run the build again? Thank you.

nacx:
rebuild please

andrewgaul:
@ttulka I do not understand your test -- it succeeds on Linux without the fix. Do you test on Windows? If so maybe we can make this fail on Linux as well by adding a trailing forward slash, `/`?

ttulka:
The behavious of the method `delete` in `Utils.java` was in version 1.9.0 so:
- if the file in the parameter is correct, delete it
- if something goes wrong, don't do anything 

But with NIO (version > 1.9.3) this works differently for the files with invalid names (for example a file name with a colon ":" on Windows). In this case `InvalidPathException` is thrown. This needs a change in the user's code.

andrewgaul:
Your test should fail before your change and succeed after, yet it succeeds before and after for me.

apis/filesystem/src/test/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImplTest.java
```diff
@@ -684,6 +685,16 @@ public void testPutIncorrectContentLength() throws Exception {
          // expected
       }
    }
+   
+   @Test
+   public void testDeletingInvalidPathFileEndsNormally() {
+      String invalidPathBlobKey = "A<!:!@#$%^&*?]8 ";
+      try {
+         storageStrategy.removeBlob(CONTAINER_NAME, invalidPathBlobKey);
```
andrewgaul: Call `Fail.failBecauseExceptionWasNotThrown(InvalidPathException.class)`.

ttulka:
I tested only on Windows, so I just added two invalid Unix characters. This should be better.

andrewgaul:
Squashed, removed unneeded whitespace, propagated exception cause, and pushed to master as 6452960c72a5f8247acae12481b94cdf9d8e4218 and 2.0.x as 9bf7c5da5acdfeac1e9298964cb591a142b48afd.  Thank you for your contribution @ttulka!

PR #1096, opened 2017-05-09
test for deleting an invalid path file

ttulka:
test for #1094 (fix for deleting an invalid file)


andrewgaul:
@ttulka I do not understand this pull request -- it succeeds on Linux without the fix in #1094.  Do you test on Windows?  If so maybe we can make this fail by adding a trailing forward slash, `/`?

PR #1097, opened 2017-05-09
Add an InternetGatewayApi to AWSEC2Api

geomacy:
This is a follow-up to https://github.com/jclouds/jclouds/pull/1091 and particularly the comment at https://github.com/jclouds/jclouds/pull/1091#issuecomment-299202429:

> I have been trying this out and I think we will need to extend it for practical purposes; if you want to create a VPC and subnet and then deploy a machine on to it, you also need to jump through a few other hoops apart from creating the subnet:
> 
> - modify the subnet attributes to permit auto-assign public IP ("ModifySubnetAttribute")
> - create an Internet Gateway on the VPC ("CreateInternetGateway")
> - get and then modify the routing table of the subnet to add a public (0.0.0.0/0) route through the newly added gateway ("CreateRoute" and friends)

There are three AWS APIs needed for the above, an `InternetGatewayApi`, a `RouteTableApi` and a method in the subnet API, I guess, for modifying attributes on subnets. This PR contains the `InternetGatewayApi`.  If this looks good, I can add the others.

@neykov, @andreaturli, this will be of interest to you.


nacx:
Agree with what you said (I haven't gone through the code; just commenting on your proposal)

A common use case for private subnets in a VPC is to create layered applications, where each VM belongs to a layer and can be on a different subnet. Not all VMs should be reachable from the outside but traffic between them should be properly routed, and VMs with private IPs that should connect to the Internet (OS package updates, etc), should be able to do so. In AWS, this is achieved by using [NAT Gateways](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-nat-gateway.html), that act as the gateway for private networks, but that limits a bit the topology of your network: VMs in the "public" network must have an elastic IP, since the traffic will be routed to the Internet Gateway (without being NATed), and VMs in the "internal" subnets shouldn't have an elastic IP, since the traffic will be routed to the NAT gateway by default.

I'm just commenting this to illustrate a common use case we should try to cover when implementing these APIs. Having the method to set the "auto create publicIP" for a certain subnet would definitely help to model the "public" networks, and we should also include a `NatGatewayApi` to the list of APIs to create. The `RouteTableApi` should allow us to create the table needed to model the public/external routing, as in the link above.

I'll review the code in a while. Many thanks for starting this @geomacy!

geomacy:
Worth mentioning that I restricted the API to the general purpose gateway - didn't do the egress only one. Could be added later.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/AWSEC2Api.java
```diff
@@ -132,4 +133,11 @@
    @Delegate
    Optional<? extends AWSSubnetApi> getAWSSubnetApiForRegion(
            @EndpointParam(parser = RegionToEndpointOrProviderIfNull.class) @Nullable String region);
+
+
+   /**
+    * Provides synchronous access to InternetGateway services.
+    */
+   @Delegate
+   Optional<? extends InternetGatewayApi> getInternetGatewayApi();
```
nacx: Should we better provide the `InRegion` variant for consistency with other APIs?

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/binders/BindInternetGatewayIdsToIndexedFormParams.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package org.jclouds.aws.ec2.binders;
+
+import org.jclouds.aws.util.AWSUtils;
+import org.jclouds.http.HttpRequest;
+import org.jclouds.rest.Binder;
+
+/**
+ * Binds the String [] to form parameters named with InstanceId.index
```
nacx: InternetGatewayId.index

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/InternetGateway.java
```diff
@@ -0,0 +1,84 @@
...
+   }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+
+      public abstract Builder id(String id);
+      public abstract Builder attachmentSet(List<InternetGatewayAttachment> attachmentSet);
+      public abstract Builder tags(Map<String, String> tags);
+
+      @Nullable public abstract String id();
```
nacx: Remove this method. If you need to access it from the builder, you could add a `toBuilder` method to the InternetGateway object.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/InternetGateway.java
```diff
@@ -0,0 +1,84 @@
...
+      @Nullable public abstract List<InternetGatewayAttachment> attachmentSet();
+      @Nullable public abstract Map<String, String> tags();
```
nacx: Hide these methods by removing the public modifier, like `autoBuid`

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/InternetGatewayAttachment.java
```diff
@@ -0,0 +1,85 @@
...
+
+      public abstract Builder state(State state);
+
+      public abstract Builder vpcId(String vpcId);
+
+      @Nullable
+      public abstract State state();
+
+      @Nullable
+      public abstract String vpcId();
```
nacx: Remove these two accessors from the builder.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/InternetGatewayAttachment.java
```diff
@@ -0,0 +1,85 @@
...
+
+      abstract InternetGatewayAttachment autoBuild();
```
nacx: We don't really need this since we are not customizing the default builder behavior. Let's remove this and leave only the abstract build() method.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/features/InternetGatewayApi.java
```diff
@@ -0,0 +1,141 @@
...
+   @Named("DetachInternetGateway")
+   @POST
+   @FormParams(keys = ACTION, values = "DetachInternetGateway")
+   @XMLResponseParser(ReturnValueHandler.class)
+   @Fallback(FalseOnNotFoundOr404.class)
+   Boolean detachInternetGateway(
+      @EndpointParam(parser = RegionToEndpointOrProviderIfNull.class) @Nullable String region,
+      @FormParam("InternetGatewayId") String internetGatewayId,
+      @FormParam("VpcId") String vpcId,
+      InternetGatewayOptions... options);
```
nacx: I know it is already a pattern in AWS, but it is legacy and wrong: I'd change the varargs parameter with an overloaded variant: one method with the options parameter and one without it. We actually don't support passing more than one options object, so the varargs parameter is not a good method signature. It's not the right way to model optional parameters.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/options/InternetGatewayOptions.java
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+package org.jclouds.aws.ec2.options;
+
+import org.jclouds.ec2.options.internal.BaseEC2RequestOptions;
+
+/**
+ * Contains options supported in the Form API for the InternetGateway operations. <h2>
+ * Usage</h2> The recommended way to instantiate such an object is to statically import
+ * CreateInternetGateway.Builder.* and invoke a static creation method followed by an instance mutator
```
nacx: InternetGatewayOptions.Builder

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/xml/DescribeInternetGatewaysResponseHandler.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public class DescribeInternetGatewaysResponseHandler extends
+      ParseSax.HandlerForGeneratedRequestWithResult<FluentIterable<InternetGateway>> {
+   private final InternetGatewayHandler gatewayHandler;
+   private boolean inAttachmentSet;
+   private boolean inTagSet;
+   private Builder<InternetGateway> gateways = ImmutableSet.builder();
+
+   @Inject
+   public DescribeInternetGatewaysResponseHandler(InternetGatewayHandler gatewayHandler) {
```
nacx: Reduce the injection constructor visibility by removing the public modifier.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/xml/InternetGatewayHandler.java
```diff
@@ -0,0 +1,100 @@
...
+   private StringBuilder currentText = new StringBuilder();
+   private InternetGateway.Builder builder = InternetGateway.builder();
+   private final TagSetHandler tagSetHandler;
+   private final InternetGatewayAttachmentSetHandler attachmentSetHandler;
+   private boolean inTagSet;
+   private boolean inAttachmentSet;
+
+
+   @Inject
+   public InternetGatewayHandler(TagSetHandler tagSetHandler, InternetGatewayAttachmentSetHandler attachmentHandler) {
```
nacx: Remove the public modifier.

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/InternetGatewayApiLiveTest.java
```diff
@@ -0,0 +1,146 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * Tests behavior of {@link InternetGatewayApi}
+ */
+@Test(groups = "live", singleThreaded = true)
+public class InternetGatewayApiLiveTest extends BaseComputeServiceContextLiveTest {
+
+   private Logger logger = Logger.CONSOLE;
```
nacx: Test classes are not instantiated by Guice so the jclouds logger won't be properly initialized. In tests we use the `java.util.logging.Logger.getAnonymousLogger()`.

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/InternetGatewayApiLiveTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+import com.google.common.collect.FluentIterable;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * Tests behavior of {@link InternetGatewayApi}
+ */
+@Test(groups = "live", singleThreaded = true)
+public class InternetGatewayApiLiveTest extends BaseComputeServiceContextLiveTest {
```
nacx: [minor] If you don't need access to the portable ComputeService, better extend the `BaseApiLiveTest`.

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/InternetGatewayApiLiveTest.java
```diff
@@ -0,0 +1,146 @@
...
+import org.testng.annotations.Test;
+
+import com.google.common.collect.FluentIterable;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * Tests behavior of {@link InternetGatewayApi}
+ */
+@Test(groups = "live", singleThreaded = true)
```
nacx: I think we can remove the single threaded flag. All tests are linked anyway.

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/InternetGatewayApiMockTest.java
```diff
@@ -0,0 +1,143 @@
...
+      assertTrue(deleted);
+
+      assertPosted(DEFAULT_REGION, "Action=DescribeRegions");
+      assertPosted(DEFAULT_REGION, "Action=DeleteInternetGateway&InternetGatewayId=igw-fada7c9c");
+   }
+
+   private InternetGatewayApi gatewayApi() {
+      return api().getInternetGatewayApi().get();
+   }
+}
```
nacx: Add the tests that exercise the fallbacks defined in the API methods by enqueuing 404 responses.

geomacy:
hi @nacx I have made the changes you suggested, thanks very much for those.  They're done as individual commits for ease of review; I will squash this before any merge, however.

nacx:
Thanks, @geomacy! I think it is ready to squash and merge.

geomacy:
@nacx thanks, changes squashed

nacx:
Thanks @geomacy! Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/cff2f87e) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a911e5ee).

PR #1100, opened 2017-05-16
Add first draft of a RouteTable API.

geomacy:
Implements the AWS RouteTable API.

This is a follow-up to https://github.com/jclouds/jclouds/pull/1097.

Limitations:
- ~~Does not contain ReplaceRoute.~~ Update: I have added ReplaceRoute but see comments in 49c5c47ada
- Does not contain support for VgwRoutePropagation, because I'm not too familiar with this myself. I propose that this could be done as a later PR.


providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/Route.java
```diff
@@ -0,0 +1,106 @@
...
+      public abstract Builder state(RouteState state);
+
+      public abstract Builder origin(String origin);
+
+      @Nullable abstract String destinationCidrBlock();
+      @Nullable abstract String gatewayId();
+      @Nullable abstract RouteState state();
+      @Nullable abstract String origin();
+
+      abstract Route autoBuild();
```
andreaturli: do you need autobuild here or `abstract Route build();` would do it?

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/RouteTableAssociation.java
```diff
@@ -0,0 +1,80 @@
...
+      public abstract Builder routeTableId(String routeTableId);
+      public abstract Builder subnetId(String subnetId);
+      public abstract Builder main(Boolean main);
+
+      @Nullable abstract String id();
+      @Nullable abstract String routeTableId();
+      @Nullable abstract String subnetId();
+      @Nullable abstract Boolean main();
+
+      abstract RouteTableAssociation autoBuild();
```
andreaturli: same comment as `Route` class

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/VPC.java
```diff
@@ -40,17 +40,17 @@
       /**
        * The subnet is not yet available for use.
        */
-      PENDING, UNRECOGNIZED;
```
andreaturli: why unrecognized is not needed?

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/VPC.java
```diff
@@ -60,18 +60,16 @@ public static State fromValue(String v) {
        */
       DEFAULT,
       DEDICATED,
-      HOST,
-      UNRECOGNIZED;
```
andreaturli: why are you removing it?

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/options/RouteTableOptions.java
```diff
@@ -0,0 +1,255 @@
...
+ * RouteTable table = connection.getRouteTableApi().get().createRouteTable(vpcId, dryRun());
+ * <code>
+ *
+ * @see <a
+ * href="http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateRouteTable.html"
+ * />
+ */
+public class RouteTableOptions extends BaseEC2RequestOptions {
+
+   public static final RouteTableOptions NONE = new RouteTableOptions();
```
andreaturli: sometimes we had issue with this pattern, I remember an interesting bug with `dockertemplateoptions`, is it useful? I'd consider removing it

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/Route.java
geomacy: Not sure; I just copied the pattern elsewhere. I'll give it a go!

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/VPC.java
geomacy: Hm, hadn't meant to change VPC, just the classes I introduced - my thought was to throw an exception rather than accept an unknown value (what use would it be?).  However, I see that pattern is used quite extensively in the domain packages; so I'll just stick with that existing way of doing things.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/RouteTableAssociation.java
geomacy: It's needed here for use in `create()`.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/Route.java
geomacy: I tried this but a concrete `build()` is needed in `RouteSetApi`.

geomacy:
hi @andreaturli, I've added a commit addressing the comments you made.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/Route.java
andreaturli: Not sure I follow, looking at Google Auto Value doc it seems possible

https://github.com/google/auto/blob/master/value/userguide/builders.md#in-animaljava

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/Route.java
```diff
@@ -0,0 +1,125 @@
...
+ * A route in an Amazon EC2 Route Table.
+ *
+ * @see <a href="http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Route.html" >doc</a>
+ */
+@AutoValue
+public abstract class Route {
+
+   @Resource
+   @Named(ComputeServiceConstants.COMPUTE_LOGGER)
+   private static Logger logger = Logger.NULL;
```
nacx: This class is not instantiated by Guice, so the logger won't be injected here. Let's just remove the logs in this class; we don't have this kind of warnings in enums.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/Route.java
```diff
@@ -0,0 +1,125 @@
...
+   public abstract String gatewayId();
+
+   @Nullable
+   public abstract RouteState state();
+
+   @Nullable
+   public abstract String origin();
+
+   @SerializedNames({"destinationCidrBlock", "gatewayId", "state"})
+   public static Route create(String destinationCidrBlock, String gatewayId, RouteState state) {
```
nacx: Add the `origin` field too.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/Route.java
nacx: We use the hidden `autoBuild` method when we need to customize how we build objects, for example, to enforce immutable collections. It is not needed here. Just a `public abstract Route build();` should suffice.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/Route.java
```diff
@@ -0,0 +1,125 @@
...
+      public abstract Builder gatewayId(String gatewayId);
+
+      public abstract Builder state(RouteState state);
+
+      public abstract Builder origin(String origin);
+
+      @Nullable abstract String destinationCidrBlock();
+      @Nullable abstract String gatewayId();
+      @Nullable abstract RouteState state();
+      @Nullable abstract String origin();
```
nacx: These non-visible accessors are not needed. Remove them. We use them when we need access to the internal builder values in a custom build method to enforce immutability.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/RouteTable.java
```diff
@@ -0,0 +1,101 @@
...
+   public abstract static class Builder {
+
+      public abstract Builder id(String id);
+      public abstract Builder vpcId(String vpcId);
+      public abstract Builder routeSet(List<Route> routeSet);
+      public abstract Builder associationSet(List<RouteTableAssociation> associationSet);
+      public abstract Builder tags(Map<String, String> tags);
+
+      @Nullable abstract String id();
+      @Nullable abstract String vpcId();
```
nacx: Remove these two hidden accessors.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/RouteTableAssociation.java
nacx: Same comments as before. Keep in mind to remove the unused hidden accessors.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/VPC.java
```diff
@@ -32,6 +37,10 @@
 @AutoValue
 public abstract class VPC {
 
+   @Resource
+   @Named(ComputeServiceConstants.COMPUTE_LOGGER)
+   private static Logger logger = Logger.NULL;
```
nacx: Remove this logger too.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/features/RouteTableApi.java
```diff
@@ -0,0 +1,277 @@
...
+    */
+   @Named("DeleteRouteTable")
+   @POST
+   @FormParams(keys = ACTION, values = "DeleteRouteTable")
+   @XMLResponseParser(ReturnValueHandler.class)
+   @Fallback(Fallbacks.FalseOnNotFoundOr404.class)
+   boolean deleteRouteTable(
+      @EndpointParam(parser = RegionToEndpointOrProviderIfNull.class) @Nullable String region,
+      @FormParam("RouteTableId") String routeTableId,
+      RouteTableOptions options);
```
nacx: This request actually only supports the `dryRun` parameter. Worth changing the signature to just accept a boolean.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/features/RouteTableApi.java
```diff
@@ -0,0 +1,277 @@
...
+    * @return The route table
+    */
+   @Named("CreateRouteTable")
+   @POST
+   @FormParams(keys = ACTION, values = "CreateRouteTable")
+   @XMLResponseParser(CreateRouteTableResponseHandler.class)
+   RouteTable createRouteTable(
+      @EndpointParam(parser = RegionToEndpointOrProviderIfNull.class) @Nullable String region,
+      @FormParam("VpcId") String vpcId,
+      RouteTableOptions options);
```
nacx: This request actually only supports the dryRun parameter. Worth changing the signature to just accept a boolean.
It looks like the options object is more for the `Route` than the `RouteTable` objects. Worth renaming to `RouteOptions`, and create specific options objects for those methods that require them? Otherwise, it can be confusing for users which options can be used with each method. Let's expose just what can be used.

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/RouteTableApiLiveTest.java
```diff
@@ -0,0 +1,242 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * Tests behavior of {@link RouteTableApi}
+ */
+@Test(groups = "live")
+public class RouteTableApiLiveTest extends BaseApiLiveTest<AWSEC2Api> {
+
+   private static final String TEST_REGION = "eu-west-1";
```
nacx: Better read a System property so users can run the tests against their preferred region? Something like `private static final String TEST_REGION = System.getProperty("jclouds.test.region");` and default to `null` to verify that behaviour too? (jclouds should pick the provider's default region).

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/RouteTableApiLiveTest.java
```diff
@@ -0,0 +1,242 @@
...
+      vpc = vpcClient.createVpc(TEST_REGION, VPC_CIDR);
+      assertNotNull(vpc, "Failed to create VPC to test attachments");
+      tagger.applyToResources(ImmutableMap.of("Name", simpleName), ImmutableList.of(vpc.id()));
+
+      // When you create a VPC it automatically gets a route table whose single route has the CIDR of the VPC
+      // and whose "target" is "local".
+      final FluentIterable<RouteTable> routeTables = routeTableApi.describeRouteTables(TEST_REGION);
+      assertNotNull(routeTables, "Failed to return list of RouteTables");
+      boolean foundVpc = false;
+      for (RouteTable rt : routeTables) {
```
nacx: More idiomatic:

```java
Optional<RouteTable> vpcRT = Iterables.tryFind(routeTables, new Predicate<RouteTable>() {
   @Override public boolean apply(RouteTable input) {
      return vpc.id().equals(input.vpcId());
   }
});
assertTrue(vpcRT.isPresent(), "...");
// All other assertions
```

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/RouteTableApiLiveTest.java
```diff
@@ -0,0 +1,242 @@
...
+      final boolean created = routeTableApi.createRoute(TEST_REGION, routeTable.id(),
+         gatewayId(gateway.id())
+            .destinationCidrBlock(TEST_DESTINATION_CIDR));
+      assertTrue(created, "Failed to add route to table " + routeTable.id());
+
+      final ImmutableList<RouteTable> routeTables =
+         routeTableApi.describeRouteTables(TEST_REGION, routeTable.id()).toList();
+      assertEquals(routeTables.size(), 1, "Could not find existing route table " + routeTable.id());
+      boolean foundRoute = false;
+      for (Route route : routeTables.get(0).routeSet()) {
```
nacx: Apply the pattern mentioned above instead of using loops and flags.

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/RouteTableApiLiveTest.java
```diff
@@ -0,0 +1,242 @@
...
+
+      final boolean deleted = routeTableApi.deleteRouteTable(TEST_REGION, routeTable.id());
+      assertTrue(deleted, "Failed to delete route table " + routeTable.id());
+
+      final ImmutableList<RouteTable> after = routeTableApi.describeRouteTables(TEST_REGION, routeTable.id()).toList();
+      assertEquals(after.size(), 0, "Unexpected response to describe after deleting " + routeTable.id() + ": " + after);
+
+   }
+
+   @AfterClass
```
nacx: alwaysRun = true

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/RouteTableApiLiveTest.java
```diff
@@ -0,0 +1,242 @@
...
+      assertTrue(result, "Failed to disassociate " + associationId + " from " + routeTable.id());
+
+      routeTable = routeTableApi.describeRouteTables(TEST_REGION, routeTable.id()).toList().get(0);
+      assertEquals(routeTable.associationSet().size(), 0,
+         "Found associations where none should exist in  " + routeTable.id() + ": " + routeTable.associationSet());
+
+      subnetApi.deleteSubnetInRegion(TEST_REGION, subnet.getSubnetId());
+   }
+
+   @Test(dependsOnMethods = "testDisassociate")
```
nacx: Can it just depend on the `testCreate` to maximize the chances of this test being run in case the associate tests failed?

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/RouteTableApiMockTest.java
```diff
@@ -0,0 +1,162 @@
...
+      assertNotNull(routeTables, "Failed to create route table description object");
+      assertEquals(routeTables.size(), 3, "Failed to return all entries from test data, returned: " + routeTables);
+
+      for (RouteTable table : routeTables) {
+         if (ImmutableList.of("rtb-80a3fae4", "rtb-d4605bb0").contains(table.id())) {
+            assertRoutesForNormalVpc(table, table.id());
+         } else if (table.id().equals("rtb-e6c98381")) {
+            assertRoutesForTestVpc(table, table.id());
+         }
+      }
```
nacx: Verify the request that has been sent.

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/RouteTableApiMockTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+   public void describeRouteTablesWithInvalidStateValue() throws Exception {
+      enqueueRegions(DEFAULT_REGION);
+      enqueueXml(DEFAULT_REGION, "/describe_route_tables_invalid.xml");
+      final ImmutableList<RouteTable> routeTables = routeTableApi().describeRouteTables(DEFAULT_REGION).toList();
+
+      assertNotNull(routeTables, "Failed to create route table description object");
+      assertEquals(routeTables.size(), 1, "Failed to return expected entry from test data, returned: " + routeTables);
+
+      assertEquals(routeTables.get(0).routeSet().get(0).state(), Route.RouteState.UNRECOGNIZED);
```
nacx: Verify the request that has been sent.

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/RouteTableApiMockTest.java
```diff
@@ -0,0 +1,162 @@
...
+         destinationCidrBlock("172.18.19.0/24"));
+      assertTrue(deleted, "Failed to match 'true' in test data response");
+
+      assertPosted(DEFAULT_REGION, "Action=DescribeRegions");
+      assertPosted(DEFAULT_REGION, "Action=DeleteRoute&RouteTableId=rtb-a77f2ac0&DestinationCidrBlock=172.18.19.0/24");
+   }
+
+   private RouteTableApi routeTableApi() {
+      return api().getRouteTableApi().get();
+   }
```
nacx: * Add mock tests that verify the API methods that accept Options objects.
* Add mock tests that enqueue 404 responses for all methods that define such fallbacks, to verify their behavior.

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/domain/Route.java
geomacy: Ah I see the point now.  Thanks gents.

geomacy:
Thanks for the comments @andreaturli, @nacx, will address them as soon as I can

geomacy:
hi @nacx, @andreaturli, I have pushed changes for the review comments made. For convenience of review I have left these as separate commits.  I can squash them before merge.

geomacy:
Note the new test util [enqueueXml(Response.Status, String, String)](https://github.com/jclouds/jclouds/pull/1100/files#diff-8c2525316d73049be95880788b72ccd1R161) in BaseAWSEC2ApiMockTest which may be of use elsewhere.

geomacy:
hi @nacx, @andreaturli, I hope this is good to go now, do you think it can be merged?

nacx:
Looks good, thanks! Mind squashing the commits?

geomacy:
Thanks @nacx, changes squashed.

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/b3d21f96) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/f7390443). Thanks @geomacy!

PR #1102, opened 2017-05-17
Add ModifySubnetAttribute

geomacy:
This is a follow-up to https://github.com/jclouds/jclouds/pull/1091#issuecomment-299202429.

>   +1 to merging this but I have been trying this out and I think we will need to extend it for practical purposes; if you want to create a VPC and subnet and then depl>oy a machine on to it, you also need to jump through a few other hoops apart from creating the subnet:
>   
>   modify the subnet attributes to permit auto-assign public IP ("ModifySubnetAttribute")
>   create an Internet Gateway on the VPC ("CreateInternetGateway")
>   get and then modify the routing table of the subnet to add a public (0.0.0.0/0) route through the newly added gateway ("CreateRoute" and friends)
>   But as mentioned let's merge this as-is, and do any such new stuff as a new PR!

This PR adds the capability to modify the subnet to auto-assign the public IP via `ModifySubnetAttribute`.

It turned out that the `ModifySubnetAttribute` is a method that has been added to the AWS API subsequent to the version `2012-06-01` of the other subnet methods; to allow the selective specification of a different version I am proposing in this PR the introduction of a new annotation `ApiVersionOverride`, q.v. and see its use in https://github.com/jclouds/jclouds/compare/master...geomacy:modify-subnet-attribute?expand=1#diff-aa0b3b1bb310c30b9f8fdbab82794412R145

The annotation is processed by `FormSignerV4`; should I do this in `FormSignerV2` as well?

### DIFF
--- a/apis/sts/src/main/java/org/jclouds/aws/filters/FormSignerV4.java
+++ b/apis/sts/src/main/java/org/jclouds/aws/filters/FormSignerV4.java
@@ -23,6 +23,7 @@ import static com.google.common.hash.Hashing.sha256;
 import static com.google.common.io.BaseEncoding.base16;
 import static com.google.common.net.HttpHeaders.AUTHORIZATION;
 import static com.google.common.net.HttpHeaders.HOST;
+import static org.jclouds.aws.filters.FormSignerUtils.getAnnotatedApiVersion;
 import static org.jclouds.aws.reference.FormParameters.ACTION;
 import static org.jclouds.aws.reference.FormParameters.VERSION;
 import static org.jclouds.http.utils.Queries.queryParser;
@@ -46,6 +47,7 @@ import org.jclouds.providers.ProviderMetadata;
 import org.jclouds.rest.annotations.ApiVersion;
 
 import com.google.common.base.Joiner;
+import com.google.common.base.Optional;
 import com.google.common.base.Splitter;
 import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableMap;
@@ -101,7 +103,18 @@ public final class FormSignerV4 implements FormSigner {
       this.serviceAndRegion = serviceAndRegion;
    }
 
-   @Override public HttpRequest filter(HttpRequest request) throws HttpException {
+   /**
+    * Adds the Authorization header to the request.
+    *
+    * Also if the method for the operation (or its class) is annotated with a version that is higher than the
+    * default (apiVersion), then the Version parameter of the request is set to be the value from the annotation.
+    *
+    * @param request The HTTP request for the API call.
+    * @return The request
+    * @throws HttpException
+    */
+   @Override
+   public HttpRequest filter(HttpRequest request) throws HttpException {
       checkArgument(request.getHeaders().containsKey(HOST), "request is not ready to sign; host not present");
       String host = request.getFirstHeaderOrNull(HOST);
       String form = request.getPayload().getRawContent().toString();
@@ -126,7 +139,15 @@ public final class FormSignerV4 implements FormSigner {
             .replaceHeader("X-Amz-Date", timestamp);
 
       if (!decodedParams.containsKey(VERSION)) {
-         requestBuilder.addFormParam(VERSION, apiVersion);
+         Optional<String> optAnnotatedVersion = getAnnotatedApiVersion(request);
+         if (optAnnotatedVersion.isPresent()) {
+            String annotatedVersion = optAnnotatedVersion.get();
+            // allow an explicit version annotation to _upgrade_ the version past apiVersion (but not downgrade)
+            String greater = annotatedVersion.compareTo(apiVersion) > 0 ? annotatedVersion : apiVersion;
+            requestBuilder.addFormParam(VERSION, greater);
+         } else {
+            requestBuilder.addFormParam(VERSION, apiVersion);
+         }
       }
 
       Credentials credentials = creds.get();
--- a/core/src/main/java/org/jclouds/rest/annotations/ApiVersionOverride.java
+++ b/core/src/main/java/org/jclouds/rest/annotations/ApiVersionOverride.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.rest.annotations;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import javax.inject.Qualifier;
+
+
+/**
+ * @Deprecated The intention is to use @SinceApiVersion for this purpose, but that would affect
+ * a number of APIs, and we would want to have good test coverage before merging that change
+ * (in {@link FormSignerUtils#getAnnotatedApiVersion}). However, there is some issue with certain tests at
+ * present that means we cannot successfully test all APIs that make use of @SinceApiVersion in order
+ * to assure ourselves that FormSignerUtils will not introduce some problem. See
+ * <a href="https://github.com/jclouds/jclouds/pull/1102#issuecomment-302682049">
+ *    comments on github</a> for details
+ * This annotation is introduced as a temporary measure in order to decouple the functionality of
+ * {@link FormSignerUtils#getAnnotatedApiVersion} from @SinceApiVersion and the tests in question.
+ * It can be removed and replaced by @SinceApiVersion when those tests are fixed.
+ *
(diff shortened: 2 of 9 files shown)


apis/sts/src/main/java/org/jclouds/aws/filters/FormSignerV4.java
```diff
@@ -101,7 +105,25 @@
       this.serviceAndRegion = serviceAndRegion;
    }
 
-   @Override public HttpRequest filter(HttpRequest request) throws HttpException {
+   private String getAnnotatedApiVersion(HttpRequest request) {
```
andreaturli: [minor] Although private could you make return `Optional<String>` instead of null ?

core/src/main/java/org/jclouds/rest/annotations/ApiVersionOverride.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Designates that a method overrides the {@link ApiVersion} on the class with a specific value.
+ *
+ * @see ApiVersion
+ */
+@Target({ METHOD })
+@Retention(RUNTIME)
+@Qualifier
+public @interface ApiVersionOverride {
```
andreaturli: I like this idea, although maybe [ARM approach](https://github.com/andreaturli/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeProviderMetadata.java#L100) -- which is meant to solve the same issue -- is slightly more flexible as you can configure the version without re-building the project.

Anyway I think it is really unlikely that a new version would not require a change at level of java implementation, so I think the Annotation is sensible here.

@nacx any further comments?

providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/features/AWSSubnetApi.java
```diff
@@ -130,4 +133,24 @@ void deleteSubnetInRegion(
...
+
+   /**
+    * Modifies a subnet attribute. You can only modify one attribute at a time.
+    *
+    * @param region The region for the subnet
+    * @param subnetId The ID of the subnet
+    * @param options The options containing the attribute to modify. You can only modify one attribute at a time.
+    * @return true if the modification was successful
+    */
+   @ApiVersionOverride("2014-06-15")
```
andreaturli: neat, I like it!

andreaturli:
I think we should edit the FormSignerV2 as well

nacx:
I like the annotation approach and agree that the flexibility needed in ARM, where they deploy new versions of the APIs quite often (but keeping backward compatibility) might not be needed here.

A couple questions:

* It would be worth considering the annotation at class level too. Could you implement that?
* I think it makes sense to reuse the `@SinceApiVersion` annotation for this purpose. It may affect existing methods, but it should be safe, as it just defines the versions where the annotated methods can work. jclouds is already using that annotation in a similar way. For example, providers that have a root API that delegates to feature APIs (such as the [AWSEC2Api](https://github.com/jclouds/jclouds/blob/master/providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/AWSEC2Api.java)) can return an `Optional`. If those methods are annotated with the `@SinceApiVersion`, jclouds will return `absent` instead of the API class (magic is done [here](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/rest/functions/PresentWhenApiVersionLexicographicallyAtOrAfterSinceApiVersion.java)). So, for me, letting the FormSigner (v2 and V4) override the API version for those methods annotated with `@SinceApiVersion` makes sense and I'd prefer this approach instead of introducing a new annotation to jclouds-core.

geomacy:
@nacx 

I can add the annotation at class level.

I originally started using `@SinceApiVersion`, and having the annotation at both method and class level,  but found that various tests broke, because various APIs are annotated `@SinceApiVersion`.    I will take a look at that again, but switched to having a separate annotation as I thought it was safer.

nacx:
Unit tests may fail because they hardcode the expected version. Live tests will tell us if the change is safe. Let's run them for the classes that use the annotation.

geomacy:
Updates made.

I was able to switch to using `@SinceApiVersion`, very pleased with that.   The [key thing](https://github.com/jclouds/jclouds/pull/1102/commits/3bff59d28c5260e18504c6570ad5cb47117a3d48#diff-7d8365e45e83df7988b1084146439d71R158) is that I needed to change the override from  a just-override-it approach to one that will allow the annotation to _upgrade_ the version as compared to `apiVersion` but not _downgrade_ it, which is much safer.  All the unit tests now pass, and my live subnet test too.

By the way how is the `@ApiVersion` actually injected in places like [FormSigner](https://github.com/jclouds/jclouds/blob/master/apis/sts/src/main/java/org/jclouds/aws/filters/FormSignerV4.java#L96), for example?  I think the value is being taken from [AWSEC2ApiMetadata](https://github.com/jclouds/jclouds/blob/master/providers/aws-ec2/src/main/java/org/jclouds/aws/ec2/AWSEC2ApiMetadata.java#L62) but I couldn't work out how jclouds configures Guice (or whatever) to actually wire them together.

nacx:
Good!

Have you been able to run the live tests for the affected APIs?

>By the way how is the @ApiVersion actually injected in places like FormSigner, for example? I think the value is being taken from AWSEC2ApiMetadata but I couldn't work out how jclouds configures Guice (or whatever) to actually wire them together.

The default value is taken from the metadata. It can [be overridden when creating the context](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/ContextBuilder.java#L269), and it is configured and bound [in a Guice module](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/providers/config/BindProviderMetadataContextAndCredentials.java#L72) when [creating the context](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/ContextBuilder.java#L403).

geomacy:
hi @nacx 

thanks for the explanation of the wiring.

I've tried to run live tests, but may not be set up for it - the tests have `@SinceApiVersion` are 
```
find . -name '*Api.java' | xargs grep -l SinceApiVersion | sed 's#.*/##' | while read file ; do   find . -name ${file%.java}LiveTest.java; done
./apis/chef/src/test/java/org/jclouds/chef/ChefApiLiveTest.java
./apis/chef/src/test/java/org/jclouds/chef/features/OrganizationApiLiveTest.java
./apis/ec2/src/test/java/org/jclouds/ec2/features/ElasticIPAddressApiLiveTest.java
./apis/ec2/src/test/java/org/jclouds/ec2/features/InstanceApiLiveTest.java
./apis/openstack-trove/src/test/java/org/jclouds/openstack/trove/v1/features/InstanceApiLiveTest.java
./providers/google-compute-engine/src/test/java/org/jclouds/googlecomputeengine/features/InstanceApiLiveTest.java
./apis/ec2/src/test/java/org/jclouds/ec2/features/SubnetApiLiveTest.java
./apis/cloudstack/src/test/java/org/jclouds/cloudstack/features/TagApiLiveTest.java
./apis/ec2/src/test/java/org/jclouds/ec2/features/WindowsApiLiveTest.java
./providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/features/AWSSubnetApiLiveTest.java
```
I can run `/providers/aws-ec2/.../AWSSubnetApiLiveTest` and `/apis/ec2/.../SubnetApiLiveTest` successfully.

I've tried running `/apis/ec2/.../ElasticIPAddressApiLiveTest` with the same credentials, and I get `org.jclouds.rest.AuthorizationException: POST https://ec2.eu-west-2.amazonaws.com/ HTTP/1.1 -> HTTP/1.1 401 Unauthorized` results for both operations; I have also tried this with a different Amazon account, with the same result.   I suspect I need to do something else to run the test correctly but I'm not sure what? I'm running them as e.g:
```
mvn clean install  -Dtest.ec2.identity=.... -Dtest.ec2.credential="..." -Dtest=InstanceApiLiveTest -Plive
```

When running the Chef tests I get errors complaining about `No such file or directory` for my `.chef` folder, so must need some more setup there, but haven't found any instructions on how to set up the chef tests.

I haven't tried the tests in `/apis/openstack-trove`, `/providers/google-compute-engine`, or `/apis/cloudstack` as the `FormSignerV4` is only used for AWS.

Have you any links to info that I might be able to use to get those ec2 tests going?

geomacy:
p.s. I have checked the same tests as above in `master` and they fail in the same way for me, so it is something to do with my environment or AWS account, not a feature of the PR itself.

nacx:
I don't know why it raises that unauthorized error. I tried with my account with the same results.
Also tried modifying the test to bind the form4 signer by default, without success, and creating a specific test in the awsec2 provider that extends the one in the generic ec2, but with the same auth errors.
@andreaturli Have you recently tested the ec2 api alone in a similar way and can help here?

geomacy:
Ah, so it's something to do with the tests themselves then, not just my environment being wrong?  I take it you get the error in `master` too?

nacx:
>I take it you get the error in master too?
Yes. Something to investigate. But I'd rather keep this PR on hold for a while, just to make sure auth failures don't come (apart from other things) from signature issues?

geomacy:
@nacx ok; I'll try to take a look at the tests to see if I can contribute anything to a fix. I'd be keen to get this merged so I can start using it.

geomacy:
hi @nacx, per our conversation on IRC I have reverted to using a custom annotation, but put a [comment](https://github.com/jclouds/jclouds/pull/1102/files#diff-d646d0465e56af59039c3489721bf61aR29) on it explaining that it is meant only as a temporary measure to decouple the new function in `FormSigner*` from other APIs for the moment.

I have squashed the other commits from this PR into one, and left the custom annotation as a separate commit, so that it's easy to tell it apart, and hopefully even just revert it when the tests are working again.

How do you think that looks?

geomacy:
p.s. I tested the Live tests for AWSEC2SubnetApi again successfully with this, as well as the mock tests

nacx:
Thanks, @geomacy! I've merged it to master and 2.0.x.

PR #1103, opened 2017-05-19
Promote Packet

nacx:
The Packet provider implements all the features from the portable compute abstraction it supports and has been stable for the last releases. There have not been live tests failures in the weekly test builds, so I think it is ready to be promoted.

/cc @andrewgaul @zack-shoylev


andreaturli:
Thanks @nacx !

PR #1105, opened 2017-05-23
JCLOUDS-1281: DLO - Lower Level Provider API changes

archupsg03:


apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,130 @@
...
+    *
+    * @param objectName
+    *           corresponds to {@link SwiftObject#getName()}.
+    */
+   
+   @Deprecated
+   @Named("dynamicLargeObject: removeLargeFile")
+   @DELETE
+   @Fallback(VoidOnNotFoundOr404.class)
+   void removeLargeFile(@PathParam("containerName") String container, @PathParam("objectName") String name);
```
andrewgaul: What does this do differently than `ObjectApi.delete`?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,130 @@
...
+    *
+    * @return {@link SwiftObject#getEtag()} of the object, which is the MD5
+    *         checksum of the concatenated ETag values of the {@code segments}.
+    */
+   @Deprecated
+   @Named("dynamicLargeObject:replaceManifest")
+   @PUT
+   @ResponseParser(ETagHeader.class)
+   @Headers(keys = "X-Object-Manifest", values = "{containerName}/{objectName}/")
+   String replaceManifest(@PathParam("containerName") String containerName, @PathParam("objectName") String objectName,
```
andrewgaul: Why do these methods take a `containerName` parameter but equivalent `StaticLargeObjectApi` methods do not?

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,194 @@
...
+      }
+   }
+
+   @SuppressWarnings("deprecation")
+   @Test
+   public void uploadLargeFile(String regionId) throws IOException, InterruptedException {
+      /*
+       * createRandomFile(SIZE, bigFile); Payload payLoad = new
+       * FilePayload(bigFile);
+       */
```
andrewgaul: Remove commented-out code.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,130 @@
...
+         @BinderParam(BindToJsonPayload.class) List<Segment> segments,
```
andrewgaul: Could `segments` be a `Collection`?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/functions/ContentMetaDataToMap.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Required to update the Content Length of Object File as Sum of Content length
+ * of Segments
+ * 
+ * @param contentMetadata
+ * 
+ * @return Map with Content Metadata.
+ */
+
+public class ContentMetaDataToMap {
```
andrewgaul: What does this do and where is it used?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/functions/ContentMetaDataToMap.java
```diff
@@ -0,0 +1,57 @@
...
+
+   public ImmutableMap<String, String> getContentMetadataForManifest(ContentMetadata contentMetadata) {
```
andrewgaul: Return `Map`.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,130 @@
...
+    * @param headers
+    *           Binds the map to headers, without prefixing/escaping the header
+    *           name/key.
+    *
+    * @return {@link SwiftObject#getEtag()} of the object, which is the MD5
+    *         checksum of the concatenated ETag values of the {@code segments}.
+    */
+   
+   @Deprecated
+   @Named("dynamicLargeObject: uploadLargeFile")
```
andrewgaul: Remove space after colon (repeated below).

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,130 @@
...
+   @PUT
+   @Headers(keys = EXPECT, values = "100-continue")
+   @ResponseParser(ETagHeader.class)
+   String uploadLargeFile(@PathParam("containerName") String container, @PathParam("objectName") String objectName,
```
andrewgaul: I don't understand this API.  Shouldn't callers upload single-part objects of the form, "obj1", "obj2", then call a method which creates the DLO with the name "obj"?

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,194 @@
...
+   public void setup() {
+      super.setup();
+      for (String regionId : regions) {
+         boolean created = getApi().getContainerApi(regionId).create(defaultContainerName);
+         if (!created) {
+            deleteAllObjectsInContainer(regionId, defaultContainerName);
+         }
+      }
+
+      megOf1s = new byte[1024 * 1024];
```
andrewgaul: Use `TestUtils.randomByteSource`.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,194 @@
...
+   }
+
+   protected void assertMegabyteAndETagMatches(String regionId, String containerName, String name, String etag1s) {
+      SwiftObject object1s = getApi().getObjectApi(regionId, containerName).get(name);
+      assertEquals(object1s.getETag(), etag1s);
+      assertEquals(object1s.getPayload().getContentMetadata().getContentLength(), Long.valueOf(1024 * 1024));
+   }
+
+   protected void deleteAllObjectsInContainerDLO(String regionId, final String containerName) {
+      Uninterruptibles.sleepUninterruptibly(10, TimeUnit.SECONDS);
```
andrewgaul: Remove sleep.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,194 @@
...
+import com.google.common.util.concurrent.Uninterruptibles;
+
+@Test(groups = "live", testName = "DynamicLargeObjectApiLiveTest", singleThreaded = true)
+public class DynamicLargeObjectApiLiveTest extends BaseSwiftApiLiveTest {
+
+   private String defaultName = getClass().getSimpleName();
+   private String defaultContainerName = getClass().getSimpleName() + "Container";
+   private byte[] megOf1s;
+   private byte[] megOf2s;
+   String objectName = "myObject";
```
andrewgaul: Add `private` access.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,194 @@
...
+       */
+      List<Segment> list1 = new ArrayList<Segment>();
+      int partNumber = 1;
+      int total_size = 0;
+      // configure the blobstore to use multipart uploading of the file
+      /*
+       * payLoad.getContentMetadata().setContentLength(bigFile.length());
+       * BasePayloadSlicer slicer = new BasePayloadSlicer();
+       */
+      while (partNumber < 3) {
```
andrewgaul: Prefer `for` loop.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,194 @@
...
+         String objName = objectName + "/dlo/" + String.valueOf(partNumber);
```
andrewgaul: Remove call to `String.valueOf`.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,194 @@
...
+         String data = "data".concat(String.valueOf(partNumber));
+         String etag = getApi().getDynamicLargeObjectApi(regionId, defaultContainerName).uploadLargeFile(
+               defaultContainerName, objName, Payloads.newPayload(data), ImmutableMap.of("myfoo", "Bar"),
+               ImmutableMap.of("myfoo", "Bar"));
+         Segment s = new Segment(objName, etag, data.length());
+         assertNotNull(etag);
+         list1.add(s);
+         partNumber++;
+         total_size = total_size + data.length();
```
andrewgaul: Prefer `+=`.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,194 @@
...
+      }
+      String etagOfEtags = getApi().getDynamicLargeObjectApi(regionId, defaultContainerName).replaceManifest(
+            defaultContainerName, objectName, list1, ImmutableMap.of("MyFoo", "Bar"), ImmutableMap.of("MyFoo", "Bar"));
+
+      SwiftObject bigObject = getApi().getObjectApi(regionId, defaultContainerName).get(objectName);
+      assertEquals(bigObject.getETag(), etagOfEtags);
```
andrewgaul: Prefer `assertThat`.  Repeated throughout.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: The method name is misleading. It does the same as you mention in the comment. Will change method name to uploadPart

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: My point of view was like, user using DLO API's to upload object, should be able to delete with same DLO API's, instead of creating new ObjectApi for deleting purpose.

If you mandate the user to use ObjectApi, then this change is not required

andrewgaul: The jclouds provider API should reflect the REST API.  Thus this method is not needed.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/functions/ContentMetaDataToMap.java
archupsg03: Replace Manifest request an argument with Content-metadata

blob.getMetadata().getContentMetadata() return 
[cacheControl=null, contentDisposition=null, contentEncoding=null, contentLanguage=null, contentLength=36701504, contentMD5=null, contentType=application/unknown, expires=null]

But we need Map{content-type=application/unknown}

Static Large Object API also does something similar to this.

Suggestion needed here :(

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: Done

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: Done

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: Done

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: Chnage method name to Upload Part. Done

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: Removed removeLargeFile API.Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,114 @@
...
+    * @return {@link SwiftObject#getEtag()} of the object, which is the MD5
+    *         checksum of the concatenated ETag values of the {@code segments}.
+    */
+   @Deprecated
+   @Named("dynamicLargeObject:replaceManifest")
+   @PUT
+   @ResponseParser(ETagHeader.class)
+   @Headers(keys = "X-Object-Manifest", values = "{containerName}/{objectName}/")
+   String replaceManifest(@PathParam("objectName") String objectName,
+         @BinderParam(BindToJsonPayload.class) Collection<Segment> segments,
```
timuralp: With DLO, manifest header specifies a segment prefix. How does the segments collection get used? This seems like a copy of the SLO API, but I don't think it actually makes sense here -- in the SLO case, the actual JSON manifest object is constructed from the segments.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,114 @@
...
+    *
+    * @return {@link SwiftObject#getEtag()} of the object, which is the MD5
+    *         checksum of the concatenated ETag values of the {@code segments}.
+    */
+   @Deprecated
+   @Named("dynamicLargeObject:replaceManifest")
+   @PUT
+   @ResponseParser(ETagHeader.class)
+   @Headers(keys = "X-Object-Manifest", values = "{containerName}/{objectName}/")
+   String replaceManifest(@PathParam("objectName") String objectName,
```
timuralp: I understand that this is the same as the SLO API, but the replaceManifest name seems like a misnomer. Maybe `putManifest` would make more sense? Don't have a strong preference on this, though.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,114 @@
...
+@RequestFilters(AuthenticateRequest.class)
+@Consumes(APPLICATION_JSON)
+@Path("/{objectName}")
+public interface DynamicLargeObjectApi {
+   /**
+    * Creates or updates a dynamic large object's manifest.
+    *
+    * @param objectName
+    *           corresponds to {@link SwiftObject#getName()}.
+    * @param segments
```
timuralp: This comment does not make sense in the DLO context. See a comment below about the segments parameter.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,114 @@
...
+    * @return {@link SwiftObject#getEtag()} of the object, which is the MD5
+    *         checksum of the concatenated ETag values of the {@code segments}.
+    */
+   
+   @Deprecated
+   @Named("dynamicLargeObject:uploadLargeFile")
+   @PUT
+   @Headers(keys = EXPECT, values = "100-continue")
+   @ResponseParser(ETagHeader.class)
+   String uploadPart(@PathParam("containerName") String container, @PathParam("objectName") String objectName,
```
timuralp: Is this any different from `PutObject`? I think a higher level wrapper is required to make DLO operations make sense, but I don't think this API method is required. Maybe you could explain how you envision `uploadPart` to fit with the higher level primitives.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,114 @@
...
+         @BinderParam(BindToJsonPayload.class) Collection<Segment> segments,
+         @BinderParam(BindObjectMetadataToHeaders.class) Map<String, String> metadata,
+         @BinderParam(BindToHeaders.class) Map<String, String> headers);
+   
+   /**
+    * Creates object segments.
+    *
+    * @param objectName
+    *           corresponds to {@link SwiftObject#getName()}.
+    * @param segments
```
timuralp: Copy/paste comment error? I don't see a segments parameter.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,114 @@
...
+   @Headers(keys = "X-Object-Manifest", values = "{containerName}/{objectName}/")
+   String replaceManifest(@PathParam("objectName") String objectName,
+         @BinderParam(BindToJsonPayload.class) Collection<Segment> segments,
+         @BinderParam(BindObjectMetadataToHeaders.class) Map<String, String> metadata,
+         @BinderParam(BindToHeaders.class) Map<String, String> headers);
+   
+   /**
+    * Creates object segments.
+    *
+    * @param objectName
```
timuralp: I don't see an objectName parameter. However, containerName and blob are not documented.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/functions/ContentMetaDataToMap.java
andrewgaul: But if you don't use it anywhere...?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
andrewgaul: Agree this seems useless.

This pull request should scope exactly to creating and getting DLO objects since the Swift API offers only this.  The portable abstraction will largely not interact with it, with the exception of getting legacy DLO objects.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,175 @@
...
+         Segment s = new Segment(objName, etag, data.length());
+         assertNotNull(etag);
+         segmentList.add(s);
+         total_size += data.length();
+      }
+      String etagOfEtags = getApi().getDynamicLargeObjectApi(regionId, defaultContainerName).replaceManifest(
+            objectName, segmentList, ImmutableMap.of("MyFoo", "Bar"), ImmutableMap.of("MyFoo", "Bar"));
+
+      SwiftObject bigObject = getApi().getObjectApi(regionId, defaultContainerName).get(objectName);
+      assertThat(bigObject.getETag().equals(etagOfEtags));
```
andrewgaul: You misused `assertThat` -- all of these checks are no-ops.  You need to write `assertThat(...).isEqualTo` or `.isTrue`.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,175 @@
...
+         if (!created) {
+            deleteAllObjectsInContainer(regionId, defaultContainerName);
+         }
+      }
+
+     /* megOf1s = new byte[1024 * 1024];
+      megOf2s = new byte[1024 * 1024];
+
+      Arrays.fill(megOf1s, (byte) 1);
+      Arrays.fill(megOf2s, (byte) 2);*/
```
andrewgaul: Again, remove commented code!

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,175 @@
...
+      
+      megOf1s = TestUtils.randomByteSource().slice(0, 1048576);
+      megOf2s = TestUtils.randomByteSource().slice(0, 1048576);
```
andrewgaul: Prefer 1024 * 1024.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,175 @@
...
+      assertThat(bigObject.getMetadata().equals(ImmutableMap.of("myfoo", "Bar")));
+
+      // segments are visible
+      assertThat(getApi().getContainerApi(regionId).get(containerName).getObjectCount().equals(Long.valueOf(3)));
+   }
+
+   protected void assertMegabyteAndETagMatches(String regionId, String containerName, String name, String etag1s) {
+      SwiftObject object1s = getApi().getObjectApi(regionId, containerName).get(name);
+      assertThat(object1s.getETag().equals(etag1s));
+      assertThat(object1s.getPayload().getContentMetadata().getContentLength().equals(Long.valueOf(1024 * 1024)));
```
andrewgaul: Prefer 1024L * 1024L over `valueOf`.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,175 @@
...
+         }
+      }
+
+     /* megOf1s = new byte[1024 * 1024];
+      megOf2s = new byte[1024 * 1024];
+
+      Arrays.fill(megOf1s, (byte) 1);
+      Arrays.fill(megOf2s, (byte) 2);*/
+      
+      megOf1s = TestUtils.randomByteSource().slice(0, 1048576);
```
andrewgaul: Move initialization of these constants to their declarations and mark `static final`.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,175 @@
...
+       ObjectList objects = getApi().getObjectApi(regionId, containerName).list(new ListContainerOptions());
+      if (objects == null) {
+         return;
+      }
+      List<String> pathsToDelete = Lists.transform(objects, new Function<SwiftObject, String>() {
+         public String apply(SwiftObject input) {
+            return containerName + "/" + input.getName();
+         }
+      });
+      if (!pathsToDelete.isEmpty()) {
```
andrewgaul: Remove useless conditional.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,175 @@
...
+      assertThat(object1s.getETag().equals(etag1s));
+      assertThat(object1s.getPayload().getContentMetadata().getContentLength().equals(Long.valueOf(1024 * 1024)));
+   }
+
+   protected void deleteAllObjectsInContainerDLO(String regionId, final String containerName) {
+       ObjectList objects = getApi().getObjectApi(regionId, containerName).list(new ListContainerOptions());
+      if (objects == null) {
+         return;
+      }
+      List<String> pathsToDelete = Lists.transform(objects, new Function<SwiftObject, String>() {
```
andrewgaul: Why do you need this `Function` if you loop over the paths below?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: @timuralp 
As per swift documentation [1], PUT request for manifest requires the objectName alone.
I tried by removing parameters except objectName, the manifest get uploaded properly, but I am facing issue in etag. Swift stat for manifest outputs the md5 of concatenated etags while etag returned in the response from swift is md5 of empty string. 

Need suggestions here. How to address this issue.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: Changed it to putManifets

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Before looping, the object by their name for delete, we need list of object in container.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Its valid check I feel

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: @timuralp Any comments on this ?

timuralp: @archupsg03 I probably was not clear on this. I was suggesting that using the Swift PutOptions class and setting the correct header (X-Object-Manifest) would be sufficient with the existing PutObject API. However, talking to @andrewgaul, I understand the preference for the separate API class, as it makes the DLO separation from the regular object API obvious.

The X-Object-Manifest header is mandatory in the API, so that header must be set -- that is correct.

In terms of your question on behavior, for DLO, Swift's behavior depends on the number of segments. If the number of segments is less than the full listing, Swift computes DLO ETag as an md5 sum of segment ETags. Otherwise, the ETag is not included. [1]

I'll try to look into the prior jclouds DLO documentation to try to understand your question about the 1.9 implementation.

1. https://github.com/openstack/swift/blob/master/swift/common/middleware/dlo.py#L307

archupsg03: @timuralp  

My thought is to have 2 methods for putManifest resembling 1.9, like

String putObjectManifest(@PathParam("container") String container, @PathParam("name") String name); ----> returns etag of empty string

String putObjectManifest(@PathParam("container") String container, @PathParam("name") @ParamParser(ObjectName.class)
      @BinderParam(BindSwiftObjectMetadataToRequest.class) SwiftObject object);
returns etag of manifest content.

Users who were using 1.9, may switch to latest version. We should be able to have the backward compatibility, which is not achievable now.

timuralp: Swift does not return the DLO ETag as a response to PUT. At least I do not observe this with docker-swift and Swift 2.14 (latest release):
```
DEBUG:swiftclient:REQ: curl -i http://localhost:32768/v1/AUTH_test/test/s3proxy.jar -X PUT -H "x-object-meta-mtime: 1440388916.150022" -H "Content-Length: 0" -H "x-object-manifest: test_segments/s3proxy.jar/1440388916.150022/7663104/1000000/" -H "Content-Type: " -H "X-Auth-Token: AUTH_tkadf8c03346214185ba387e3888893305"
DEBUG:swiftclient:RESP STATUS: 201 Created
DEBUG:swiftclient:RESP HEADERS: {u'Content-Length': u'0', u'Last-Modified': u'Tue, 06 Jun 2017 06:09:31 GMT', u'Etag': u'd41d8cd98f00b204e9800998ecf8427e', u'X-Trans-Id': u'tx9825d3a9fadb4c50bf4f8-005936471a', u'Date': u'Tue, 06 Jun 2017 06:09:30 GMT', u'Content-Type': u'text/html; charset=UTF-8', u'X-Openstack-Request-Id': u'tx9825d3a9fadb4c50bf4f8-005936471a'}
```

The ETag above is of the empty string and jclouds should expose exactly that as a response to a putManifest request.

In jclouds 1.9, the interaction you're describing is the end of performing an MPU where a 0-sized DLO manifest is uploaded (0-sized object with the X-Object-Manifest header). The two methods that are implemented are for the case of specifying manifest metadata through an existing Blob or through the object name. I don't think jclouds 2.0 should replicate either of the approaches.

In jclouds 2.0, you could introduce two DLO methods, akin to the SLO setup, where one has only the metadata parameter (```@BinderParam(BindObjectMetadataToHeaders.class) Map<String, String> metadata```), while the other one includes a headers parameter (```@BinderParam(BindToHeaders.class) Map<String, String> headers```). That would be inline with the current code.

We should not attempt to satisfy old clients, as the prior API has been marked deprecated and jclouds should prefer to use SLO whenever possible. DLO is an option only for clients who explicitly have to use it to interact with old data or if they have some other special reason.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/functions/ContentMetaDataToMap.java
archupsg03: Removed this file, since its decided as not need based on Comment https://github.com/jclouds/jclouds/pull/1105#discussion_r120260682

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/blobstore/functions/ContentMetaDataToMap.java
archupsg03: Not required

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: Removed segment parameter based on #1105

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: Done

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: Done

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: @timuralp Have made changes accordingly. Can you have a look ? Suggestions/Comments are welcome

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/domain/Segment.java
```diff
@@ -37,7 +37,7 @@
    @Named("size_bytes")
    private final long sizeBytes;
 
-   private Segment(String path, String etag, long sizeBytes) {
+   public Segment(String path, String etag, long sizeBytes) {
```
timuralp: Is this change still necessary? Is it only used in tests?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,127 @@
...
+    *           corresponds to {@link SwiftObject#getName()}.
+    * @param metadata
+    *           corresponds to {@link SwiftObject#getMetadata()}.
+    * @param headers
+    *           Binds the map to headers, without prefixing/escaping the header
+    *           name/key.
+    *
+    * @return {@link SwiftObject#getEtag()} of the object, which is the MD5
```
timuralp: As a result of a PUT of a 0-sized object with X-Object-Manifest header set, I'm observing that Swift returns an ETag of the 0-sized object. Are you observing a different behavior in your tests?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,127 @@
...
+   
+   /**
+    * Creates or updates a dynamic large object's manifest.
+    *
+    * @param objectName
+    *           corresponds to {@link SwiftObject#getName()}.
+    * @param metadata
+    *           corresponds to {@link SwiftObject#getMetadata()}.
+    *           
+    * @return {@link SwiftObject#getEtag()} of the object, which is the MD5
```
timuralp: Same comment as above about the ETag applies here.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
timuralp: @archupsg03 can you address the questions above? I'm still unclear on why this method is required. The point is that callers can and should call `putObject()` to upload new or replace existing parts of the DLO. Could you explain how you expect this method to be used and why it would be preferred to calling `putObject()`?

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,167 @@
...
+      for (String regionId : regions) {
+         assertReplaceManifest(regionId, defaultContainerName, defaultName);
+         uploadLargeFile(regionId);
+      }
+   }
+
+   @SuppressWarnings("deprecation")
+   @Test
+   public void uploadLargeFile(String regionId) throws IOException, InterruptedException {
+      List<Segment> segmentList = new ArrayList<Segment>();
```
timuralp: This is no longer used in the test. Can we remove this code entirely? On the other hand, it would be useful to upload some blobs and verify that the ETag is returned as expected from a HEAD.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+      String etag1s = objectApi.put(name + "/1", newByteSourcePayload(megOf1s));
+      awaitConsistency();
+      assertMegabyteAndETagMatches(regionId, containerName, name + "/1", etag1s);
+
+      String etag2s = objectApi.put(name + "/2", newByteSourcePayload(megOf2s));
+      awaitConsistency();
+      assertMegabyteAndETagMatches(regionId, containerName, name + "/2", etag2s);
+
+      List<Segment> segments = ImmutableList.<Segment> builder()
```
timuralp: I don't think these segments are ever uploaded?

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
timuralp: If `pathsToDelete` is empty, the for loop would be a NOOP.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: Sorry. Did not went through this comment. Yeah it looks like not needed. I added those methods from Andrew's comment

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: I am getting etag of empty string. Same like what you have mentioned. Its copy paste error.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
archupsg03: I am getting etag of empty string. Same like what you have mentioned. Its copy paste error.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
archupsg03: Done

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
timuralp: Sorry -- I called the method wrong. It's `put()` in `ObjectApi` or `putBlob()` in the `RegionScopedSwiftBlobStore` class. I suspect consumers should call `putBlob()` and then upload the manifest through the API you're adding. You can probably call either one in the tests.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/domain/Segment.java
timuralp: @archupsg03 I think we can remove this change?

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
timuralp: Sorry -- I called the method wrong. It's put() in ObjectApi or putBlob() in the RegionScopedSwiftBlobStore class. I suspect consumers should call putBlob() and then upload the manifest through the API you're adding. You can probably call either one in the tests.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,167 @@
...
+         return;
+      }
+      List<String> pathsToDelete = Lists.transform(objects, new Function<SwiftObject, String>() {
+         public String apply(SwiftObject input) {
+            return containerName + "/" + input.getName();
+         }
+      });
+      if (!pathsToDelete.isEmpty()) {
+         for (String name : pathsToDelete)
+            getApi().getObjectApi(regionId, containerName).delete(name);
```
timuralp: The object API takes in the containerName. Does the containerName still need to be in the object Name, as well? In the above List transformation, we prepend containerName to every object -- that seems wrong or I'm missing something.

archupsg03:
@timuralp Done with almost all the changes. can you review once

timuralp:
@archupsg03 thanks! This looks pretty good. I will try to run the tests this weekend to verify. @andrewgaul is still traveling for next couple of weeks, so we'll have to wait until the first week of July or so to merge this and for his comments on the PR.

archupsg03:
Thanks @timuralp . Were you able to run test sucessfully ?

Sure we ll wait for @andrewgaul

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+   @SuppressWarnings("deprecation")
+   @Test
+   public void uploadLargeFile(String regionId) throws IOException, InterruptedException {
+      int partNumber = 1;
+      int total_size = 0;
+      RegionScopedBlobStoreContext ctx = RegionScopedBlobStoreContext.class.cast(view);
+      BlobStore blobStore = ctx.getBlobStore();
+      // configure the blobstore to use multipart uploading of the file
+      for (int i = partNumber; i <= 3; partNumber++) {
```
timuralp: This looks like a bug. Wouldn't `i` be incremented? As is, I have a test that never completes, but creates 1000s of parts.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,155 @@
...
+         getApi().getObjectApi(regionId, containerName).delete(name);
+      
+   }
+
+   @Override
+   @BeforeClass(groups = "live")
+   public void setup() {
+      super.setup();
+      for (String regionId : regions) {
+         boolean created = getApi().getContainerApi(regionId).create(defaultContainerName);
```
timuralp: You shouldn't have to do this. The jclouds tests provide containers. Check out this method and the S3 provider on how to use them: `blobstore/src/test/java/[...]/integration/internal/BaseBlobStoreIntegrationTest.java:434`. This a copy/paste from the SLO test, but we shouldn't use the same pattern.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,155 @@
...
+         ByteSource payload = ByteSource.wrap(data.getBytes(Charsets.UTF_8));
+         Blob blob = blobStore.blobBuilder(objName)
+               .payload(payload)
+               .build();
+         String etag = blobStore.putBlob(defaultContainerName, blob);
+         assertNotNull(etag);
+         total_size += data.length();
+      }
+      
+      SwiftObject bigObject = getApi().getObjectApi(regionId, defaultContainerName).get(objectName);
```
timuralp: After fixing the above bug, I get a null pointer exception here.

archupsg03:
@timuralp  Have fixed the NPE exception and tests run successfully for me. Can you give a try ?

timuralp:
Great! Tests pass for me. Unfortunately, they take a long time, but I think that's related to the jclouds handling of the containers to be created for testing. @archupsg03 there are some small nits with white space changes in the commits. Before merging, can you make sure to remove these? I'll let @andrewgaul comment on any other changes he'd like to see.

archupsg03:
@timuralp Am not facing any checkstyle exception in the final commit. Can you just point out to few white space

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,136 @@
...
+               .payload(payload)
+               .build();
+         String etag = blobStore.putBlob(defaultContainerName, blob);
+         assertNotNull(etag);
+         total_size += data.length();
+      }
+      
+      getApi().getDynamicLargeObjectApi(regionId, defaultContainerName).putManifest(objectName,
+            ImmutableMap.of("myfoo", "Bar"));
+
```
timuralp: There is trailing white space on this line. I don't know if there is a way to get github to highlight it.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,136 @@
...
+       ObjectList objects = getApi().getObjectApi(regionId, containerName).list(new ListContainerOptions());
+      if (objects == null) {
+         return;
+      }
+      List<String> pathsToDelete = Lists.transform(objects, new Function<SwiftObject, String>() {
+         public String apply(SwiftObject input) {
+            return containerName + "/" + input.getName();
+         }
+      });
+
```
timuralp: Trailing white space here and line 134.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiMockTest.java
```diff
@@ -0,0 +1,122 @@
...
+      server.enqueue(new MockResponse().setBody("").addHeader(ETAG, "d41d8cd98f00b204e9800998ecf8427e"));
+      server.enqueue(addCommonHeaders(new MockResponse().addHeader("X-Object-Manifest", "myContainer/myObject")));
+
+      try {
+         SwiftApi api = api(server.getUrl("/").toString(), "openstack-swift");
+         assertEquals(api.getObjectApi("DFW", containerName).put(objectName.concat("1"), Payloads.newPayload("data1")),
+               "89d903bc35dede724fd52c51437ff5fd");
+         assertEquals(api.getDynamicLargeObjectApi("DFW", containerName).putManifest(objectName,
+               ImmutableMap.of("MyFoo", "Bar"), ImmutableMap.of("MyFoo", "Bar")), "d41d8cd98f00b204e9800998ecf8427e");
+
```
timuralp: Trailing white space here.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiMockTest.java
```diff
@@ -0,0 +1,122 @@
...
+      try {
+         SwiftApi api = api(server.getUrl("/").toString(), "openstack-swift");
+         assertEquals(api.getObjectApi("DFW", containerName).put(objectName.concat("1"), Payloads.newPayload("data1")),
+               "89d903bc35dede724fd52c51437ff5fd");
+         assertEquals(api.getDynamicLargeObjectApi("DFW", containerName).putManifest(objectName,
+               ImmutableMap.of("MyFoo", "Bar"), ImmutableMap.of("MyFoo", "Bar")), "d41d8cd98f00b204e9800998ecf8427e");
+      
+         assertEquals(server.getRequestCount(), 3);
+         assertAuthentication(server);
+
```
timuralp: Trailing white space.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiMockTest.java
```diff
@@ -0,0 +1,122 @@
...
+               ImmutableMap.of("MyFoo", "Bar"), ImmutableMap.of("MyFoo", "Bar")), "d41d8cd98f00b204e9800998ecf8427e");
+      
+         assertEquals(server.getRequestCount(), 3);
+         assertAuthentication(server);
+         
+         RecordedRequest uploadRequest = server.takeRequest();
+         assertEquals(uploadRequest.getRequestLine(),
+               "PUT /v1/MossoCloudFS_5bcf396e-39dd-45ff-93a1-712b9aba90a9/myContainer/myObjectTest1 HTTP/1.1");
+         assertEquals(new String(uploadRequest.getBody()), "data1");
+
```
timuralp: Trailing white space.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/SwiftApi.java
```diff
@@ -67,4 +68,9 @@ ObjectApi getObjectApi(@EndpointParam(parser = RegionToEndpoint.class) String re
    @Path("/{containerName}")
    StaticLargeObjectApi getStaticLargeObjectApi(@EndpointParam(parser = RegionToEndpoint.class) String region,
          @PathParam("containerName") String containerName);
+
```
timuralp: Trailing white space.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,97 @@
...
+    */
+   @Deprecated
+   @Named("dynamicLargeObject:putManifest")
+   @PUT
+   @ResponseParser(ETagHeader.class)
+   @Headers(keys = "X-Object-Manifest", values = "{containerName}/{objectName}/")
+   String putManifest(@PathParam("objectName") String objectName,
+         @BinderParam(BindObjectMetadataToHeaders.class) Map<String, String> metadata,
+         @BinderParam(BindToHeaders.class) Map<String, String> headers);
+
```
timuralp: Trailing white space.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiLiveTest.java
```diff
@@ -0,0 +1,136 @@
...
+         String data = String.format("%s%s", "data", partNumber);
+         ByteSource payload = ByteSource.wrap(data.getBytes(Charsets.UTF_8));
+         Blob blob = blobStore.blobBuilder(objName)
+               .payload(payload)
+               .build();
+         String etag = blobStore.putBlob(defaultContainerName, blob);
+         assertNotNull(etag);
+         total_size += data.length();
+      }
+
```
timuralp: Trailing white space.

andrewgaul:
@archupsg03 Sorry I was not available the last month.  Thanks @timuralp for stepping up!  This looks close to ready, please address review comments so we can merge this.  Should we defer DLO delete for a subsequent pull request?

archupsg03:
@andrewgaul  Removed trailing whitespaces that was visible to me :(. If you are still spacing any issues I think its better you can change accordingly.

archupsg03:
@andrewgaul @timuralp any other blockers for merging this PR

webfolderio:
@archupsg03 Could you please fix those minor issues.

archupsg03:
@andrewgaul . Sure will work on it by this week end.

One more clarification here, incase of swift MPU say you have 3 parts and due to some reason if the uploading of one part failed, current behaviour is it tries continuing with rest of the parts. But expectation is not to proceed if there is a failure in uploading path. If so what is the way to stop all the submitted task to the executor on exception.  Am referring to RegionScopedSwiftBlobStore call to completeMultipartUplaod -> Futures.getUnchecked(list)

andrewgaul:
I do not understand your comment about `completeMultipartUpload` which is not part of this pull request.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiMockTest.java
```diff
@@ -0,0 +1,122 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.net.HttpHeaders;
+import com.squareup.okhttp.mockwebserver.MockResponse;
+import com.squareup.okhttp.mockwebserver.MockWebServer;
+import com.squareup.okhttp.mockwebserver.RecordedRequest;
+
+@Test(groups = "unit", testName = "DynamicLargeObjectApiMockTest")
+public class DynamicLargeObjectApiMockTest extends BaseOpenStackMockTest<SwiftApi> {
+
+   String containerName = "myContainer";
```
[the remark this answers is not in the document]
archupsg03: Error on changing to private visibility.

apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApi.java
```diff
@@ -0,0 +1,97 @@
...
+    * @param objectName
+    *           corresponds to {@link SwiftObject#getName()}.
+    * @param metadata
+    *           corresponds to {@link SwiftObject#getMetadata()}.
+    * @param headers
+    *           Binds the map to headers, without prefixing/escaping the header
+    *           name/key.
+    *
+    * @return {@link SwiftObject#getEtag()} of the object, which is the MD5
+    *         checksum of the concatenated ETag values of the {@code segments}.
```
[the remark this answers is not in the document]
archupsg03: Not getting this comment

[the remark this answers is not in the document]
andrewgaul: In addition to the `@Deprecated` annotation which javac consumes, we can document in Javadoc for humans to refers to the SLO API.

apis/openstack-swift/src/test/java/org/jclouds/openstack/swift/v1/features/DynamicLargeObjectApiMockTest.java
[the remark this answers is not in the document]
andrewgaul: How is that possible?  You do not extend `DynamicLargeObjectApiMockTest` anywhere.  Also this class should be `final`.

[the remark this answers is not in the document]
archupsg03: Made it final. But private is not possible here

andrewgaul:
I added the classes to test this against Rackspace, an OpenStack provider:

https://github.com/andrewgaul/jclouds/tree/Lower_Level_Provider_API

These tests skip and fail:

```
[pool-2-thread-2] Test uploadLargeFile(org.jclouds.rackspace.cloudfiles.us.features.CloudFilesUSDynamicLargeObjectApiLiveTest) skipped.
Test suite progress: tests succeeded: 0, failed: 0, skipped: 1.
Starting test testReplaceManifest(org.jclouds.rackspace.cloudfiles.us.features.CloudFilesUSDynamicLargeObjectApiLiveTest)
[pool-2-thread-1] Test testReplaceManifest(org.jclouds.rackspace.cloudfiles.us.features.CloudFilesUSDynamicLargeObjectApiLiveTest) failed.
Test suite progress: tests succeeded: 0, failed: 1, skipped: 1.
Tests run: 2, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 119.459 sec <<< FAILURE! - in org.jclouds.rackspace.cloudfiles.us.features.CloudFilesUSDynamicLargeObjectApiLiveTest
testReplaceManifest(org.jclouds.rackspace.cloudfiles.us.features.CloudFilesUSDynamicLargeObjectApiLiveTest)  Time elapsed: 96.624 sec  <<< FAILURE!
org.jclouds.http.HttpResponseException: Server rejected operation connecting to PUT https://storage101.syd2.clouddrive.com/v1/MossoCloudFS_27bdd8b2-bd25-4e71-8782-9df11aac95a2/gaul-blobstore2/CloudFilesUSDynamicLargeObjectApiLiveTest/1 HTTP/1.1
        at org.jclouds.http.internal.BaseHttpCommandExecutorService.invoke(BaseHttpCommandExecutorService.java:122)
        at org.jclouds.rest.internal.InvokeHttpMethod.invoke(InvokeHttpMethod.java:90)
        at org.jclouds.rest.internal.InvokeHttpMethod.apply(InvokeHttpMethod.java:73)
        at org.jclouds.rest.internal.InvokeHttpMethod.apply(InvokeHttpMethod.java:44)
        at org.jclouds.reflect.FunctionalReflection$FunctionalInvocationHandler.handleInvocation(FunctionalReflection.java:117) 
        at com.google.common.reflect.AbstractInvocationHandler.invoke(AbstractInvocationHandler.java:87)
        at com.sun.proxy.$Proxy72.put(Unknown Source)
        at org.jclouds.openstack.swift.v1.features.DynamicLargeObjectApiLiveTest.assertReplaceManifest(DynamicLargeObjectApiLiveTest.java:87)
        at org.jclouds.openstack.swift.v1.features.DynamicLargeObjectApiLiveTest.testReplaceManifest(DynamicLargeObjectApiLiveTest.java:50)
```

archupsg03:
Line number 87 in DynamicLargetObjectApiLiveTest is call to ObjectApi.put methods which was existing and this is a basic call. It may be because of some other issue

andrewgaul:
> Line number 87 in DynamicLargetObjectApiLiveTest is call to ObjectApi.put methods which was existing and this is a basic call. It may be because of some other issue

Please investigate this.

andrewgaul:
@archupsg03 Do you plan to continue work on this pull request or should I close it?

archupsg03:
Regarding the rackspace issue, I am not sure of it, and its a basic and existing call. @timuralp Any comment on this

archupsg03:
@andrewgaul, Since we are not working with rackspace cloud, we dont have a setup to test against it. If possible can you share rackspace cloud credentials, so that I can try to figure out the issue

andrewgaul:
I decline to share my Rackspace credentials which are linked to my credit card with some random Internet person.

archupsg03:
ok no problem. But can you check whether the issue is because of one of the 2 Jira's mentioned in above comment.

andrewgaul:
Sorry how do these issues relate to your problem?

Please do not make substantial edits to comments which are better addressed as new comments.  GitHub does not generate mails for these and it ruins the flow of the discussion.

archupsg03:
I am doubting that container doesnot exist in rackspace cloud, so why we are facing the issue. Can you make sure that the container used for DLO exist and then give a try

andrewgaul:
Please just sign up for a Rackspace account which takes a few minutes.  I have spent hours reviewing code with you.

gaul:
The tests fail since they create a container name in the default region then try to create DLOs in a different region.

gaul:
I rebased and squashed your commit and pushed to master as eaf3c779dce63ddc808d5ff8fe3ed9aade10b2f2.  I included some test fixes and pushed to master as 9dc85ecaabb23bdc0b1d6b3b6650f22673631203.

PR #1108, opened 2017-06-03
JCLOUDS-1293 Add custom IOException retry handler for AWS-EC2

stuarthendren:
As all methods use POST we can not use the method to determine if funciton is idempotent.
We therefore set all as idempotent to nullify that check and add a custom
IOException retry handler to determine if commands should be retried on IOException.

The custom hander extends the BackoffLimitedRetryHandler, the current handler, so all other behaviour is not affected.

This does not retry any POST methods unless it's ACTION starts with 'Describe'. These functions are idempotent, and therefore safe to retry.

See JCLOUDS-1293

### DIFF
--- a/providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/compute/strategy/AWSEC2IOExceptionRetryHandlerTest.java
+++ b/providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/compute/strategy/AWSEC2IOExceptionRetryHandlerTest.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/compute/strategy/AWSEC2IOExceptionRetryHandlerTest.java
```diff
@@ -0,0 +1,56 @@
...
+import static org.testng.Assert.assertFalse;
+
+import java.io.IOException;
+
+import org.jclouds.aws.reference.FormParameters;
+import org.jclouds.http.HttpCommand;
+import org.jclouds.http.HttpRequest;
+import org.testng.annotations.Test;
+
+public class AWSEC2IOExceptionRetryHandlerTest {
```
nacx: Can you annotate this test class so it belongs to the `unit` group?

taroyutao:
china azure storage has different endpoint, such as EndpointSuffix=core.chinacloudapi.cn. in common, we need to set EndpointSuffix, so that the request can point to china azure storage.

stuarthendren:
This is for Amazon, not Azure. You probably need to raise your issue
elsewhere
S

providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/compute/strategy/AWSEC2IOExceptionRetryHandlerTest.java
stuarthendren: No problem
done

nacx:
Pushed to master as [5113be22](http://git-wip-us.apache.org/repos/asf/jclouds/commit/5113be22). Thanks @stuarthendren!

PR #1109, opened 2017-06-07
2.0.x Does this support China Azure storage？

taroyutao:
china azure storage has different endpoint, such as EndpointSuffix=core.chinacloudapi.cn. in common, we need to set EndpointSuffix, so that the request can point to china azure storage does this library support set the EndpointSuffix parameter？


andrewgaul:
Pull requests are not an appropriate place to file issues; please use [JIRA](https://issues.apache.org/jira/browse/JCLOUDS) instead.

PR #1110, opened 2017-06-09
JCLOUDS-1306: Fix SG cache invalidation when deleting

neykov:
When deleting a security group on openstack-nova the cache is not invalidated correctly, so the next create operation will fetch the group from local cache instead of re-creating it in the cloud. Subsequent operations will fail due to the missing group and stale ID.

https://issues.apache.org/jira/browse/JCLOUDS-1306


apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
```diff
@@ -191,14 +191,20 @@ public boolean removeSecurityGroup(String id) {
...
       }
 
-      if (sgApi.get().get(groupId) == null) {
-         return false;
+      // Would be nice to delete the group and invalidate the cache atomically - i.e. use a mutex.
+      // Will make sure that a create operation in parallel won't see inconsistent state.
+
+      boolean deleted = sgApi.get().delete(groupId);
+
+      for (SecurityGroupInRegion cachedSg : groupCreator.asMap().values()) {
```
andreaturli: do you need to invalidate if `deleted` is `false` ?

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
```diff
@@ -191,14 +191,20 @@ public boolean removeSecurityGroup(String id) {
          return false;
       }
 
-      if (sgApi.get().get(groupId) == null) {
-         return false;
+      // Would be nice to delete the group and invalidate the cache atomically - i.e. use a mutex.
+      // Will make sure that a create operation in parallel won't see inconsistent state.
+
+      boolean deleted = sgApi.get().delete(groupId);
```
andreaturli: what happens if the `groupId` doesn't exist?

neykov:
The test `deleteSecurityGroupFromAnotherView ` will fail on aws-ec2 and probably other implementations, will address those separately. See https://issues.apache.org/jira/browse/JCLOUDS-1307.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
neykov: The return value is false. This code path is covered in the test `testSecurityGroupCacheInvalidatedWhenDeletedExternally`

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
neykov: Yes - the security group could've been deleted externally and the local cache still holds a reference to the group. This tries to avoid the situation described in https://issues.apache.org/jira/browse/JCLOUDS-1307.

andreaturli: ok thanks

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
andreaturli: I see it now, thanks

andreaturli:
rebuild please

andreaturli:
previous failure looks unrelated, let's run the builder again before merging.

@neykov do you want this to be backported to 2.0.x as well?

neykov:
Yes, I think it would be useful to backport.

andreaturli:
thanks @neykov, merging now

andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/1d4cb6c3) and backported to [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/04ff09e6)

andreaturli:
thanks @neykov

nacx:
Can he JIRA issue be marked as Fixed?

PR #1112, opened 2017-06-09
Improve OS detection and SSH configuration in Packet

nacx:
* Adds Alpine Linux as a known OSFamily to jclouds-compute, and improves the OS detection in Packet.
* Packet has an API to set the key pairs, so this PR removes the default script that implicitly uploads the key via SSH if it is configured in the options.
* Also removes a couple unused classes.


providers/packet/src/main/java/org/jclouds/packet/domain/ActionType.java
```diff
@@ -1,39 +0,0 @@
...
-
-/**
- * Performs an action for the given device. Possible actions include:
-
- power_on
- power_off
- reboot
- rescue: reboot the device into rescue OS.
- */
-public enum ActionType {
```
andreaturli: is this not needed anymore?

providers/packet/src/main/java/org/jclouds/packet/compute/config/PacketComputeServiceContextModule.java
```diff
@@ -80,6 +81,7 @@ protected void configure() {
       });
       bind(TemplateOptions.class).to(PacketTemplateOptions.class);
       bind(CreateNodesInGroupThenAddToSet.class).to(CreateSshKeysThenCreateNodes.class);
+      bind(NodeAndTemplateOptionsToStatement.class).to(NodeAndTemplateOptionsToStatementWithoutPublicKey.class);
```
andreaturli: didn't know that! :)

looks quite useful! thanks

providers/packet/src/main/java/org/jclouds/packet/domain/ActionType.java
nacx: It wasn't used anywhere

andreaturli: ok :D

andreaturli:
lgtm +1

nacx:
Merged to master and 2.0.x. The packet specific parts for 2.0.x have been pushed to the jclouds-labs repo.

PR #1114, opened 2017-06-20
Async interface changes for putBlob

kishore25kumar:


blobstore/src/main/java/org/jclouds/blobstore/internal/BlobStoreContextImpl.java
```diff
@@ -36,18 +41,26 @@
...
       this.consistencyModel = checkNotNull(consistencyModel, "consistencyModel");
       this.blobStore = checkNotNull(blobStore, "blobStore");
       this.utils = checkNotNull(utils, "utils");
       this.blobRequestSigner = checkNotNull(blobRequestSigner, "blobRequestSigner");
+      Binding<AsyncBlobStore> asyncBlobStoreBinding = injector.getExistingBinding(Key.get(AsyncBlobStore.class));
+      if (asyncBlobStoreBinding != null) {
+         asyncBlobStore = asyncBlobStoreBinding.getProvider().get();
+      } else {
+         asyncBlobStore = null;
+      }
```
nacx: Instead of resolving the injection in the constructor, better bind directly an `Optional<AsyncBlobStore>` in a Guice module like we do for the [compute extensions](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/config/BaseComputeServiceContextModule.java#L301-L321). This way you can directly inject the Optional and return it in the getter without the need for null checks after construction.

core/src/main/java/org/jclouds/http/internal/BaseHttpCommandExecutorService.java
```diff
@@ -80,6 +81,64 @@ protected BaseHttpCommandExecutorService(HttpUtils utils, ContentMetadataCodec c
...
+
+            @Override
+            public void onFailure(Throwable t) {
+               handleInvokeAsyncFailure(t, command, finalFuture);
+            }
+         });
+      } catch (Exception e) {
+         handleInvokeAsyncFailure(e, command, finalFuture);
+
+      }
```
nacx: I'm a bit concerned here about the duplicated logic: the retry logic and exception handling duplicate the one in the sync invoke method, with small differences to accommodate the async futures. Would it be worth refactoring the `sync` implementation to use, say, the Guava's [ImmediateFuture](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/util/concurrent/Futures.html#immediateFuture(V)) so we can have just one implementation of both approaches? Would this introduce a non-negligible overhead to the `sync` methods (the common use-case) to make it not worth it?

core/src/main/java/org/jclouds/http/internal/JavaUrlHttpCommandExecutorService.java
```diff
@@ -127,6 +128,11 @@ protected HttpResponse invoke(HttpURLConnection connection) throws IOException,
    }
 
    @Override
+   protected ListenableFuture<HttpResponse> invokeAsync(final HttpURLConnection nativeRequest) {
+      throw new UnsupportedOperationException("unsupported operation");
```
nacx: Add some more information to let users know that what is not supported are the Async calls in general, and suggest using a different HTTP driver.

nacx (commented):
Thanks, @kishore25kumar!
I've added a couple comments for discussion but in general, it looks good!

blobstore/src/main/java/org/jclouds/blobstore/internal/BlobStoreContextImpl.java
kishorebattula: @nacx I understood the example about compute extensions but not able to figure out how to use it in this context. For injecting optional I have to write Provider in AzureBlobStoreContextModule.java? Then I have to override in all the BobStoreContextModules. Is this what you are suggesting?

core/src/main/java/org/jclouds/http/internal/BaseHttpCommandExecutorService.java
kishorebattula: @nacx In the retry handler, we are calling different methods i.e invoke in one case and invokeAsync in another case. I have to think whether we can avoid it or not. We can wrap sync in immediateFuture and I don't think it will have any performance impact as such. Are you suggesting to change the return type of invoke method to Future?

blobstore/src/main/java/org/jclouds/blobstore/internal/BlobStoreContextImpl.java
nacx: Hmmm, I've just seen there is no base class for all BlobStore context modules (unlike in compute). A clean option would be to create that base module class and declare there the method that `@Provides` the `Optional<AsyncBlobStore>` based on the presence of that binding. The context modules for all blobstore providers should extend that class, and then you can just have the optional async blobstore injected. I'd like something like these, so provider code just can inject an `Optional<AsyncBlobStore>` without having to inspect the bindings.

core/src/main/java/org/jclouds/http/internal/BaseHttpCommandExecutorService.java
nacx: >Are you suggesting to change the return type of invoke method to Future?

Yes. It could not look OK, but it is an internal method, and I wanted to explore if we could benefit from having just one way of invoking the underlying transport method. I just wanted to see if the overhead introduced by the immediate future are worth having a cleaner implementation with just one method.

blobstore/src/main/java/org/jclouds/blobstore/internal/BlobStoreContextImpl.java
kishorebattula: @nacx Can you point me to the classes in compute so that I can follow similar pattern here

kishorebattula: @nacx Can you point me to some references in Compute where this pattern exists

nacx: Apologies for the late reply.

The idea is to create a base module class for all blobstore context modules. Something like:

```java
public abstract class BaseBlobStoreServiceContextModule extends AbstractModule {
   @Override
   protected void configure() {
   }

   @Provides
   @Singleton
   protected final Optional<AsyncBlobStore> provideAsyncBlobStore(Injector i) {
      Binding<AsyncBlobStore> binding = i.getExistingBinding(Key.get(AsyncBlobStore.class));
      return binding == null ? Optional.<AsyncBlobStore> absent() : Optional.of(binding.getProvider().get());
   }
}
```
Then you can have the `BlobStoreContextImpl` directly inject the `Optional<AsyncBlobStore>` like this. The previous module populates an optional to the Guice injector based on the presence of a binding for the `AyncBlobStore`. Whether is present or not, the base module class configures the optional in a generic way, and you can have it directly injected in your class, without having to resolve the binding in the constructor:

```java
@Singleton
public class BlobStoreContextImpl extends BaseView implements BlobStoreContext {
   ...
   private final Optional<AsyncBlobStore> asyncBlobStore;

   @Inject
   public BlobStoreContextImpl(@Provider Context backend, @Provider TypeToken<? extends Context> backendType,
         Utils utils, ConsistencyModel consistencyModel, BlobStore blobStore, BlobRequestSigner blobRequestSigner,
         Optional<AsyncBlobStore> asyncBlobStore) {
      ...
      this.asyncBlobStore = asyncBlobStore;
   }
}
```
This configuration will cause the `BlobStoreContextImpl` to get an `absent()` injected for all providers. The only thing that has to be done in those providers where we support async, is to bind the async implementation. In Azure:
```java
public class AzureBlobStoreContextModule extends BaseBlobStoreServiceContextModule {
   @Override
   protected void configure() {
      ...
      bind(AsyncBlobStore.class).to(AsyncAzureBlobStore.class).in(Scopes.SINGLETON);
   }
}
```

So, with the proposed change:

* We have a Guice module that configures the `Optional<AsyncBlobStore>` by default.
* Modules that support async only have to define the binding. A proper optional will be always injected to the `BlobStoreContextImpl` (modules not defining the binding will get an `absent` injected).

The only implication of this approach is that **all blobstore modules, of all providers** (whether they use async or not) need to be changed to extend the new `BaseBlobStoreServiceContextModule` module since it defines the Guice binding for the optional, which is now needed by the `BlobStoreContextImpl` constructor.
This implies small changes in all blobstore providers but IMO it is a cleaner implementation and it makes sense to have a base class for all blobstore Guice modules where we can define common behavior.

Does this make sense?

nacx: Well, inheritance is not really the only option. You can:

* Make a base class with the optional binding and let all blobstore modules extend it, as mentioned above.
* Or you can create a separate module with that optional binding, say `AsyncBlobStoreModule`, and add it to the `defaultModules` in the metadata class of **all** blobstore providers.

The important thing is that all providers must have that binding in the Guice context, whether they use async or not.

kishorebattula:
@nacx Thanks for the detailed description on injecting optional async blob store context. I have followed the inheritance approach and updated the pull request.
I have also made changes to invoke function to return ListenableFuture. 
Please let me know any other changes to be made in this pull request.

nacx:
>I have also made changes to invoke function to return ListenableFuture.

That was the suggestion to simplify the base driver class code, if that didn't introduce significant overhead. 

I think it actually makes sense for the driver implementations to have the invoke and invokeAsync signatures meaningful, returning, or not, a future. What about keeping the old signatures in the abstract methods and just wrap the sync result in an immediateFuture in the base class? This way each driver implementation does not need to worry about returning a Future in their sync method; that should be just the base class and internal jclouds logic.

kishorebattula:
@nacx Sorry I took it wrong. Wrapping the sync result in the future makes more sense. I this way no driver needs to be touched. I have made changes to my pull request.

nacx:
> Wrapping the sync result in the future makes more sense. I this way no driver needs to be touched. I have made changes to my pull request.

My comment wasn't clear. Thanks! :)

kishorebattula:
@nacx Please let me know what are the next steps towards this change.

nacx:
@kishorebattula conceptually and syntactically lgtm. I'd like to have feedback from @andrewgaul though.

kishore25kumar:
@andrewgaul Any update on this. We are currently blocked on this

gaul:

PR #1117, opened 2017-07-07
[JCLOUDS-1318] fix based on nodeTerminatePredicate

andreaturli:
- wait for server deletion, before deleting the security group

I'd like this to be backported to 2.0.x as well


apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/CleanupServer.java
```diff
@@ -95,6 +101,7 @@ public Boolean apply(String id) {
       }
 
       boolean serverDeleted = novaApi.getServerApi(regionAndId.getRegion()).delete(regionAndId.getId());
+      checkState(nodeTerminatedPredicate.apply(id), "server was not destroyed in the configured timeout");
```
geomacy: Is it desirable to throw an exception here, or should we just log a warning?

andreaturli: good question @geomacy, maybe a warning is enough

andreaturli: Thinking about it more, I guess `destroyNode` should not invoke `CleanUpServer` but simply
```
      boolean serverDeleted = novaApi.getServerApi(regionAndId.getRegion()).delete(regionAndId.getId());
      checkState(nodeTerminatedPredicate.apply(id), "server was not destroyed in the configured timeout");
```
as `CleanUpServer` will be invoked by https://github.com/jclouds/jclouds/blob/master/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeService.java#L99 anyways. I'll test it, meanwhile wdyt?

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/config/NovaComputeServiceContextModule.java
```diff
@@ -165,6 +169,14 @@ protected TemplateOptions provideTemplateOptions(Injector injector, TemplateOpti
    }
 
    @Provides
+   @com.google.inject.name.Named(TIMEOUT_NODE_TERMINATED)
+   protected Predicate<String> provideDropletTerminatedPredicate(final NovaApi api, ComputeServiceConstants.Timeouts timeouts,
```
nacx: Rename method to `provideServerTerminatedPredicate`.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/CleanupServer.java
geomacy: No real opinion on that to be honest!   I would say though, that if the `cleanupServer.apply()` is being invoked as per the link above (inside `cleanUpIncidentalResourcesOfDeadNodes`), then you probably want to avoid the exception here, as otherwise the first exception when iterating over `deadNodes` will prevent any remaining nodes even being attempted.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/strategy/ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSet.java
```diff
@@ -95,63 +91,73 @@ protected ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSe
...
-                  "Key Pairs are required by options, but the extension is not available! options: %s", templateOptions);
-         if (templateOptions.getLoginPrivateKey() != null) {
-            String pem = templateOptions.getLoginPrivateKey();
-            KeyPair keyPair = KeyPair.builder().name(templateOptions.getKeyPairName())
-                     .fingerprint(fingerprintPrivateKey(pem)).privateKey(pem).build();
-            keyPairCache.asMap().put(RegionAndName.fromRegionAndName(region, keyPair.getName()), keyPair);
+         checkArgument(novaApi.getKeyPairApi(region).isPresent(),
+                 "Key Pairs are required by options, but the extension is not available! options: %s", templateOptions);
+      }
+      if (templateOptions.shouldGenerateKeyPair()) {
```
nacx: Should we check instead if the user configured inbound ports?
I see that the groups are just passed through to the server options, so n need for the extension api if the user only sets the security group names?

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeServiceAdapter.java
```diff
@@ -123,41 +128,32 @@ public NovaComputeServiceAdapter(NovaApi novaApi, @Region Supplier<Set<String>>
...
       }
 
-      Optional<String> privateKey = Optional.absent();
       if (templateOptions.getKeyPairName() != null) {
          options.keyPairName(templateOptions.getKeyPairName());
-         KeyPair keyPair = keyPairCache.getIfPresent(RegionAndName.fromRegionAndName(template.getLocation().getId(), templateOptions.getKeyPairName()));
-         if (keyPair != null && keyPair.getPrivateKey() != null) {
-            privateKey = Optional.of(keyPair.getPrivateKey());
-            credentialsBuilder.privateKey(privateKey.get());
-         }
```
nacx: We need this code. Users using an existing, pre-created keypair won't be able to login to the node since the private key won't be populated to the login credentials.

nacx:
On a side note, we are changing the name of the tags we use and the life cycle of the key pairs. People could be relying on this, so I'm not sure if it is a good idea to backport this to 2.0.x.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/strategy/ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSet.java
andreaturli: good point, changing it now

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeServiceAdapter.java
andreaturli: right! 
I've re-added same logic in `ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSet`
thanks!

Tested with
```
               templateOptions.generateKeyPair(true);
```
and with
```
               templateOptions.keyPairName("test");
               String privateKey = Files.toString(new File("/Users/andrea/Downloads/test.pem"), Charsets.UTF_8);
               templateOptions.overrideLoginPrivateKey(privateKey);
```
looks good to me

andreaturli:
rebuild please

andreaturli:
@nacx I think the build issue is related to the PR, need to understand why *AdapterExpectTest is soo slow now, I guess it's because the refactoring involves somehow the injector

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/strategy/ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSet.java
```diff
@@ -95,63 +91,78 @@ protected ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSe
...
-            String pem = templateOptions.getLoginPrivateKey();
-            KeyPair keyPair = KeyPair.builder().name(templateOptions.getKeyPairName())
-                     .fingerprint(fingerprintPrivateKey(pem)).privateKey(pem).build();
-            keyPairCache.asMap().put(RegionAndName.fromRegionAndName(region, keyPair.getName()), keyPair);
+         keyPair = generateKeyPair(region, namingConvention.create().sharedNameForGroup(group));
+         // If a private key has not been explicitly set, configure the auto-generated one
+         if (Strings.isNullOrEmpty(templateOptions.getLoginPrivateKey())) {
+            templateOptions.overrideLoginPrivateKey(keyPair.getPrivateKey());
          }
+      } else if (templateOptions.getKeyPairName() != null && templateOptions.getLoginPrivateKey() != null) {
```
nacx: Why checking also the private key here? If users configure a key pair we don't care if they also configure the private key.  Creating a node and not accessing it is a valid use case. Perhaps users access it in another connection (where the private key will be configured). I'd remove the private key check from here and always set the keypair when its name is set.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/strategy/ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSet.java
```diff
@@ -95,63 +91,78 @@ protected ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSe
...
                  templateOptions);
       }
+      if (templateOptions.shouldGenerateKeyPair() || templateOptions.getKeyPairName() != null) {
+         checkArgument(novaApi.getKeyPairApi(region).isPresent(),
+                 "Key Pairs are required by options, but the extension is not available! options: %s", templateOptions);
+      }
 
-      boolean keyPairExtensionPresent = novaApi.getKeyPairApi(region).isPresent();
+      List<Integer> inboundPorts = Ints.asList(templateOptions.getInboundPorts());
+      if (!templateOptions.getGroups().isEmpty() && !inboundPorts.isEmpty()) {
```
nacx: This should be `||`.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/strategy/ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSet.java
andreaturli: I was a bit unsure as well, I'll remove the extra check

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/strategy/ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSet.java
andreaturli: Ops, yes, thanks

andreaturli:
thanks @nacx let's wait for the builder as extra-check.
Meanwhile I'm rebasing and squashing

andreaturli:
rebuild please

andreaturli:
rebuild please

andreaturli:
merging to master only

andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/aa11765b)

neykov:
Heads up, the commit changes how security groups are set in `templateOptions`. Previously the human readable name was accepted. Now the UUID of the security group is required.

nacx:
This is worth mentioning explicitly in the release notes as a breaking change. And we already have other comments. We should create a wiki page where we list them all to make sure we don't forget any when it's time to release.
This said, @andreaturli is there a clean way to preserve the old behavior too?

nacx:
BTW, thanks for the heads up @neykov!

neykov:
Added a jira as a reminder for the release notes - https://issues.apache.org/jira/browse/JCLOUDS-1323.

PR #1118, opened 2017-07-08
Remove clojure bindings

andrewgaul:
These have not seen any development in many years.


demobox:
> Remove clojure bindings

+1 - looks good to me. Thanks for picking this up, @andrewgaul!

PR #1119, opened 2017-07-11
JCLOUDS-1321: Use separate credential stores per context

neykov:
With a shared credential store the configuration of one compute service leaks in all others, causing the wrong credentials to be used when not overridden.

The particular problem this fixes - [Azure sets default username/password for images](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeProviderMetadata.java#L98) which leaks to other providers (through [GetLoginForProviderFromPropertiesAndStoreCredentialsOrReturnNull](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/config/GetLoginForProviderFromPropertiesAndStoreCredentialsOrReturnNull.java#L51)), causing the wrong username/password to be used.
Even if the credentials are scoped to a specific provider in the map, it still will cause cross-talk between differently configured compute services. The cache should only apply to the current compute service.

### DIFF
--- a/core/src/main/java/org/jclouds/rest/config/CredentialStoreModule.java
+++ b/core/src/main/java/org/jclouds/rest/config/CredentialStoreModule.java
@@ -42,7 +42,6 @@ import com.google.inject.TypeLiteral;
 @Beta
 @ConfiguresCredentialStore
 public class CredentialStoreModule extends AbstractModule {
-   private static final Map<String, ByteSource> BACKING = new ConcurrentHashMap<String, ByteSource>();
    private final Map<String, ByteSource> backing;
 
    public CredentialStoreModule(Map<String, ByteSource> backing) {
@@ -50,7 +49,7 @@ public class CredentialStoreModule extends AbstractModule {
    }
 
    public CredentialStoreModule() {
-      this(null);
+      this(new ConcurrentHashMap<String, ByteSource>());
    }
 
    @Override
@@ -59,13 +58,8 @@ public class CredentialStoreModule extends AbstractModule {
       }).to(CredentialsToJsonByteSource.class);
       bind(new TypeLiteral<Function<ByteSource, Credentials>>() {
       }).to(CredentialsFromJsonByteSource.class);
-      if (backing != null) {
-         bind(new TypeLiteral<Map<String, ByteSource>>() {
-         }).toInstance(backing);
-      } else {
-         bind(new TypeLiteral<Map<String, ByteSource>>() {
-         }).toInstance(BACKING);
-      }
+      bind(new TypeLiteral<Map<String, ByteSource>>() {
+      }).toInstance(backing);
    }
 
    public static class CredentialsToJsonByteSource implements Function<Credentials, ByteSource> {
--- a/core/src/test/java/org/jclouds/rest/CredentialStoreModuleTest.java
+++ b/core/src/test/java/org/jclouds/rest/CredentialStoreModuleTest.java
@@ -95,34 +95,38 @@ public class CredentialStoreModuleTest {
 
    }
 
-   public void testDefaultConsistentAcrossMultipleInjectors() throws IOException {
-      Map<String, ByteSource> map = getMap(createInjector());
+   public void testDefaultDifferentAcrossMultipleInjectors() throws IOException {
+      Injector injector = createInjector();
+      Map<String, ByteSource> map = getMap(injector);
+      put(map, getStore(injector), "test", new Credentials("user", "pass"));
 
-      put(map, getStore(createInjector()), "test", new Credentials("user", "pass"));
-      checkConsistent(map, getStore(createInjector()), "test", new Credentials("user", "pass"));
-      checkConsistent(map, getStore(createInjector()), "test", new Credentials("user", "pass"));
-      remove(map, getStore(createInjector()), "test");
-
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/jclouds/rest/config/CredentialStoreModule.java
```diff
@@ -59,13 +59,8 @@ protected void configure() {
       }).to(CredentialsToJsonByteSource.class);
       bind(new TypeLiteral<Function<ByteSource, Credentials>>() {
       }).to(CredentialsFromJsonByteSource.class);
-      if (backing != null) {
-         bind(new TypeLiteral<Map<String, ByteSource>>() {
-         }).toInstance(backing);
-      } else {
-         bind(new TypeLiteral<Map<String, ByteSource>>() {
-         }).toInstance(BACKING);
```
andrewgaul: Remove unused `BACKING`?

neykov: Good spot, removed.

nacx:
I'm not sure we should fix it this way and remove the static. The default static backing map was chosen as a practical way to provide some basic "credential persistence" and allow credentials to be cached and survive context creation/destruction, which is convenient in many cases.

There is a trivial configuration to be applied when you need context isolation. Just create the context with a `CredentialStoreModule` initialized with a local `Map`. It has a public constructor just for this purpose.

```java
Module isolatedCredStore = new CredentialStoreModule(new ConcurrentHashMap<String, ByteSource>());

ContextBuilder.newBuilder("provider")
   .modules(ImmutableSet.of(isolatedCredStore, ...))
   ...
```

I think the current default is a good compromise of convenience vs conflicting use cases (since most users will be using one single account for each provider), and the existing implementation already provides a clean way to isolate the credential store between contexts. I don't see the need to change that default.

neykov:
@nacx That's a great idea, thanks for the example.

Don't disagree having the shared cache could be useful. It's the way it's used in [GetLoginForProviderFromPropertiesAndStoreCredentialsOrReturnNull](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/config/GetLoginForProviderFromPropertiesAndStoreCredentialsOrReturnNull.java) that I don't like. And that's used in all providers.

Having the configuration of one compute service affect all remaining compute services even when they have different configuration is surprising to say the least :)

I still think that it's the wrong thing to do. Perhaps it should be the other way around - users which really need the cross-service cache should inject it themselves.

Can you give some good examples of using the cache? From looking at the code it seems that its not even needed - no heavy computation is done to extract the username/password in the default case.

andrewgaul:
I find the shared credential store surprising.  Instead users should configure via the constructor parameter if they desire this behavior?

andreaturli:
Same here it is very surprising. I agree with @neykov the default should probably be an isolated credentials store, and give the user the ability to override if they really know what they are doing!

demobox:
> give the user the ability to override if they really know what they are doing!

Worth asking in `user@` and/or `dev@` how many users are aware of and/or are using the feature? Irrespective of whether this is surprising or not, given that it's a breaking change it would be useful to know what the (intentional) usage out there of this is..?

nacx:
I'm OK with making the per-context store the default one. My only concern is that users might be relying on this without really knowing it.
If we do that, let's make sure we properly document in the release notes how to keep the old behavior. We should also add, as soon as we merge the change, a `credential store` section to the [jclouds compute guide](http://jclouds.apache.org/start/compute/) to explain what it is, what it is used for, and how you can leverage the shared store.

neykov:
These are all great ideas @nacx. I'll do the necessary updates.
I'll send an email to the mailing list so the change gets wider exposure.
It's also a breaking change so should be merged only on master.

andrewgaul:
> I'm OK with making the per-context store the default one. My only concern is that users might be relying on this without really knowing it.

When I removed a similar static from the transient blobstore in 4ac7629f44e70bf4fd4a17da541745a9769fcdb1 one user complained about it.  I can not find the reference but we should mention the changed behavior in the release notes at least.

neykov:
* Jira (to be included in release notes): https://issues.apache.org/jira/browse/JCLOUDS-1321
* Heads up to the mailing lists:
  * [user@j.a.o](https://lists.apache.org/thread.html/c65711e3ca3b3ffe068fc6562f0db98336772984924f24857ee0bdcd@%3Cuser.jclouds.apache.org%3E)
  * [dev@j.a.o](https://lists.apache.org/thread.html/b3edeca6b742d122908c570dd779af4bb1c93c6405e0bf5b52c17267@%3Cdev.jclouds.apache.org%3E)
* Documenting a work around in the [compute guide](https://github.com/jclouds/jclouds-site/pull/200)

nacx:
Good to merge this and publish the changes to the site with the docs?

neykov:
+1. I'll merge this and publish the changes by the end of the week due to passive consensus.

neykov:
Merged to [master](https://git1-us-west.apache.org/repos/asf/jclouds/?p=jclouds.git;a=commit;h=2487b0c5132191de5e68413e33dfa551b5679ebb).

neykov:
Web site updated.

PR #1120, opened 2017-07-17
JCLOUDS-1322: Remove Expect header for requests with empty body

ChaithanyaGK:
Without request body, there's no point in asking for 100-continue.

With [JavaUrlHttpCommandExecutor](https://github.com/jclouds/jclouds/blob/f3c3f3b30620ad62dc502cf79bf121ec1773396e/core/src/main/java/org/jclouds/http/internal/JavaUrlHttpCommandExecutorService.java), HttpUrlConnection is throwing an [IOException](https://github.com/jclouds/jclouds/blob/f3c3f3b30620ad62dc502cf79bf121ec1773396e/core/src/main/java/org/jclouds/http/internal/JavaUrlHttpCommandExecutorService.java#L96) if we expect 100-continue for an PUT/POST request with empty body although it is returning 201 response code, it is resulting in another network call when we do [getHeaderFields](https://github.com/jclouds/jclouds/blob/f3c3f3b30620ad62dc502cf79bf121ec1773396e/core/src/main/java/org/jclouds/http/internal/JavaUrlHttpCommandExecutorService.java#L113) on the connection.

This issue can be easily replicated by invoking putBlob operation with zero-length blob.
```
ByteSource payload = ByteSource.empty();
Blob blob = blobStore.blobBuilder(blobName)
            .payload(payload)
            .contentLength(payload.size())
            .build();
blobStore.putBlob(containerName, blob);
```


core/src/main/java/org/jclouds/rest/internal/RestAnnotationProcessor.java
```diff
@@ -357,6 +357,22 @@ public GeneratedHttpRequest apply(Invocation invocation) {
       if (request.getPayload() != null) {
          contentMetadataCodec.fromHeaders(request.getPayload().getContentMetadata(), headers);
       }
+
+      boolean isEmptyBody = false;
+      if ("POST".equals(requestMethod) || "PUT".equals(requestMethod)) {
```
nacx: Should we consider "PATCH" operations too?

core/src/main/java/org/jclouds/rest/internal/RestAnnotationProcessor.java
```diff
@@ -357,6 +357,22 @@ public GeneratedHttpRequest apply(Invocation invocation) {
...
+            if (length != null && length == 0) {
+               isEmptyBody = true;
+            }
+         } else {
+            isEmptyBody = true;
+         }
+      }
+      if (isEmptyBody) {
+         request = request.toBuilder().removeHeader("Expect").build();
+      }
```
nacx: Let's refactor this to a method that gets a request and returns the modified one so it can be properly unit-tested.

core/src/main/java/org/jclouds/rest/internal/RestAnnotationProcessor.java
ChaithanyaGK: @nacx Mostly, I came across Expect header being set for either POST or PUT request. But, I couldn't find any mention of that Expect header is restricted to some Http methods. I think we can remove this check altogether. WDYT ?

demobox:
I recall a lot of discussion and trial and error in the past trying to get this to work across providers and versions - from what I recall, there were API bugs to work around etc. etc.

As such, unless we can verify that those conditions no longer exist, I'd be very hesitant to changing more than is absolutely necessary here to get whatever is currently broken working.

@andrewgaul @nacx Do either of you two happen to recall some more details? I'll also go looking a bit through the archives...

demobox:
> I'll also go looking a bit through the archives...

Here's a few:

* https://groups.google.com/forum/#!topic/jclouds-dev/H6Q5mhpCq0c
* https://github.com/rackspace/jclouds/commit/d897c5e44717e3d66062cd1275033e414f6a2de2
* https://issues.apache.org/jira/browse/JCLOUDS-181

apis/s3/src/test/java/org/jclouds/s3/S3ClientMockTest.java
```diff
@@ -73,7 +73,7 @@ public void testZeroLengthPutHasContentLengthHeader() throws IOException, Interr
       RecordedRequest request = server.takeRequest();
       assertEquals(request.getRequestLine(), "PUT /bucket/object HTTP/1.1");
       assertEquals(request.getHeaders(CONTENT_LENGTH), ImmutableList.of("0"));
-      assertEquals(request.getHeaders(EXPECT), ImmutableList.of("100-continue"));
+      assert request.getHeaders(EXPECT).isEmpty();
```
andrewgaul: Call `assertThat(request.getHeaders(EXPECT).isEmpty())`.

core/src/main/java/org/jclouds/rest/internal/RestAnnotationProcessor.java
ChaithanyaGK: @nacx Refactored the code to a separate method and added a unit test case to check the functionality.

apis/s3/src/test/java/org/jclouds/s3/S3ClientMockTest.java
ChaithanyaGK: Done

ChaithanyaGK:
@andrewgaul Please have a look at the Charles screenshot attached, where two calls are made in case of azure putBlob operation with zero length blob.

charles1

I've added a unit test case to check whether Expect header is stripped in the case of the empty request body.  Not sure how to add an integration test to actually check that only one network call is being made. Any leads on this would be very helpful.

core/src/main/java/org/jclouds/rest/internal/RestAnnotationProcessor.java
```diff
@@ -779,6 +782,22 @@ private static void addHeader(Multimap<String, String> headers, Headers header,
...
+   private static GeneratedHttpRequest stripExpectHeaderIfContentZero(GeneratedHttpRequest request) {
+      boolean isBodyEmpty = false;
+      if (request.getPayload() != null) {
+         Long length = request.getPayload().getContentMetadata().getContentLength();
+         if (length != null && length == 0) {
+            isBodyEmpty = true;
+         }
+      } else {
+         isBodyEmpty = true;
+      }
```
nacx: This could be a bit cleaner:
```java
boolean isBodyEmpty = true;
if (request.getPayload() != null) {
   Long length = request.getPayload().getContentMetadata().getContentLength();
   isBodyEmpty = (length == null || length == 0);
}
```

ChaithanyaGK: Thanks @nacx , will update the PR.

nacx: This version also makes it explicit to set the body to `empty` if the `content-length` is `null`. It is unlikely and I'd say jclouds takes care of always setting the content length for payloads with content, but in any case, I think it is more correct (unlike the current implementation) to mark the body as empty if there content-length is not present

ChaithanyaGK: @nacx jclouds don't set the content length in case of `chunked` transfer, that is the reason I'm not considering the body to be empty if content length is not set. Please check [this ](https://github.com/jclouds/jclouds/blob/4bbca9edf943852ce1ea5aa579fa0554f770a3ea/core/src/main/java/org/jclouds/http/HttpUtils.java#L242)

nacx: Could we better add explicitly the `chunked` case here? I think we should be explicit about which exact cases represent empty bodies.

ChaithanyaGK: @nacx I've debugged this further and below are my findings:

1. `content-length` is always set for non-stream payloads like [String](https://github.com/jclouds/jclouds/blob/0bc935dd57dc8009731d05c533edd831c8642664/core/src/main/java/org/jclouds/io/payloads/StringPayload.java#L38),  [ByteArray](https://github.com/jclouds/jclouds/blob/4bbca9edf943852ce1ea5aa579fa0554f770a3ea/core/src/main/java/org/jclouds/io/payloads/ByteArrayPayload.java#L31) etc.,
2. The only scenario I'm worried is in the case of `chunked` transfer with Stream payloads (like [InputStream](https://github.com/jclouds/jclouds/blob/4bbca9edf943852ce1ea5aa579fa0554f770a3ea/core/src/main/java/org/jclouds/io/payloads/InputStreamPayload.java#L23), [ByteSource ](https://github.com/jclouds/jclouds/blob/4bbca9edf943852ce1ea5aa579fa0554f770a3ea/core/src/main/java/org/jclouds/io/payloads/ByteSourcePayload.java#L30)etc., ) how to check the body is empty?

Just to verify the case, I tried to do azure putBlob without setting the content-length for ByteSource payload:

```
ByteSource payload = ByteSource.wrap("testdata".getBytes(Charsets.UTF_8));
Blob blob = blobStore.blobBuilder(blobName)
          		.payload(payload)
       			//.contentLength(payload.size())
           		.build();
blobStore.putBlob(containerName, blob);
```

I'm getting the below error saying I've to specify the content-length (which is not the case with non-stream payloads):
```
Exception in thread "main" java.lang.IllegalArgumentException: size must be set
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:125)
	at org.jclouds.azureblob.binders.BindAzureBlobMetadataToRequest.bindToRequest(BindAzureBlobMetadataToRequest.java:56)
	at org.jclouds.rest.internal.RestAnnotationProcessor.decorateRequest(RestAnnotationProcessor.java:660)
	at org.jclouds.rest.internal.RestAnnotationProcessor.apply(RestAnnotationProcessor.java:354)
	at org.jclouds.rest.internal.RestAnnotationProcessor.apply(RestAnnotationProcessor.java:137)
```
So, I'm wondering whether there would be a scenario where content-length is not set ?

andrewgaul:
> Which existing integration test exercises this code path? If none exists, please add one. Also post the results from some real providers.

You did not address my comment.  You should add a test for putting a zero-length blob to `BaseBlobStoreIntegrationTest` and test this against a few providers.

andrewgaul:
Please open an [JIRA issue](https://issues.apache.org/jira/projects/JCLOUDS) and include the issue in the commit.

andrewgaul:
> As such, unless we can verify that those conditions no longer exist, I'd be very hesitant to changing more than is absolutely necessary here to get whatever is currently broken working.

@ChaithanyaGK could you explain the motivation for this change?  I don't understand what this addresses except optimizing the uncommon case of a zero-length PUT.

apis/s3/src/test/java/org/jclouds/s3/S3ClientMockTest.java
```diff
@@ -73,7 +74,7 @@ public void testZeroLengthPutHasContentLengthHeader() throws IOException, Interr
       RecordedRequest request = server.takeRequest();
       assertEquals(request.getRequestLine(), "PUT /bucket/object HTTP/1.1");
       assertEquals(request.getHeaders(CONTENT_LENGTH), ImmutableList.of("0"));
-      assertEquals(request.getHeaders(EXPECT), ImmutableList.of("100-continue"));
+      assertThat(request.getHeaders(EXPECT).isEmpty());
```
andrewgaul: This is written incorrectly.  Instead:

```java
assertThat(request.getHeaders(EXPECT)).isEmpty();
```

core/src/main/java/org/jclouds/rest/internal/RestAnnotationProcessor.java
andrewgaul: I tested removing Content-Length from `BaseBlobIntegrationTest.testPutInputStream` and found:

requires C-L:

* Atmos
* AWS-S3, [reference](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html)
* Azure Blob, [reference](https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob)
* B2, [reference](https://www.backblaze.com/b2/docs/b2_upload_file.html)

requires C-L but should not:

* GCS, [reference](https://cloud.google.com/storage/docs/json_api/v1/objects/insert)

does not require C-L but should:

* Rackspace Cloud Files/OpenStack Swift, [reference](https://developer.openstack.org/api-ref/object-storage/?expanded=create-or-replace-object-detail#create-or-replace-object)

I do not understand jclouds chunked encoding support well enough to comment but otherwise our portable call sites should set Content-Length.  For what it is worth I prefer the original code which does not assume a Content-Length and not the suggested change.

apis/s3/src/test/java/org/jclouds/s3/S3ClientMockTest.java
ChaithanyaGK: Good spot,  will update.

ChaithanyaGK:
> could you explain the motivation for this change? I don't understand what this addresses except optimizing the rare case of a zero-length PUT.

@andrewgaul  Every time I initiate a multipart upload to azure using S3Proxy it creates an [empty blob](https://github.com/andrewgaul/s3proxy/blob/master/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java#L1905), where I've encountered this issue.

> Please open an JIRA issue and include the issue in the commit.

Updated the PR with JIRA issue-id

core/src/main/java/org/jclouds/rest/internal/RestAnnotationProcessor.java
ChaithanyaGK: Thanks for the analysis @andrewgaul Will stick to the original code.

andrewgaul:
> > could you explain the motivation for this change? I don't understand what this addresses except optimizing the rare case of a zero-length PUT.
> 
> @andrewgaul Every time I initiate a multipart upload to azure using S3Proxy it creates an empty blob, where I've encountered this issue.

So you want to optimize a single round-trip out of a set of thousands or millions with large multi-part uploads?  Seems like a curious pursuit...

ChaithanyaGK:
> You should add a test for putting a zero-length blob to BaseBlobStoreIntegrationTest and test this against a few providers. We probably want a second test for streaming zero-length InputStream payloads as well.

@andrewgaul I've added integration tests for zero-length `String`, `ByteSource` and `InputStream` payloads,  and successfully tested against Azure blob and aws-s3

ChaithanyaGK:
@andrewgaul Could you please let me know is there anything else that needs to be addressed here.

andrewgaul:
To clear up any confusion, the proposed change introduces a regression in google-cloud-storage and must be addressed before merging.  Previously zero-byte writes succeeded and with this pull request they fail.

ChaithanyaGK:
@andrewgaul This PR doesn't introduce any regression. `putBlob` with zero-length inputstream is failing before as well on google-cloud. The thing is we don't have a test case for that scenario before, so you didn't see any test failure.

andrewgaul:
Thanks for investigating the failure which I addressed in 9e73bbec16cbf91c5f228d4c3ae99c6c526081f8.  You will need to rebase your commit since I included your integration tests to validate my fix.  I guess this is fine to merge although I still do not understand why you are so motivated to optimize an uncommon path.  I have asked this twice before and did not received an answer -- which benchmark do you have which shows that this is a performance issue?

ChaithanyaGK:
@andrewgaul  Currently, I don't have any performance numbers with me. As I have mentioned before, we were comparing the no. of requests being made for a multi-part upload using S3Proxy vs using Azure directly, where I've encountered this redundant request and raised the PR to fix the same.

andrewgaul:
Pushed to master as e331a000d17d36043165cf79c505a7c0a7370f05.  I still do not understand the usefulness but there is no technical argument against this change.  If you want to improve on this commit you could investigate which providers actually need/want 100-continue and provide configuration to disable 100-continue for small, non-zero payloads.

PR #1124, opened 2017-07-27
JCLOUDS-1325: Ignore Unicode BOM in XML parser

andrewgaul:
This caused failures to parse Azure Queue Storage list requests.

### DIFF
--- a/core/src/main/java/org/jclouds/http/functions/ParseXMLWithJAXB.java
+++ b/core/src/main/java/org/jclouds/http/functions/ParseXMLWithJAXB.java
@@ -82,7 +82,8 @@ public class ParseXMLWithJAXB<T> implements Function<HttpResponse, T> {
 
    public <V> V apply(final InputStream stream, final Class<V> type) throws IOException {
       try {
-         return xml.fromXML(Strings2.toStringAndClose(stream), type);
+         String str = Strings2.toStringAndClose(stream);
+         return xml.fromXML(str, type);
       } finally {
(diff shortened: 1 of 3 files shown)


nacx:
LGTM so far. Is there any existing test that verifies the behavior of this class? If not, could we add some of them to make sure we don't introduce regressions?

andrewgaul:
Added test.

neykov:
The change was surprising to me, I expected Java to handle the BOM. After digging deeper I found [[1]](https://stackoverflow.com/questions/4897876/reading-utf-8-bom-marker) which points to two JDK bugs [[2]](http://bugs.java.com/view_bug.do?bug_id=4508058) and [[3]](http://bugs.java.com/view_bug.do?bug_id=6378911). Turns out they fixed it at some point to consume the BOM but then reverted because it breaks backwards compatibility.

Also of interest, the UTF character `0xFEFF` is serialized as `EF BB BF` in the UTF-8 byte sequence [[4]](http://www.unicode.org/faq/utf_bom.html#BOM)

\[1\] https://stackoverflow.com/questions/4897876/reading-utf-8-bom-marker
\[2\] http://bugs.java.com/view_bug.do?bug_id=4508058
\[3\] http://bugs.java.com/view_bug.do?bug_id=6378911
\[4\] http://www.unicode.org/faq/utf_bom.html#BOM

andrewgaul:
Thanks for the background @neykov!

PR #1126, opened 2017-08-03
Handle HTTP 429 in google-cloud-storage

andrewgaul:
This addresses `rateLimitExceeded` errors encountered during integration
tests.

### DIFF
--- a/providers/google-cloud-storage/src/main/java/org/jclouds/googlecloudstorage/handlers/GoogleCloudStorageClientErrorRetryHandler.java
+++ b/providers/google-cloud-storage/src/main/java/org/jclouds/googlecloudstorage/handlers/GoogleCloudStorageClientErrorRetryHandler.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.googlecloudstorage.handlers;
+
+import javax.inject.Singleton;
+
+import org.jclouds.http.HttpCommand;
+import org.jclouds.http.HttpResponse;
+import org.jclouds.http.handlers.BackoffLimitedRetryHandler;
+import org.jclouds.http.HttpRetryHandler;
+
+import com.google.inject.Inject;
+
+@Singleton
+public final class GoogleCloudStorageClientErrorRetryHandler implements HttpRetryHandler {
+   /** The user has sent too many requests in a given amount of time ("rate limiting"). */
+   // TODO: remove when upgrading to jax-rs api 2.1
+   private static final int TOO_MANY_REQUESTS = 429;
+
+   private final BackoffLimitedRetryHandler backoffHandler;
+
+   @Inject
+   protected GoogleCloudStorageClientErrorRetryHandler(BackoffLimitedRetryHandler backoffHandler) {
+      this.backoffHandler = backoffHandler;
+   }
+
+   @Override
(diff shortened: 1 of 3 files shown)


andrewgaul:
Handling HTTP 429 seems like core should handle this.

nacx:
> Handling HTTP 429 seems like core should handle this.

We already have the [RateLimitRetryHandler](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/http/handlers/RateLimitRetryHandler.java) in core. It is a base class to be extended by retry handlers that deal with the provider-specific rate limits. You have a couple example implementations for [digitalocean2](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/handlers/DigitalOcean2RateLimitRetryHandler.java) and [azurecompute-arm](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/handlers/AzureRateLimitRetryHandler.java). Usually these rate limit handlers are configured in an [optional module](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/config/DigitalOcean2RateLimitModule.java) to let users include it only if they want jclouds to automatically throttle the requests upon rate limit errors.

providers/google-cloud-storage/src/main/java/org/jclouds/googlecloudstorage/handlers/GoogleCloudStorageClientErrorRetryHandler.java
```diff
@@ -0,0 +1,45 @@
...
+
+   @Inject
+   protected GoogleCloudStorageClientErrorRetryHandler(BackoffLimitedRetryHandler backoffHandler) {
+      this.backoffHandler = backoffHandler;
+   }
+
+   @Override
+   public boolean shouldRetryRequest(HttpCommand command, HttpResponse response) {
+      if (response.getStatusCode() == 429) {
+         return true;
```
nacx: According to [the best practices](https://cloud.google.com/storage/docs/request-rate#best_practices), we should use a backoff if there is a 429 error. However, returning true here will retry the request immediately. Shouldn't it apply the backoff for 429 errors and return false otherwise?

andrewgaul: Agree that this is backwards/wrong.

andrewgaul:
Updated the pull request to use `RateLimitRetryHandler`.  Thanks for the pointer!  I don't understand why these modules should be optional -- most users would want this behavior.

nacx:
>I don't understand why these modules should be optional -- most users would want this behavior.

I think I kept it optional in DO to keep backward compatibility, initially. In Azure ARM, for example, it is included by default.

The point is that if the module is included in the metadata there is no trivial way to disable this functionality. Users may prefer to fail fast, instead of having a system that takes a very long time to complete a request, so I opted to, in general, provide a separate module users can include where needed.

nacx:
Redargind the implementation, returning an absent will cause the request not to be retried. I think extending the rate limit handler does not make sense here, since we don't want to throttle the requests in a particular way. We just want to apply an exponential backoff if (and only if) the response code is 429.

I'd change the implementation class to directly extend the `BackoffLimitedRetryHandler`, and then just override the `shouldRetryRequest` method as follows:
```java
@Override
public boolean shouldRetryRequest(HttpCommand command, HttpResponse response) {
   return response.getStatusCode() == 429 && super.shouldRetryRequest(command, response);
}
```
This way 429 errors will be retried using the exponential backoff, but no other 4xx error will be retried by defaut.

andrewgaul:
Switched back to `HttpRetryHandler` with corrected logic.

providers/google-cloud-storage/src/main/java/org/jclouds/googlecloudstorage/handlers/GoogleCloudStorageClientErrorRetryHandler.java
```diff
@@ -0,0 +1,45 @@
...
+   private final BackoffLimitedRetryHandler backoffHandler;
+
+   @Inject
+   protected GoogleCloudStorageClientErrorRetryHandler(BackoffLimitedRetryHandler backoffHandler) {
+      this.backoffHandler = backoffHandler;
+   }
+
+   @Override
+   public boolean shouldRetryRequest(HttpCommand command, HttpResponse response) {
+      if (response.getStatusCode() == 429) {
```
demobox: Would it make sense to add a link to documentation explaining what code 429 means?

andrewgaul: Done.

PR #1127, opened 2017-08-03
Fix for Java 1.8 and Guice 4.1

wltjr:
core/src/main/java/org/jclouds/rest/internal/TransformerForRequest.java:
Fix for Java 1.8 and Guice 4.1. Not sure if this is correct, but does
compile. Committing to see if it passes tests to confirm fix or not.

Modified per @nacx on jclouds PR1036
https://github.com/jclouds/jclouds/pull/1036#issuecomment-319904820

Please see other PR for issues, etc. Feel free to edit commit message per jclouds formatting. Its an old style I was using :)


nacx:
Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/6de7974a) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/78881d2f). Many thanks @wltjr!

wltjr:
Sure thing, thanks for the suggestion. I am more a fan of short hand, not having variables when not needed. This was a good learning lesson. It seems when using generics, sometimes that is bad, as it does not pass along the correct generic type. First time I have seen such, but good to be aware in case I run into it again :)

PR #1129, opened 2017-08-08
JCLOUDS-769: streaming putBlob

gaul:
This allows clients to write their payloads via an OutputStream
instead of providing an InputStream during construction.  This can
avoid buffering in some situations although one must still provide the
Content-Length.

### DIFF
--- a/apis/s3/src/main/java/org/jclouds/s3/S3Client.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/S3Client.java
@@ -43,7 +43,9 @@ import org.jclouds.blobstore.BlobStoreFallbacks.ThrowContainerNotFoundOn404;
 import org.jclouds.blobstore.BlobStoreFallbacks.ThrowKeyNotFoundOn404;
 import org.jclouds.blobstore.attr.BlobScope;
 import org.jclouds.http.functions.ParseETagHeader;
+import org.jclouds.http.functions.ReturnOutputStream;
 import org.jclouds.http.options.GetOptions;
+import org.jclouds.io.ETagOutputStream;
 import org.jclouds.io.Payload;
 import org.jclouds.javax.annotation.Nullable;
 import org.jclouds.rest.annotations.BinderParam;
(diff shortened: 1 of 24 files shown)


andrewgaul:
This demonstrates one approach to allowing streaming PUTs of data.  This is useful when clients want to drive the interaction themselves, e.g., interacting with async frameworks, constructing/transforming input on the fly.  This functionality maps poorly onto the existing request-response model.  Presently I fake a bunch of the response data to show what I want to do.  Any better suggestions?

Do not commit; straw man to provoke feedback.

andrewgaul:
I have started to implement the other providers; if you have feedback on the approach I would appreciate it sooner rather than later.

nacx:
I've finally had some time to look at this. Thanks for starting this proposal!

I like the idea of having the API methods return a stream clients can write to, but I'm not sure about several things that are needed to have that.

It's weird for me to have an OutputStream in the HttpResponse. If I get it right it's there just to be accessible from the response transformers, a convenient way to propagate it to the API return value. But it looks weird; I understood it better if it was in the request. We should also take into account that, although not so widely used, there is a generic Http utils class that allows using the HttpRequest and HttpResponse directly, and there are also binders and transformers for those objects. If we modify the request/response objects we need to make sure the features we already leverage that use them directly can still be used with the new additions.

This said, could an alternate approach be payload-based? I mean, instead of changing the return type of a method, could we play with specialized Payloads instead? For example, with the [OkHttp driver we're already streaming data](https://github.com/jclouds/jclouds/blob/master/drivers/okhttp/src/main/java/org/jclouds/http/okhttp/OkHttpCommandExecutorService.java#L120-L146). Could we modify that to let clients provide a specialized payload that leveraged the streaming PUT? I think a PoC with this driver and specific payloads would also be great.

stevegomez17:
thanks

gaul:
@nacx Sorry for my delayed response; this recently became interesting again.  I follow that OkHttp allows returning `BufferedSink.outputStream` but I have a problem at the jclouds `HttpRequest` and `HttpResponse` abstraction.  My callers generate dynamic content and want to interact with the underlying `OutputStream` which the current system does not provide.  I agree that abusing `HttpResponse` for something that is not actually a response is an ugly hack and did something slightly more palatable with `Object`.  Reviewing my own code, `EtagOutputStream` was stupid and should be `HttpRequestOutputStream` whose `close` method returns the `HttpResponse`.

nacx:
Hmmmmm, I see where you're going.

The main issue I see with this approach is that a client won't know actually when a streaming request has finished, and won't have access to the response code, which might be important...
How are we handling streaming requests to endpoints return a redirect (3xx) we should follow before posting the content?

Access to the response status is important, and that was one of my reasons to suggest the specialized payload approach. My thinking was something like having a specialised binding for, say, `Writer` parameters in API methods, and handle that at the writer level by blocking until the writer is done, then completing the method call and returning following the existing flows and response and error handling mechanisms.

Something like:

```java
@Named("PutObject")
@PUT
@Path("/{key}")
@Headers(keys = EXPECT, values = "100-continue")
@ResponseParser(ReturnOutputStream.class)
ETagOutputStream putObjectStreaming(
      @Bucket @EndpointParam(parser = AssignCorrectHostnameForBucket.class) @BinderParam(BindAsHostPrefixIfConfigured.class) @ParamValidators(BucketNameValidator.class) String bucketName,
      @PathParam("key") @ParamParser(ObjectKey.class) @BinderParam(BindS3ObjectMetadataToRequest.class) S3Object object,
      Writer streamingContent,   // This is an object the client controlls that should be piped to the request outputstream
      PutObjectOptions... options);
```

Does this make sense?

gaul:

PR #1132, opened 2017-08-25
Upgrade jax-rs and animal-sniffer

andrewgaul:
jax-rs 2.1 adds a new status code which addresses a TODO.  animal-snipper upgrade needed to upgrade jax-rs.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -47,7 +47,7 @@
     <dependency>
       <groupId>javax.ws.rs</groupId>
       <artifactId>javax.ws.rs-api</artifactId>
-      <version>2.0.1</version>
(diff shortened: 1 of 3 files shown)


andreaturli:
@andrewgaul do you think we should update also https://github.com/jclouds/jclouds-karaf/blob/master/pom.xml#L199 if this PR gets merged?

Mind adding a description for this to know the rationale behind it? Thanks!

andrewgaul:
Huh, it looks like the new jax-rs requires Java 7:

```
java.lang.UnsupportedClassVersionError: javax/ws/rs/GET : Unsupported major.minor version 52.0
```

PR #1133, opened 2017-08-26
JCLOUDS-1334: Guava 23 compatibility (SimpleTimeLimiter creation)

Tembrel:
Added helper method that uses reflection (initialized statically) to call `SimpleTimeLimiter.create(ExecutorService)` if possible, otherwise `new SimpleTimeLimiter(ExecutorService)`.


core/src/main/java/org/jclouds/concurrent/config/ExecutorServiceModule.java
```diff
@@ -113,7 +153,7 @@ protected void configure() { // NO_UCD
    @Provides
    @Singleton
    final TimeLimiter timeLimiter(@Named(PROPERTY_USER_THREADS) ListeningExecutorService userExecutor) {
-      return new SimpleTimeLimiter(userExecutor);
+      return createSimpleTimeLimiter(userExecutor);
```
andrewgaul: Could you also make changes to `BaseRestApiExpectTest` and `BaseRestApiTest.java` for consistency?

core/src/main/java/org/jclouds/concurrent/config/ExecutorServiceModule.java
```diff
@@ -54,6 +58,42 @@
 @ConfiguresExecutorService
 public class ExecutorServiceModule extends AbstractModule {
 
+   private static final Method CREATE_STL;
```
andrewgaul: Please add a comment explaining that this code works around Guava 18 and 23 compatibility.

core/src/main/java/org/jclouds/concurrent/config/ExecutorServiceModule.java
```diff
@@ -54,6 +58,42 @@
...
+         } else if (CONSTRUCT_STL != null) {
+            return CONSTRUCT_STL.newInstance(executorService);
+         }
+         // fall through
+      } catch (IllegalAccessException iae) {
+         // fall through
+      } catch (InstantiationException ie) {
+         // fall through
+      } catch (InvocationTargetException ite) {
+         // fall through
```
andrewgaul: It would be better to propagate the cause of these exceptions.

core/src/main/java/org/jclouds/concurrent/config/ExecutorServiceModule.java
Tembrel: Done

core/src/main/java/org/jclouds/concurrent/config/ExecutorServiceModule.java
Tembrel: Done

core/src/main/java/org/jclouds/concurrent/config/ExecutorServiceModule.java
Tembrel: Done

core/src/main/java/org/jclouds/concurrent/config/ExecutorServiceModule.java
```diff
@@ -54,6 +58,51 @@
...
+         } else if (CONSTRUCT_STL != null) {
+            return CONSTRUCT_STL.newInstance(executorService);
+         }
+         // fall through
+      } catch (IllegalAccessException iae) {
+         // fall through
+      } catch (InstantiationException ie) {
+         // fall through
+      } catch (InvocationTargetException ite) {
+          throw Throwables.propagate(ite.getCause());
```
andrewgaul: This is not quite right -- you want to propagate the actual exception, not its cause.  Also this does not propagate the first two exceptions.  Finally ou can just `throw RuntimeException`; Guava deprecated `Throwables.propagate`.

core/src/test/java/org/jclouds/rest/internal/BaseRestApiExpectTest.java
```diff
@@ -92,22 +92,22 @@
 import com.google.inject.TypeLiteral;
 
 /**
- * 
+ *
```
andrewgaul: Can you remove these spurious changes?

Tembrel: I think so, but I'll wait until I get a response on the exception handling point before making those changes.

core/src/main/java/org/jclouds/concurrent/config/ExecutorServiceModule.java
Tembrel: Earlier you wrote "It would be better to propagate the cause of these exceptions", but the only one of the exceptions in the three catch clauses that has a (relevant) cause is `InvocationTargetException`. Maybe you meant "propagate these exceptions to indicate the cause of the failure"?

Freed from that misunderstanding, here's my analysis. There are four failure cases:

1. Neither `SimpleTypeLimiter.create(ExecutorService)` nor `SimpleTypeLimiter(ExecutorService)` was found. Should never happen. but if it did, the `NoSuchMethodException` that gave rise to this state has been lost by this point.
1. `IllegalAccessException`: don't have access to the method or constructor.
1. `InstantiationException`: constructor invocation failed. 
1. `InvocationTargetException`: the method or constructor invocation threw an exception.

In the first case, there's no exception to propagate, so `UnsupportedMethodException` is a pretty good description of the situation. In the second and third cases, these checked exceptions could be wrapped in an unchecked exception and rethrown.

In the last case, the cause of the `InvocationTargetException` is the main point. The `createSimpleTimeLimiter` method is a substitute for either a call to `SimpleTimeLimiter.create(exec)` or `new SimpleTimeLimiter(exec)`, so we'd like its semantics to mirror those of the things it's replacing as closely as possible. I think it would better to unwrap the cause and then wrap in an unchecked exception if necessary before rethrowing.

I stopped using `Throwables.propagate` when Guava deprecated it, but JClouds' `TypeTokenUtils`, on which I modeled this PR, **does** use `Throwables.propagate`. But I'm happy to wrap with an unchecked exception instead when necessary.

However, I'm not convinced that `RuntimeException` is really the best choice for wrapping checked exceptions. For one thing, my linting tools frown on using such a generic exception type if there's a more subtype that's more specifically appropriate. In this case, I think `UnsupportedOperationException` is a better choice for all of the failure cases except when the wrapped invocation target exception is already unchecked.

Tembrel: I think the right thing to throw in the static initialization block, instead of `Throwables.propagate` is `NoSuchMethodError`, instead of wrapping `NoSuchMethodException` with an unchecked exception.

Tembrel: I'll go ahead and commit code consistent with what I've written above without waiting for a response, which might save a round-trip.

core/src/test/java/org/jclouds/rest/internal/BaseRestApiExpectTest.java
Tembrel: As I wrote above, instead I'm going to commit now so you have something concrete to respond to.

Tembrel: Done

andrewgaul:
Squashed and pushed to master as e08acc6ed6b2761b6063f9ae105ee9f4077ef418.  Thank you for your contribution @Tembrel!

PR #1135, opened 2017-08-28
[JCLOUDS-1332] destroyNode and destroyNodesMatchingPredicate

andreaturli:
Modify `BaseComputeService` to have similar beahvior

/cc @nacx


andreaturli:
rebuild please

nacx:
Changes to the logic LGTM.
We need to check if there are providers that are overriding this behavior and see if we need to change logic there too.

andreaturli:
I think the GCE failures are related to this PR: GCE is one of the provider that overrides that logic and possibly requires some refactoring. 

Thanks @nacx for suggestions, as you consider the patch good I'll invest some more time trying to fix the broken tests

andreaturli:
@nacx could you have another look at it, please? thanks!

nacx:
Worth running the GCE live tests? The same methods should be removed from Azure ARM.

andreaturli:
ok merging now @nacx I will run gce live tests immediately after the merge

andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/af4bd3a8) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/7785379d)

andreaturli:
@nacx gce live tests at https://jclouds.ci.cloudbees.com/view/live%20tests/job/jclouds-live-tests-google-compute-engine/127/ don't look good. 
There are some failures unrelated to this PR, possibly the previous aborted run (126) is creating conflicts. I'll re-run them when `delete-all` task will run

PR #1142, opened 2017-09-07
JCLOUDS-1258: Allow China endpoints in FormSignerV4

nacx:
The creation of a compute context for a China region failed because the `FormSignerV4` didn't allow China endpoints.

The error happens at constructor injection time when Guice is building all singleton objects. The `IllegalArgumentException` thrown there causes an infinite loop and ends up in an `OutOfMemoryexception`.

@andrewgaul @demobox Despite having only added the China suffix in this commit, I'd rather remove completely this validation, as it provides very little value. If the endpoint does not support V4 signatures, users will probably see a self-explanatory error response coming back. WDYT?


nacx:
Done. Thanks!

nacx:
Squashed and merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/1c1f81c3) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/cfdbc810).

PR #1144, opened 2017-09-07
Fixes for jclouds openstack swift tests.

timuralp:
The ContainerApiLiveTest class has tests that mostly assume that a
single test container always exists. They collide on changes to the
state of the container. Some of the tests, though, assume the
container does _not_ exist and attempt to create it and subsequently
delete it. The change in this patch makes their behavior uniform with
respect to that test container: assume it exists and if the test tries
to delete it or mutate it, delete the container and create it at the
end of the test.

The fix in ObjectApi is for the Object GET requests. Currently,
jclouds passes "format=json", which is not a supported query parameter
for object GET.

Lastly, ObjectApiLiveTest that test server-side copy should ignore the
X-Openstack-Request-ID header, as its value will change between the
two requests.

### DIFF
--- a/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/ObjectApi.java
+++ b/apis/openstack-swift/src/main/java/org/jclouds/openstack/swift/v1/features/ObjectApi.java
@@ -172,7 +172,6 @@ public interface ObjectApi {
    @Path("/{objectName}")
    @ResponseParser(ParseObjectFromResponse.class)
    @Fallback(NullOnNotFoundOr404.class)
-   @QueryParams(keys = "format", values = "json")
    @Nullable
    SwiftObject get(@PathParam("objectName") String objectName);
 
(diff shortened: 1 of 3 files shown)


andrewgaul:
Successfully tested against rackspace-cloudfiles-us and verified that this commit reduces openstack-swift failures from:

```
Failed tests:
  TemporaryUrlSignerLiveTest.signForPublicAccess:55 » IO Server returned HTTP re...
  ContainerApiLiveTest.testDeleteMetadata:178 » ContainerNotFound ContainerApiLi...
  ContainerApiLiveTest.testGet:156 NullPointer
  ContainerApiLiveTest.testListWithOptions:106 » IllegalState Optional.get() can...
  RegionScopedSwiftBlobStoreParallelLiveTest.setup:96->getBlobStore:84 » IllegalArgument
  ContainerApiLiveTest.testUpdateMetadata:167 » ContainerNotFound ContainerApiLi...
  ObjectApiLiveTest.testCopyObjectWithMetadata:217 Could not find: X-Openstack-Request-Id=tx8979c3b6252d4ed1818a8-0059b227fb expected [true] but found [false]

Tests run: 156, Failures: 7, Errors: 0, Skipped: 23
```

to:

```
Failed tests:
  RegionScopedSwiftBlobStoreParallelLiveTest.setup:96->getBlobStore:84 » IllegalArgument

Tests run: 156, Failures: 1, Errors: 0, Skipped: 23
```

PR #1145, opened 2017-09-12
JCLOUDS-1340: Disable BackoffLimitedRetryHandlerTest

andrewgaul:
These often fail in our continuous integration tests.


andrewgaul:
I successfully ran `mvn test -pl core` 300 times with this commit.  Previously it would fail every 10-20 times.

PR #1147, opened 2017-09-21
JCLOUDS-1335: Azure Blob object access tiers

andrewgaul:


andrewgaul:
The new tests will fail until the tiering goes GA or if you opt into the preview:

https://azure.microsoft.com/en-us/blog/announcing-the-public-preview-of-azure-archive-blob-storage-and-blob-level-tiering/

PR #1150, opened 2017-10-23
Exclude tier tests on OSX

nacx:


apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
```diff
@@ -89,6 +91,15 @@ protected void checkUserMetadata(Map<String, String> userMetadata1, Map<String,
       }
    }
 
+   /* Java on OS X does not support extended attributes, which the filesystem backend
+    * uses to implement storage tiers */
+   @Override
+   protected void testPutBlobTierHelper(Tier tier, PutOptions options) throws Exception {
+      if (!isMacOSX()) {
```
gaul: Instead could you flag this as skipped via:

```java
if (isMacOSX()) {
    throw new SkipException("Extended attributes not supported on Mac OS X");
}
```

jclouds-mirror:
I followed the same approach than the other methods in this class. Should
we better change them all to throw skip exceptions?

gaul:
@nacx Agreed that this is inconsistent.  I think it is better to throw `SkipException` since this flags tests in an obvious way, although several tests use `dataProvider` or conditional logic to avoid running some tests.  If you do change this, note that methods like `validateMetadata` are not tests and need to conditionally skip Mac OS X so that other tests methods still run.

nacx:
> I think it is better to throw SkipException

Done!

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
```diff
@@ -82,26 +79,34 @@ public void testCreateBlobWithExpiry() throws InterruptedException {
...
+      checkExtendedAttributesSupport();
+      super.testPutBlobTierHelper(tier, options);
    }
 
    @Override
    public void testSetBlobAccess() throws Exception {
       throw new SkipException("filesystem does not support anonymous access");
    }
+
+   private void checkExtendedAttributesSupport() {
```
demobox: [minor] Does this need to be an instance method, or could this be static?

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
```diff
@@ -82,26 +79,34 @@ public void testCreateBlobWithExpiry() throws InterruptedException {
...
+      if (isMacOSX()) {
+         throw new SkipException("filesystem does not support extended attributes on Mac OSX");
```
demobox: If we end up using `SkipException` in lots of places, thoughts on creating a `TestUtils.skip` helper for readability? Then we could write:

```
if (isMacOSX()) {
   skip("filesystem does not support extended attributes on Mac OSX");
}
```
?

Thanks for taking care of this, @nacx!

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
```diff
@@ -49,27 +51,24 @@ protected Properties setupProperties() {
...
    }
 
    // Mac OS X HFS+ does not support UserDefinedFileAttributeView:
    // https://bugs.openjdk.java.net/browse/JDK-8030048
    @Override
    protected void validateMetadata(BlobMetadata metadata) throws IOException {
-      if (!isMacOSX()) {
-         super.validateMetadata(metadata);
-      }
+      checkExtendedAttributesSupport();
```
gaul: This is what I mentioned in my previous review -- this will cause a test which checks many metadata, including non-xattr, to skip.  We should only skip full tests.

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
```diff
@@ -80,17 +79,34 @@ public void testCreateBlobWithExpiry() throws InterruptedException {
...
+   }
+
+   /*
+    * Java on OS X does not support extended attributes, which the filesystem
+    * backend uses to implement storage tiers
+    */
+   @Override
+   protected void testPutBlobTierHelper(Tier tier, PutOptions options) throws Exception {
+      checkExtendedAttributesSupport();
+      super.testPutBlobTierHelper(tier, options);
```
gaul: This correctly skips.

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
gaul: I don't think this adds anything an prevents `grep` from finding tests we skip, part of the expanded scope of this pull request.

nacx:
Ok, I've cleaned up the commits and the code, leaving just the proper test skip, as the other methods were already ok.

nacx:
Merged to master as [f7b74d95](http://git-wip-us.apache.org/repos/asf/jclouds/commit/f7b74d95).

PR #1152, opened 2017-10-25
Define the JDK version in a Maven property

nacx:
This changes how we reference the toolchain to use a Maven property. This will allow us to be more flexible and generic when configuring builds in Jenkins and we can avoid having a dedicated maven profile for each JDK version we want to support. A property can be overridden in the command-line, so it looks more convenient to configure it this way.


nacx:
If this works, the profile must also be defined in `jclouds-karaf` and `jclouds-cli`. I forgot those projects when merging the first PR.

nacx:
Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/6f7e4b92) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a03e047d).

project/pom.xml
```diff
@@ -1393,35 +1394,7 @@
...
-                  <goal>toolchain</goal>
-                </goals>
-              </execution>
-            </executions>
-            <configuration>
-              <toolchains>
-                <jdk>
-                  <version>1.8</version>
-                  <vendor>openjdk</vendor>
+                  <version>${jdk.version}</version>
```
demobox: @nacx Does removing the `oracle` vs. `openjdk` `<vendor>` attribute have any impact?

nacx: No. We were not building against different JDK flavors, so it does not have an impact. If we want to do so, we can introduce the vendor again and revisit this. The property approach comes in handy for a clearer [jclouds-compat build](https://jclouds.ci.cloudbees.com/job/jclouds-compat/).

PR #1153, opened 2017-11-02
JCLOUDS-1351: fixes to correctly recognise and boot the official CentOS images from the AWS Marketplace

rdowner:
CentOS’ officially-supported AMIs are hosted on the AWS Marketplace. This adds support for those images, recognising the AMI naming convention and ensuring the OS metadata is parsed correctly and the correct SSH login name is used.

JCLOUDS-1351 is fixed, which was a blocker to using these AMIs. Other issues have been resolved so the jclouds detects the correct data needed to populate the node metadata and allow jclouds to connect.

There is no change to the default jclouds configuration and the official CentOS images will not be detected by default. To use these images, you must alter the ami-query properties to include searching the “AWS Marketplace”, which has an owner ID of 679593333241. You must also manually log on to the AWS Marketplace, select your chosen CentOS image, and “subscribe” to it (you can do this by proceeding as if to launch an image, but stopping after you have agreed to the subscription and before launching).


rdowner:
Thanks @nacx. There's a chance that a user could be caught out by a change in jclouds behaviour, but these changes should *increase* the accuracy of the metadata (e.g. reporting a specific Linux distribution than simply "Linux"). Inspecting the code, I can't see a case where it would ever decrease the accuracy. Therefore I think it's safe to backport to 2.0.x.

nacx:
Merged to 2.0.x. Thanks again!

PR #1156, opened 2017-11-13
[JCLOUDS-1354] avoid memory leak by reusing BouncyCastleProvider

thomasgl-orange:
See [JCLOUDS-1354](https://issues.apache.org/jira/browse/JCLOUDS-1354).  Reuse a single instance of `BouncyCastleProvider`, because every instance used at least once will be retained forever in the JCE providers map (in `javax.crypto.JceSecurity` - see also [JDK-8168469](https://bugs.openjdk.java.net/browse/JDK-8168469)).

### DIFF
--- a/drivers/bouncycastle/src/main/java/org/jclouds/encryption/bouncycastle/BouncyCastleCrypto.java
+++ b/drivers/bouncycastle/src/main/java/org/jclouds/encryption/bouncycastle/BouncyCastleCrypto.java
@@ -17,6 +17,8 @@
 package org.jclouds.encryption.bouncycastle;
 
 import java.security.NoSuchAlgorithmException;
+import java.security.Provider;
+import java.security.Security;
 import java.security.cert.CertificateException;
 
 import javax.crypto.Cipher;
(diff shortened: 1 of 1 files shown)


nacx:
Thanks for the detailed bug report and for the PR @thomasgl-orange!

Although I fully understand the memory leak, I'm a bit concerned about introducing a cross-context instance. If at some point the JDK bug gets fixed, users might expect all context-scoped resources to be released when the context is closed, and having a static reference by default there might not be expected.

Instead of preconfiguring a shared instance, what about modifying the [BouncyCastleCryptoModule](https://github.com/jclouds/jclouds/blob/master/drivers/bouncycastle/src/main/java/org/jclouds/encryption/bouncycastle/config/BouncyCastleCryptoModule.java) to provide an alternate constructor that gets a `BouncyCastleProvider`, and have that one (or a per-context default one, to keep the current behavior) injected? This way users could keep using the current behavior or instantiate the provider themselves (by passing a reference to a static instance). This is also the approach we already use in the [CredentialStoreModule](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/rest/config/CredentialStoreModule.java) for users that want credential persistence across contexts.

gaul:

PR #1159, opened 2017-11-22
[JCLOUDS-1360] add support for block device mappings to Nova

andreaturli:


apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeServiceAdapter.java
```diff
@@ -131,7 +131,8 @@ public NovaComputeServiceAdapter(NovaApi novaApi, @Region Supplier<Set<String>>
       if (templateOptions.getKeyPairName() != null) {
          options.keyPairName(templateOptions.getKeyPairName());
       }
-      
+      if (!templateOptions.getBlockDeviceMappings().isEmpty()) options.blockDeviceMappings(templateOptions.getBlockDeviceMappings());
```
nacx: No need to check for empty.

andreaturli:
thanks @nacx 

Merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/ec03b710)

PR #1162, opened 2017-11-29
Fix an unexpected test failure in Windows path handling

demobox:
See http://markmail.org/message/tahusme6svuhpew6.

From a quick local test, it seems that this test has been failing on Windows since it was added in 6452960c72a5f8, so I'm not sure if the test input was incorrectly constructed at the time, or if the actual functionality has been broken on Windows from the beginning.

@gaul Any light you could shed on that?

### DIFF
--- a/apis/filesystem/src/test/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImplTest.java
+++ b/apis/filesystem/src/test/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImplTest.java
@@ -686,9 +686,21 @@ public class FilesystemStorageStrategyImplTest {
       }
    }
 
-   @Test
-   public void testDeletingInvalidPathFileEndsNormally() {
+   @DataProvider
+   public Object[][] getInvalidPathBlobKey() {
       String invalidPathBlobKey = "A<!:!@#$%^&*?]8 /\0";
+      // the JDK can't handle '/' in Windows file paths
+      if (TestUtils.isWindowsOs()) {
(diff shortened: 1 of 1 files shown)


apis/filesystem/src/test/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImplTest.java
```diff
@@ -686,9 +686,21 @@ public void testPutIncorrectContentLength() throws Exception {
       }
    }
 
-   @Test
-   public void testDeletingInvalidPathFileEndsNormally() {
+   @DataProvider
+   public Object[][] getInvalidPathBlobKey() {
       String invalidPathBlobKey = "A<!:!@#$%^&*?]8 /\0";
+      // the JDK can't handle '/' in Windows file paths
```
gaul: `FilesystemStorageStrategyImpl.normalize` should handle this translation.  I wonder why it does not?

demobox: Aaaaah...so what seems to be happening here is that removing `\` means that `parentPath` [here](https://github.com/jclouds/jclouds/blob/master/apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java#L846) is then empty, so the code that throws the error (`directory.toPath()`, a few lines down) is never actually called.

@gaul Do you recall what the intention is here? Is `invalidPathBlobKey` supposed to represent a blob key **without** path information, or is it supposed to represent a path, too?

gaul:

PR #1164, opened 2017-11-30
Fix case sensitivity of blobstore metadata

yuvalk:
HTTP headers are case insensitive by nature (see RFC 2616). When addUserMetadataTo check if this is indeed a user metadata header, it must properly ignore case.
The fix make sure that both metadataPrefix and the header key are compared with toLowerCase().

This solves issue with minio metadata read


stevegomez17:
Thanks for the info.   
  
  
  

  
  

  
  
>

nacx:
Thanks, @yuvalk! Could you file an issue in our [JIRA](https://issues.apache.org/jira/projects/JCLOUDS) so we can properly track this?

yuvalk:
https://issues.apache.org/jira/browse/JCLOUDS-1363

nacx:
Merged to master as [17cef565](http://git-wip-us.apache.org/repos/asf/jclouds/commit/17cef565). Thanks, @yuvalk!
@gaul given this slightly changes the behavior, should we cherry-pick it to 2.0.x?

PR #1166, opened 2018-01-04
JCLOUDS-1368: Correct use of slicing algorithm

gaul:
`MultipartUploadSlicingAlgorithm` creates multiple equal-sized parts and
a remaining amount.  `BaseBlobStore.putMultipartBlob` used this
interface incorrectly, which could create more parts than intended
since the remaining size could be larger than the part size.  This
manifested with Google Cloud Storage which only allows 32 parts.


gaul:
Sigh, I'll look at this tomorrow.

gaul:
I reworked this commit to correct the use of the slicing algorithm.  [JCLOUDS-1369](https://issues.apache.org/jira/browse/JCLOUDS-1369) tracks improving the algorithm itself.

PR #1167, opened 2018-01-05
JCLOUDS-1370: Add CannedAccessPolicy constants

gaul:
Also use CaseFormat instead of extra logic.

### DIFF
--- a/apis/s3/src/main/java/org/jclouds/s3/domain/CannedAccessPolicy.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/domain/CannedAccessPolicy.java
@@ -16,6 +16,8 @@
  */
 package org.jclouds.s3.domain;
 
+import com.google.common.base.CaseFormat;
+
 /**
  * Description from Amazon's documentation:
  * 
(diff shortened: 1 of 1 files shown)


apis/s3/src/main/java/org/jclouds/s3/domain/CannedAccessPolicy.java
```diff
@@ -36,29 +38,50 @@
     /**
      * Owner gets FULL_CONTROL. No one else has access rights (default).
      */
-    PRIVATE("private"),
```
stevegomez17: “PRIVATE”

apis/s3/src/main/java/org/jclouds/s3/domain/CannedAccessPolicy.java
```diff
@@ -36,29 +38,50 @@
...
+    PRIVATE,
     /**
      * Owner gets FULL_CONTROL and the anonymous identity is granted READ
      * access. If this policy is used on an object, it can be read from a
      * browser with no authentication.
      */
-    PUBLIC_READ("public-read"),
```
stevegomez17: “PUBLIC _READ”

apis/s3/src/main/java/org/jclouds/s3/domain/CannedAccessPolicy.java
```diff
@@ -36,29 +38,50 @@
...
+    PUBLIC_READ,
     /**
      * Owner gets FULL_CONTROL, the anonymous identity is granted READ and
      * WRITE access. This can be a useful policy to apply to a bucket, but is
      * generally not recommended.
      */
```
stevegomez17: PUBLIC_READ

gaul: I don't understand any of your comments.  Can you use complete sentences?

PR #1168, opened 2018-01-08
Handle "file" and "file/" collision

gaul:
Fixes gaul/s3proxy#240.


gaul:
@kahing Please review.

kahing:
what happens when I put key/ and get(key)?

gaul:
> what happens when I put key/ and get(key)?

Added a test for this behavior.

PR #1173, opened 2018-01-17
Promote OpenStack Neutron

nacx:
The OpenStack neutron API has been stable for a long time and it's been used in production for more than a year. It's time to promote it to the main repo, as this will allow us to implement some cool stuff to combine it with OpenStack Nova.

/cc @andreaturli


nacx:
Merged to master.

PR #1178, opened 2018-02-15
[JCLOUDS-1382] improve usage of Neutron capabilities from Nova

andreaturli:
- move Neutron FloatingIP API from extension to feature
- introduce FloatingIpForServer as abstract domain object to represents Nova and Neutron objects
- manage floating ip cache and its invalidation

---

cc @nacx

### DIFF
--- a/apis/openstack-neutron/src/test/java/org/jclouds/openstack/neutron/v2/features/FloatingIPApiLiveTest.java
+++ b/apis/openstack-neutron/src/test/java/org/jclouds/openstack/neutron/v2/features/FloatingIPApiLiveTest.java
@@ -15,28 +15,25 @@
  * limitations under the License.
  */
 
-package org.jclouds.openstack.neutron.v2.extensions;
+package org.jclouds.openstack.neutron.v2.features;
 
-import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertTrue;
 
 import java.util.Set;
 
+import com.google.common.base.Predicate;
 import org.jclouds.openstack.neutron.v2.domain.FloatingIP;
 import org.jclouds.openstack.neutron.v2.domain.IP;
 import org.jclouds.openstack.neutron.v2.domain.Network;
-import org.jclouds.openstack.neutron.v2.domain.NetworkType;
-import org.jclouds.openstack.neutron.v2.domain.Subnet;
-import org.jclouds.openstack.neutron.v2.features.NetworkApi;
-import org.jclouds.openstack.neutron.v2.features.SubnetApi;
 import org.jclouds.openstack.neutron.v2.internal.BaseNeutronApiLiveTest;
+import org.testng.Assert;
 import org.testng.annotations.Test;
 
 import com.google.common.collect.ImmutableSet;
 
 /**
- * Tests parsing and Guice wiring of RouterApi
+ * Tests parsing and Guice wiring of FloatingIPApi
  */
 @Test(groups = "live", testName = "FloatingIPApiLiveTest")
 public class FloatingIPApiLiveTest extends BaseNeutronApiLiveTest {
@@ -44,44 +41,32 @@ public class FloatingIPApiLiveTest extends BaseNeutronApiLiveTest {
    public void testCreateUpdateAndDeleteFloatingIP() {
       for (String region : api.getConfiguredRegions()) {
 
-         SubnetApi subnetApi = api.getSubnetApi(region);
-         FloatingIPApi floatingIPApi = api.getFloatingIPApi(region).get();
+         FloatingIPApi floatingIPApi = api.getFloatingIPApi(region);
          NetworkApi networkApi = api.getNetworkApi(region);
 
          FloatingIP floatingIPGet = null;
-         String ipv4SubnetId = null;
-         Network network = null;
+         Network network;
 
          try {
-            network = networkApi.create(
-                  Network.createBuilder("jclouds-network-test").external(true).networkType(NetworkType.LOCAL).build());
-            assertNotNull(network);
+            network = networkApi.list().concat().firstMatch(new Predicate<Network>() {
+               @Override
+               public boolean apply(Network input) {
+                  return input.getExternal();
+               }
+            }).orNull();
 
-            ipv4SubnetId = subnetApi.create(Subnet.createBuilder(network.getId(), "198.51.100.0/24").ipVersion(4)
-                  .name("JClouds-Live-IPv4-Subnet").build()).getId();
+            if (network == null) Assert.fail("Cannot find a suitable external network. Please add it manually or contact your administrator");
 
-            floatingIPApi.create(FloatingIP.createBuilder(network.getId()).build());
-
+            FloatingIP floatingIP = floatingIPApi.create(FloatingIP.createBuilder(network.getId()).availabilityZone(network.getAvailabilityZone()).build());
             /* List and Get test */
             Set<FloatingIP> floatingIPs = floatingIPApi.list().concat().toSet();
-            FloatingIP floatingIPList = floatingIPs.iterator().next();
-            floatingIPGet = floatingIPApi.get(floatingIPList.getId());
+            floatingIPGet = floatingIPApi.get(floatingIP.getId());
 
             assertNotNull(floatingIPGet);
-            assertEquals(floatingIPGet, floatingIPList);
+            assertTrue(floatingIPs.contains(floatingIP));
          }
          finally {
-            try {
-               assertTrue(floatingIPApi.delete(floatingIPGet.getId()));
-            }
-            finally {
-               try {
-                  assertTrue(subnetApi.delete(ipv4SubnetId));
-               }
-               finally {
-                  assertTrue(networkApi.delete(network.getId()));
-               }
-            }
+            assertTrue(floatingIPApi.delete(floatingIPGet.getId()));
          }
       }
    }
--- a/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
+++ b/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
@@ -25,19 +25,29 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 
+import javax.annotation.Nullable;
 import javax.annotation.Resource;
-import javax.inject.Inject;
 import javax.inject.Named;
 
+import com.google.common.collect.Sets;
+import org.jclouds.Context;
 import org.jclouds.compute.domain.NodeMetadata;
 import org.jclouds.compute.domain.NodeMetadataBuilder;
 import org.jclouds.compute.reference.ComputeServiceConstants;
 import org.jclouds.logging.Logger;
+import org.jclouds.openstack.neutron.v2.NeutronApi;
+import org.jclouds.openstack.neutron.v2.domain.Network;
+import org.jclouds.openstack.neutron.v2.domain.Networks;
+import org.jclouds.openstack.neutron.v2.domain.Port;
+import org.jclouds.openstack.neutron.v2.features.NetworkApi;
+import org.jclouds.openstack.neutron.v2.features.PortApi;
 import org.jclouds.openstack.nova.v2_0.NovaApi;
 import org.jclouds.openstack.nova.v2_0.compute.options.NodeAndNovaTemplateOptions;
 import org.jclouds.openstack.nova.v2_0.domain.FloatingIP;
+import org.jclouds.openstack.nova.v2_0.domain.FloatingIpForServer;
 import org.jclouds.openstack.nova.v2_0.domain.regionscoped.RegionAndId;
 import org.jclouds.openstack.nova.v2_0.extensions.FloatingIPApi;
+import org.jclouds.rest.ApiContext;
 import org.jclouds.rest.InsufficientResourcesException;
 import org.jclouds.rest.ResourceNotFoundException;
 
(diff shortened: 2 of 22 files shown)


apis/openstack-neutron/src/test/java/org/jclouds/openstack/neutron/v2/features/FloatingIPApiLiveTest.java
```diff
@@ -15,73 +15,60 @@
...
-            network = networkApi.create(
-                  Network.createBuilder("jclouds-network-test").external(true).networkType(NetworkType.LOCAL).build());
-            assertNotNull(network);
-
-            ipv4SubnetId = subnetApi.create(Subnet.createBuilder(network.getId(), "198.51.100.0/24").ipVersion(4)
-                  .name("JClouds-Live-IPv4-Subnet").build()).getId();
+            network = networkApi.list().concat().firstMatch(new Predicate<Network>() {
+               @Override
+               public boolean apply(@Nullable Network input) {
+                  return input.getExternal();
```
nacx: If input is nullable better guard against a NPE

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
```diff
@@ -156,8 +215,57 @@ public boolean apply(FloatingIP arg0) {
...
+      String regionId = node.getLocation().getParent().getId();
+
+      Optional<Network> networkOptional = getNetworkApi(regionId).list().concat().firstMatch(new Predicate<Network>() {
+         @Override
+         public boolean apply(@Nullable Network input) {
+            return input.getExternal() && input.getAvailabilityZone().equals(availabilityZone);
+         }
+      });
+      if (!networkOptional.isPresent())
+         throw new InsufficientResourcesException("Failed to find a suitable external network.");
```
nacx: To be more close to the old behavior, instead of looking for *any* external network, could we use the existing `poolNames` configuration to filter external networks by name? This way users will configure the same name `where` they want the floating IPs from, regardless of the API they are using.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
```diff
@@ -156,8 +215,57 @@ public boolean apply(FloatingIP arg0) {
...
+
+      org.jclouds.openstack.neutron.v2.domain.FloatingIP createFloatingIP = org.jclouds.openstack.neutron.v2.domain.FloatingIP.CreateFloatingIP
+              .createBuilder(networkOptional.get().getId())
+              .availabilityZone(networkOptional.get().getAvailabilityZone())
+              .build();
+
+      org.jclouds.openstack.neutron.v2.domain.FloatingIP floatingIP = neutronFloatingApi.create((org.jclouds.openstack.neutron.v2.domain.FloatingIP.CreateFloatingIP) createFloatingIP);
+      floatingIpCache.asMap().putIfAbsent(RegionAndId.fromSlashEncoded(node.getId()), ImmutableList.of(FloatingIpForServer.create(RegionAndId.fromSlashEncoded(node.getId()), floatingIP.getId(), floatingIP.getFloatingIpAddress())));
```
nacx: Why put if absent? Shouldn't we override the old value, if any?

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
```diff
@@ -78,37 +94,77 @@ public AllocateAndAddFloatingIpToNode(@Named(TIMEOUT_NODE_RUNNING) Predicate<Ato
...
+               }
+
+               org.jclouds.openstack.neutron.v2.domain.FloatingIP ip = neutronFloatingApi.update(floatingIP.getId(),
+                       org.jclouds.openstack.neutron.v2.domain.FloatingIP.UpdateFloatingIP
+                               .updateBuilder()
+                               .portId(optionalPort.get().getId())
+                               .build());
+
+               input.get().getNodeMetadata().set(NodeMetadataBuilder.fromNodeMetadata(node).publicAddresses(ImmutableSet.of(ip.getFloatingIpAddress())).build());
+            }
```
nacx: If the node has no ports associated (although unlikely) we'd better at least log a warning instead of silently doing nothing.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/NeutronFloatingIpToFloatingIp.java
```diff
@@ -0,0 +1,52 @@
...
+      implements Function<org.jclouds.openstack.neutron.v2.domain.FloatingIP, FloatingIP> {
+
+   public interface Factory {
+      NeutronFloatingIpToFloatingIp create(Location location);
+   }
+
+   private final Location location;
+
+   @Inject
+   public NeutronFloatingIpToFloatingIp(@Assisted Location location) {
```
nacx: Remove the public modifier

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/NeutronFloatingIpToFloatingIp.java
```diff
@@ -0,0 +1,52 @@
...
+      this.location = location;
+   }
+
+   @Override
+   public FloatingIP apply(@Nullable org.jclouds.openstack.neutron.v2.domain.FloatingIP floatingIP) {
+      FloatingIP.Builder<?> builder = FloatingIP.builder();
+      builder.id(floatingIP.getId());
+      builder.ip(floatingIP.getFloatingIpAddress());
+      builder.fixedIp(floatingIP.getFixedIpAddress());
+      builder.instanceId(floatingIP.getPortId());
```
nacx: Isn't `instanceId` the ID of the node? In that case, the value should be the `deviceId` of the port; not the id of the port itself.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/domain/FloatingIpForServer.java
```diff
@@ -0,0 +1,40 @@
...
+import org.jclouds.openstack.nova.v2_0.domain.regionscoped.RegionAndId;
+
+@AutoValue
+public abstract class FloatingIpForServer {
+
+   public abstract RegionAndId serverId();
+   public abstract String floatingIpId();
+   public abstract String ip();
+
+   @SerializedNames({"serverId", "floatingIpId", "ip" })
```
nacx: This object is not meant to be serialized to json. Remove this annotation.

apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeServiceLiveTest.java
```diff
@@ -51,7 +78,66 @@ protected LoggingModule getLoggingModule() {
...
+      return ImmutableSet.<Module> of(
+              ContextLinking.linkContext(neutronApiContext),
+              getLoggingModule(),
+              credentialStoreModule,
+              getSshModule()
+      );
+   }
+
+   @Override
+   protected TemplateBuilder templateBuilder() {
```
nacx: Remove thsi test with cloudsoft specific stuff?

apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeServiceLiveTest.java
```diff
@@ -62,17 +148,20 @@ public void testOptionToNotBlock() {
...
-         getAnonymousLogger().info(
-               "testSuspendResume() threw, probably due to unavailable AdminActions extension: " + e.getMessage());
+         getAnonymousLogger().info("testSuspendResume() threw, probably due to unavailable AdminActions extension: " + e.getMessage());
       }
    }
 
+   @Override
+   public void testAScriptExecutionAfterBootWithBasicTemplate() throws Exception {
+      super.testAScriptExecutionAfterBootWithBasicTemplate();
+   }
```
nacx: Remove this

apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeServiceLiveTest.java
```diff
@@ -51,7 +78,66 @@ protected LoggingModule getLoggingModule() {
...
    @Override
    protected Iterable<Module> setupModules() {
-      return ImmutableSet.of(getLoggingModule(), credentialStoreModule, getSshModule());
+      return ImmutableSet.<Module> of(
+              ContextLinking.linkContext(neutronApiContext),
+              getLoggingModule(),
+              credentialStoreModule,
+              getSshModule()
+      );
+   }
```
nacx: With this config we will *always* test nova+neutron, but not the path of only Nova. I think we should be able to still live test the old path for legacy environments.

apis/openstack-neutron/src/test/java/org/jclouds/openstack/neutron/v2/features/FloatingIPApiLiveTest.java
andreaturli: good point, I think `@Nullable` is wrong here as networkApi.list uses `EmptyPagedIterableOnNotFoundOr404`

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
andreaturli: hmm, makes sense, although I was using the same approach used for Nova. Should I change line 141 as well?

nacx: I'd say so?

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
andreaturli: very good point, shouldn't we fail instead as the user asked to attach a floating IP but jclouds can't satisfy the request because the port is not there?

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/NeutronFloatingIpToFloatingIp.java
andreaturli: ouch this function is not used anyway, so I'll remove it

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
nacx: Makes sense too :)

apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeServiceLiveTest.java
andreaturli: damn!

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
andreaturli: good point! as poolnames are optional, what shall we do if poolnames is empty? fail seems reasonable as we try to re-use an existing floating ip and to create from a poolname: if no poolnames are specified we can't infer much, right?

nacx: If pool names is empty I'd say it is ok to pick the first external network we find.  We should add proper logs about the network we are selecting.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
```diff
@@ -225,28 +228,64 @@ public boolean apply(@Nullable org.jclouds.openstack.neutron.v2.domain.FloatingI
...
+                 "Failed to find suitable external networks or to allocate from poolNames specified: "
+                 + Iterables.toString(poolNames.get()));
+      }
+      floatingIpCache.asMap().put(RegionAndId.fromSlashEncoded(node.getId()), ImmutableList.of(FloatingIpForServer.create(RegionAndId.fromSlashEncoded(node.getId()), floatingIP.getId(), floatingIP.getFloatingIpAddress())));
       return floatingIP;
    }
 
+
+   private List<Network> getNetworksFromPoolNames(String regionId, final Optional<Set<String>> poolNames) {
+      return getNetworkApi(regionId).list().concat().filter(new Predicate<Network>() {
```
nacx: if the pool name set is absent or empty, just return an empty list here. There is no need to filter in that case, nor to have that check in the predicate, applying it on every iteration.

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
```diff
@@ -225,28 +228,64 @@ public boolean apply(@Nullable org.jclouds.openstack.neutron.v2.domain.FloatingI
...
-              .availabilityZone(networkOptional.get().getAvailabilityZone())
-              .build();
+      // 2.) Attempt to allocate from a external network
+      if (floatingIP == null) {
+         Optional<Network> networkOptional = Iterables.tryFind(networks, new Predicate<Network>() {
+            @Override
+            public boolean apply(Network input) {
+               return availabilityZone.equals(input.getAvailabilityZone()) && input.getExternal();
+            }
+         });
```
nacx: If the pool names is a list of more than one element, we try to allocate the floating IP in one of the given networks. Why not apply the same pattern if no network is defined? We can try every external network in the same availability zone, following the exact same pattern.

apis/openstack-nova/src/test/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeServiceLiveTest.java
```diff
@@ -16,130 +16,31 @@
...
-              .credentials(setIfTestSystemPropertyPresent(overrides,
-                      "openstack-nova.identity"),
-                      setIfTestSystemPropertyPresent(overrides, "openstack-nova.credential"))
-              .modules(ImmutableSet.<Module>of(
-                      new SshjSshClientModule(),
-                      new SLF4JLoggingModule(),
-                      new BouncyCastleCryptoModule())
-              )
-              //        TypeToken<? extends Context> contextType = new TypeToken<ApiContext<NeutronApi>>() {};
-              .build();
```
nacx: Now there is no way to test the neutron integration? We should provide a NovaWithNeutron live test class to test this new feature.

andreaturli: ok I'll add it

apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/AllocateAndAddFloatingIpToNode.java
```diff
@@ -275,15 +258,24 @@ public boolean apply(Network input) {
...
-            return poolNames.isPresent() &&
-                   !poolNames.get().isEmpty() &&
-                   poolNames.get().contains(network.getName());
+   /**
+    * Get all suitable networks to allocate a floating ip
+    *
+    * It will prefer networks specified using the poolNames first and then the external networks in the given availability zone
+    */
+   private List<Network> getNetworksFromPoolNamesAndExternal(String regionId, final String availabilityZone, final Optional<Set<String>> poolNames) {
+      if (poolNames.isPresent() && poolNames.get().isEmpty()) return Lists.newArrayList();
```
nacx: This is not OK. If the pool names are not configured you'll never try to find an external network.
You'll need two different methods, or a method that accepts a filtering predicate.

andreaturli: you are right, I will fix it

andreaturli:
merging now

andreaturli:
Merged at [master](https://git1-us-west.apache.org/repos/asf?p=jclouds.git;a=commit;h=50ae6828)

PR #1181, opened 2018-02-16
core: Update method argument changes for lastest gson presently 2.8.2

wltjr:
Minor changes to update to latest gson. Seems passing null is save. The underlying class does null checks. Though may require further modifications.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -80,7 +80,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.5</version>
+      <version>2.8.2</version>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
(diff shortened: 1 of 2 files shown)


nacx:
Thanks for looking at this!

There is a limitation that prevents us from using gson >= 2.7. See: https://issues.apache.org/jira/browse/JCLOUDS-1166

Unless that is addressed too, we can't upgrade gson as we would break jclouds-karaf and our OSGi compatibility. Would you like to have a look at this?

wltjr:
No problem, it was minor. I was addressing it for my own needs so figured I would pass it along. But I am not using maven, so did not test that out...

I may look into that, though not super familiar with OSGi. Though not sure the issue relates to that as much as what is made available in gson. Thanks for mentioning the issue. I was not aware. Already have bit off way to much... So cautious to take on more. But if it becomes an issue for me I will address it for sure. Feel free to close PR. The changes are really minor. Can be done when ever. The OSGi changes likely are more substantial...

gaul:

PR #1182, opened 2018-02-19
JCLOUDS-1384: UT reproducing the problem

cstamas:
https://issues.apache.org/jira/browse/JCLOUDS-1384

Showing the effect of wrong steps: sort then lowercase produces
wrong string to sign.


nacx:
Thanks, @cstamas! Given that you already took your time to dig into the code and find the root cause, mind adding a fix to the PR?

cstamas:
Fixed by https://github.com/jclouds/jclouds/pull/1185

PR #1185, opened 2018-03-07
[JCLOUDS-1391] Sort headers alphabetically/lowercase before signing

adreghiciu:
https://issues.apache.org/jira/browse/JCLOUDS-1391


adreghiciu:
Please guide me to an integration test that creates a blob with user metadata and on information how to run it.
Test is as simple as creating a blob with user metadata like:
B=v1
a=v2

gaul:
You probably want to add your integration test to `GoogleCloudStorageBlobIntegrationLiveTest` and you can run it via:

```
mvn integration-test -pl :google-cloud-storage -Plive \
        -Dtest.google-cloud-storage.identity="${JCLOUDS_IDENTITY}" \
        -Dtest.google-cloud-storage.credential="${JCLOUDS_CREDENTIAL}"
        -Dtest=GoogleCloudStorageBlobIntegrationLiveTest#yourTestName
```

adreghiciu:
Added the test. Recalled that the bug only manifests when copying a blob with changed metadata.
You could lowercase the user metadata keys in copy options but still the fix will do the right thing which is sort headers lowercase as will Azure do when it will receive an uppercase header

gaul:
Removed a stray comment, squashed, and pushed to master as cbb76523cc4a8f306329ed5cc97f96af759a9967 and 2.1.x as fa88bcc0dd456048ae703bf22384c551c8794563.  Thank you for your contribution @adreghiciu!

PR #1192, opened 2018-04-07
JCLOUDS-1400: Honor signed URL content type

gaul:


gaul:
> Is there any test we should add to complete the patch?

Done.

gaul:
Upon reflection, I should change these tests to remove all the headers and rely only on URL parameters.

nacx:
LGTM, thanks!

PR #1194, opened 2018-04-09
Added new Google Storage locations

felipelopess:
**Regions removed:** 
US_CENTRAL2, US_EAST2, US_EAST3;

**Regions added:**
ASIA_NORTHEAST1, ASIA_SOUTH1, ASIA_SOUTHEAST1, NORTHAMERICA_NORTHEAST1, US_EAST4, SOUTHAMERICA_EAST1, EUROPE_WEST1, EUROPE_WEST2, EUROPE_WEST3, EUROPE_WEST4, AUSTRALIA_SOUTHEAST1;

### DIFF
--- a/providers/google-cloud-storage/src/main/java/org/jclouds/googlecloudstorage/domain/DomainResourceReferences.java
+++ b/providers/google-cloud-storage/src/main/java/org/jclouds/googlecloudstorage/domain/DomainResourceReferences.java
@@ -32,7 +32,34 @@ public final class DomainResourceReferences {
    }
 
    public enum Location {
-      ASIA, EU, US, ASIA_EAST1, US_CENTRAL1, US_CENTRAL2, US_EAST1, US_EAST2, US_EAST3, US_WEST1;
+
+     /*
+      *  Multi-Regional
+      */
+      ASIA, EU, US,
+
+      /*
+       *  Regional
+       */
+      ASIA_EAST1,
+      ASIA_NORTHEAST1,
+      ASIA_SOUTH1,
+      ASIA_SOUTHEAST1,
+
+      NORTHAMERICA_NORTHEAST1,
+      US_CENTRAL1,
+      US_EAST1,
+      US_EAST4,
+      US_WEST1,
+
+      SOUTHAMERICA_EAST1,
+
+      EUROPE_WEST1,
+      EUROPE_WEST2,
+      EUROPE_WEST3,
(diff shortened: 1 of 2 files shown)


providers/google-cloud-storage/src/main/java/org/jclouds/googlecloudstorage/domain/DomainResourceReferences.java
```diff
@@ -32,7 +32,34 @@ private DomainResourceReferences() {
    }
 
    public enum Location {
-      ASIA, EU, US, ASIA_EAST1, US_CENTRAL1, US_CENTRAL2, US_EAST1, US_EAST2, US_EAST3, US_WEST1;
```
gaul: The documentation suggests that `US_CENTRAL2` exists:

https://cloud.google.com/storage/docs/bucket-locations

Perhaps we should retain the region and mark it as `@Beta`?

felipelopess: I had looked at this same page you suggested and had not seen this location with this remark, sorry.
I think that DomainResourceReferences class should have different enums for each storage class because some locations can not be used in certain storage classes.
But I think at the moment this location can be placed back.

felipelopess: But there's another factor, I'm working on another system that needs to match the available resources with their respective price, and in Google's price list that region is not found in any resource, it might not be ideal to maintain that region.
This is the price list table offered by Google: https://cloudpricingcalculator.appspot.com/static/data/pricelist.json

gaul: I prefer not to break any application using this valid region.  Could you special case your code for this situation or consume the suggested `@Beta` annotation?

felipelopess: All right, I've put the annotation @Beta, thank you!

gaul:
Squashed, reworded commit message, and pushed to master as b04d2df6614996f850fffa05fa2203023ee1b2e7 and 2.1.x as 501fc93b6718fbb81c2f6d1cb50d40909d0719df.  Thank you for your contribution @felipelopess!

PR #1198, opened 2018-04-17
Allow jclouds to provision SoftLayer instances with a private image.

tommclaughlan:
Fix JCLOUDS-1410.

https://issues.apache.org/jira/browse/JCLOUDS-1410

If possible, can this be backported to 2.0.x?


nacx:
Pushed to master, 2.1.x and 2.0.x. It is unlikely, though, that there is a new bugfix release for 2.0. It will depend on user demand.
Thanks for the patch @tommclaughlan!

tommclaughlan:
Thanks! Is there a roadmap for jclouds 2.1.x releases?

nacx:
Not really. We usually release bugfix versions when there is a number of bugs or some critical ones, or when there is user demand, but we don't have hard dates for releases. Starting a discussion about the next bugfix release on the [user mailing list](https://lists.apache.org/list.html?user@jclouds.apache.org) would be the best way to make things start moving forward :)

PR #1200, opened 2018-04-20
JCLOUDS-1389: When composing, set destination.storageClass from object

psilly:
This aligns that the destination's storageClass with that of the sub-object(s) being composed


gaul:
This seems reasonable.  I wonder why none of the existing integration tests found this symptom?  Does it happen with only certain storage classes?

gaul:
retry

gaul:
rebuild

psilly:
Yes, it only happens when the storage class is not STANDARD. I left some testing notes in the Jira

gaul:
I added a test and pushed to master as def6db48645fb6c043a4889b45c582335b331d30 and f7b45341328410dea583901a31218a3588cb5aad and to 2.1.x as 5fa2df1bbcde533e8fa1386f8151af86aad52035 and c689d317580433e0289f4bd1dc54a1da9bf33a3b.  Thank you for your contribution @psilly!

PR #1201, opened 2018-05-03
Protects from NPE since SecurityGroup.getLocation() is nullable

danielestevez:
Right now running the AzureComputeSecurityGroupExtensionLiveTest will fail with a NPE if you have any securityGroup in you subscription in any location not included in Set<String> locations


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/7c592703) and [2.1.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a32690a2). Thanks @danielestevez!

PR #1205, opened 2018-05-11
Fixes Run SSH script for passwords with special characters 

danielestevez:
Passwords can contain special characters as parentheses, password parameter should be wrapped in '' to avoid this syntax errors in SSH command


andreaturli:
Haven't tested myself but lgtm, thanks @danielestevez

danielestevez:
It's just a sometimes since these kind of characters can be included in password random generator

https://github.com/danielestevez/jclouds/blob/9fef6ed06b027c1525579c88b58aa9f5833687fd/core/src/main/java/org/jclouds/util/PasswordGenerator.java#L65

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/82289948). @andreaturli I think the change is safe enough to cherry-pick it to 2.1.x, WDYT?

compute/src/main/java/org/jclouds/compute/callables/RunScriptOnNodeUsingSsh.java
```diff
@@ -112,7 +112,8 @@ protected ExecResponse runCommand(String command) {
    public String execAsRoot(String command) {
       if (node.getCredentials().identity.equals("root")) {
       } else if (node.getCredentials().shouldAuthenticateSudo()) {
-         command = String.format("sudo -S sh <<'%s'\n%s\n%s%s\n", MARKER, node.getCredentials().getOptionalPassword().get(), command, MARKER);
+         command = String.format("sudo -S sh <<'%s'\n'%s'\n%s%s\n", MARKER, node.getCredentials().getOptionalPassword
+               ().get(), command, MARKER);
       } else {
          command = String.format("sudo sh <<'%s'\n%s%s\n", MARKER, command, MARKER);
```
demobox: Does this line need a similar fix?

compute/src/test/java/org/jclouds/compute/callables/RunScriptOnNodeUsingSshTest.java
```diff
@@ -103,7 +103,7 @@ public void simpleRootTestWithSudoPassword() {
       expect(sshClient.getUsername()).andReturn("tester");
       expect(sshClient.getHostAddress()).andReturn("somewhere.example.com");
       expect(
-            sshClient.exec("sudo -S sh <<'RUN_SCRIPT_AS_ROOT_SSH'\n" + "testpassword!\n" + "echo $USER\n"
+            sshClient.exec("sudo -S sh <<'RUN_SCRIPT_AS_ROOT_SSH'\n" + "'testpassword!'\n" + "echo $USER\n"
```
demobox: Would it make sense to modify this test so that it actually reflects the problem we're trying to fix (e.g. space in the password)? Also, would it be worth adding a test for the additional code path (see comment above)?

nacx: +1. Let's change the default value used in tests to a conflicting one. @danielestevez can you have a look at these comments and submit a PR?

danielestevez: @demobox 
That line of code https://github.com/jclouds/jclouds/pull/1205#discussion-diff-187774732R118 is already covered in simpleRootTest and testUserAddAsRoot tests

This PR adds a conflictive password for all tests https://github.com/jclouds/jclouds/pull/1207

nacx:
Cherry-picked to [2.1.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/889a7f1d).

PR #1206, opened 2018-05-11
Increases session interval from 60 seconds to a higher value

danielestevez:
To avoid problems with the cache, and specifically running the AzureComputeImageExtensionLiveTest suite.

ARM is quite slow and it takes usually more than one minute (default value) for certain operations. The testImageIsCachedAfterBeingCreated in that suite fails often since it takes longer than one minute to destroy the created node in testCreateImage and cache does not contain the image.

It should be safe to increase this value in the provider metadata. 5 minutes looks good but open to discuss

Would be good to add this also to https://github.com/danielestevez/jclouds/blob/83c0a3c7b255ec744c6150ce76c40e8301280c79/providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/AzureLiveTestUtils.java ?

### DIFF
--- a/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeProviderMetadata.java
+++ b/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeProviderMetadata.java
@@ -17,6 +17,7 @@
 package org.jclouds.azurecompute.arm;
 
 import static org.jclouds.Constants.PROPERTY_MAX_RATE_LIMIT_WAIT;
+import static org.jclouds.Constants.PROPERTY_SESSION_INTERVAL;
 import static org.jclouds.azurecompute.arm.config.AzureComputeProperties.API_VERSION_PREFIX;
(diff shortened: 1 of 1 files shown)


nacx:
>Would be good to add this also to https://github.com/danielestevez/jclouds/blob/83c0a3c7b255ec744c6150ce76c40e8301280c79/providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/AzureLiveTestUtils.java ?

That class just adds *overrides* for default provider properties, so as long as you put the right default in the provider metadata there is no need to also configure it in the tests.

nacx:
The build failed. It looks like there's a missing static import.

danielestevez:
Sorry about that, fixed

nacx:
>It should be safe to increase this value in the provider metadata. 5 minutes looks good but open to discuss

Looks like a sane default.

Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a08de5b2) and [2.1.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/491911ec). Thanks @danielestevez!

PR #788, opened 2015-06-25
JCLOUDS-930: Plumb through the prefix option

timuralp:
Plumbs the prefix option to the S3, Azure, OpenStack Swift, Local, and Transient providers. The option allows listing all blobs that start with a given prefix.

PR #1212, opened 2018-05-23
[JCLOUDS-1422] Ensure that LocalBlobStore.list honors !recursive even when a prefix is defined

jglick:
[JCLOUDS-1422](https://issues.apache.org/jira/browse/JCLOUDS-1422)

Amends #788.

FYI @carlossg: encountered during https://github.com/jenkinsci/artifact-manager-s3-plugin/pull/28


blobstore/src/test/java/org/jclouds/blobstore/strategy/internal/ListContainerTest.java
```diff
@@ -72,7 +72,7 @@ public void testListWithPrefix() {
       blobStore.putBlob(containerName, blobStore.blobBuilder("bar").payload("").build());
 
       Iterable<? extends StorageMetadata> results = concatter.execute(containerName,
-            ListContainerOptions.Builder.prefix(prefix).recursive());
+            ListContainerOptions.Builder.prefix(prefix)./* irrelevant */recursive());
```
gaul: Stray comment?

jglick: No, noting that this test passes regardless of which way you set the `recursive` flag, which makes the name of the test a bit surprising.

jglick:
also CC @kuisathaverat

gaul:
Rebased and pushed to master as 36353d6157a1c6c3f8f5e8497a68802a79784b12 and 11fac4f020b02a45ef9809e19656325ddf9efb25 and 2.1.x as 8d2808e3091a888e9812149e68fecd294628c0d3 and 3cb8d4f87ccb69b0ab4705ab6da6a1976b0c537d.  Thank you for your contribution @jglick!

PR #1217, opened 2018-06-14
Fixes/adds fields primary for NetworkInterfaceCard

danielestevez:
* Field 'primary' was missing in NetworkInterfaceCardProperties
* Field 'primary' was misplaced in IpConfiguration instead of IpConfigurationProperties

This causes problems when using the provider to configure multiple IPs per NIC and multiple NICs per VirtualMachine

Check https://docs.microsoft.com/en-us/rest/api/virtualnetwork/networkinterfaces/get#networkinterface


nacx:
Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/e4f1823f) and [2.1.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/cace304e). Thanks @danielestevez!

PR #1197, opened 2018-04-15
JCLOUDS-1333: Require JDK 1.8

gaul:
This requires Guava 21 and allows use with later releases.  This also
sets modernizer to JDK 1.6 to work around the many new violations.

PR #1224, opened 2018-07-09
Configure Java 7 language level

nacx:
There is already https://github.com/jclouds/jclouds/pull/1197, but while we complete the work to properly support Java 8 in every project (just jclouds-karaf pending) we should not block contributions from using Java 7 stuff.


andreaturli:
lgtm

nacx:
Merged to master

project/pom.xml
```diff
@@ -1280,7 +1280,10 @@
             <groupId>org.gaul</groupId>
             <artifactId>modernizer-maven-plugin</artifactId>
             <configuration>
+              <!-- Fix fiolations before uncommenting
```
demobox: @nacx Is the plan to address this as part of the Java 8 support?

nacx: Yes, definitely. I wanted to enter this change because this was blocking https://github.com/jclouds/jclouds-labs/pull/439, but we definitely need to get those warnings fixed.

PR #1225, opened 2018-07-09
Adds listall methods to NetworkInterfaceCardApi and PublicIPAddressApi

danielestevez:


providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiLiveTest.java
```diff
@@ -105,6 +105,12 @@ public void listNetworkInterfaceCards() {
       assertTrue(nicList.contains(api().get(nicName)));
    }
 
+   @Test(dependsOnMethods = "createNetworkInterfaceCard")
+   public void listAllNetworkInterfaceCards() {
```
nacx: Make the delete test depend on this method too, to avoid deleting the networks first, causing this test to fail.

providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiMockTest.java
```diff
@@ -87,6 +88,28 @@ public void listNetworkInterfaceCards() throws InterruptedException {
...
+      assertEquals(nicList.get(0).properties().ipConfigurations().get(0).properties().privateIPAllocationMethod(),
+            "Dynamic");
+      assertTrue(nicList.get(1).properties().ipConfigurations().size() > 0);
+      assertEquals(nicList.get(1).properties().ipConfigurations().get(0).properties().privateIPAllocationMethod(),
+            "Static");
+      assertTrue(nicList.get(2).properties().ipConfigurations().size() > 0);
+      assertNotEquals(IdReference.extractResourceGroup(nicList.get(2).id()), resourcegroup);
+      assertEquals(nicList.get(2).properties().ipConfigurations().get(0).properties().privateIPAllocationMethod(),
+            "Static");
+   }
```
nacx: Also add a test that exercises the 404 fallback defined for this new method.

providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/PublicIPAddressApiLiveTest.java
```diff
@@ -119,6 +119,16 @@ public void listPublicIPAddresses() {
       assertTrue(ipList.size() > 0);
    }
 
+   @Test(groups = "live", dependsOnMethods = "createPublicIPAddress")
+   public void listAllPublicIPAddresses() {
```
nacx: Same here. Add it as a dependency to the delete test.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApi.java
```diff
@@ -45,19 +44,26 @@
...
    @GET
    @Fallback(EmptyListOnNotFoundOr404.class)
    List<NetworkInterfaceCard> list();
 
+   @Named("networkinterfacecard:list_all")
+   @Path("/providers/Microsoft.Network/networkInterfaces")
+   @SelectJson("value")
+   @GET
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<NetworkInterfaceCard> listAll();
```
nacx: The method name looks confusing, as `list` usually means "list all". I think it would make sense to rename it to `listAllInSubscription`, or similar. WDYT? The same applies for the one added to the public ip api.

providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiLiveTest.java
danielestevez: Good catch, that test could definitely fail.

providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiMockTest.java
danielestevez: Forgot that one, sorry

providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/PublicIPAddressApiLiveTest.java
danielestevez: same :)

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApi.java
danielestevez: Been thinking about this one... i like *listAll* since the path already has the subscription and requires no resourcegroup, so *listAll* is (obviously?) retrieving all the IPs in subscription.
However this led me to notice a bigger problem: to retrieve all Ips in subscription you're now required to provide a resourcegroupname just to get the reference this API and this makes no sense
 https://github.com/danielestevez/jclouds/blob/00b6697d096b54a1f4dc6cbb6a308f91cb0728fc/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeApi.java#L118

I think we should remove the required *@PathParam("resourcegroup") String resourcegroup* in this case, but it looks like a big change... see any problem with this?

Same applies to the publicIpAddressApi

nacx: Yes, that was my point. To get the API you have to pass the resource group, then the method name is confusing.

I see three options:

* Create another api, say `GlobalNetworkApi` with an accessor that does not require the resource group, and put there the method to list all. The con is that you need one specific API for every *listAll*  method you want to expose.
* Create a `SubscriptionApi` and put there all methods that access subscription-wide resources.
* Since it is just one method that causes confusion, it is not a big deal to rename it to make it clear what it does.

As per my comment, I'd go for the third option, as users already know that all network operations are exposed in the network API, and for me it makes more sense, even though that resource group is not going to be used by this concrete method.

danielestevez: But in that 3rd option, the user is forced to pass a resource group he doesn't know anything about (it could even not exist) and that's very confusing

I added 2 commits with the changes to those APIs but i guess that breaks compatibility with clients and won't be accepted, right? I believe, however, it's the best way to go

nacx: You can declare it nullable and pass a null value there. I'd say that's a better option (for one method that was not needed until now, so we can assume there is no big impact in that low-usage path).
Otherwise, let's go for the global APIs instead of changing the existing APIs that much. The azure provider is no longer in labs and we must keep backward compatibility. If you want to go that path, then you need to `@Deprecate` the methods first till the next release, leave one release for deprecation with a clear upgrade path, then finally remove it in the next one.

danielestevez: The Nullable solution seems the less intrusive solution for this case, i guess. Will do

danielestevez:
Removed previous solution and added the optional/nullable resourcegroup parameter as solution!

PR #1236, opened 2018-08-24
[JCLOUDS-1415] - Support GCE shared VPC

tormath1:
This is the [jira](https://issues.apache.org/jira/browse/JCLOUDS-1415). I've wrote one test for the firewall loader, but I think it's not enough. 
And I've tried to make a Minimum Viable Change in order to don't have a lot of new files and things to review. 
Thanks for your review !

### DIFF
--- a/providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/config/GoogleComputeEngineServiceContextModule.java
+++ b/providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/config/GoogleComputeEngineServiceContextModule.java
@@ -55,17 +55,14 @@ import org.jclouds.googlecomputeengine.compute.functions.OrphanedGroupsFromDeadN
 import org.jclouds.googlecomputeengine.compute.functions.Resources;
 import org.jclouds.googlecomputeengine.compute.functions.ResetWindowsPassword;
 import org.jclouds.googlecomputeengine.compute.loaders.DiskURIToImage;
+import org.jclouds.googlecomputeengine.compute.loaders.FirewallLoader;
 import org.jclouds.googlecomputeengine.compute.loaders.SubnetworkLoader;
 import org.jclouds.googlecomputeengine.compute.options.GoogleComputeEngineTemplateOptions;
 import org.jclouds.googlecomputeengine.compute.predicates.AtomicInstanceVisible;
 import org.jclouds.googlecomputeengine.compute.predicates.AtomicOperationDone;
 import org.jclouds.googlecomputeengine.compute.predicates.GroupIsEmpty;
 import org.jclouds.googlecomputeengine.compute.strategy.CreateNodesWithGroupEncodedIntoNameThenAddToSet;
-import org.jclouds.googlecomputeengine.domain.Image;
-import org.jclouds.googlecomputeengine.domain.Instance;
-import org.jclouds.googlecomputeengine.domain.MachineType;
-import org.jclouds.googlecomputeengine.domain.Operation;
-import org.jclouds.googlecomputeengine.domain.Subnetwork;
+import org.jclouds.googlecomputeengine.domain.*;
 import org.jclouds.location.suppliers.ImplicitLocationSupplier;
 import org.jclouds.location.suppliers.implicit.FirstZone;
 
@@ -143,6 +140,9 @@ public final class GoogleComputeEngineServiceContextModule
       
       bind(new TypeLiteral<CacheLoader<RegionAndName, Optional<Subnetwork>>>() {
       }).to(SubnetworkLoader.class);
+
+      bind(new TypeLiteral<CacheLoader<URI, Optional<Firewall>>>(){
+      }).to(FirewallLoader.class);
 
       bindHttpApi(binder(), Resources.class);
    }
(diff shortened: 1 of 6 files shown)


providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/config/GoogleComputeEngineServiceContextModule.java
```diff
@@ -220,6 +225,14 @@ protected void configure() {
       return CacheBuilder.newBuilder().build(in);
    }
 
+
+@Provides
+@Singleton
+protected LoadingCache<URI, Optional<Firewall>> firewallsMap(
+        CacheLoader<URI, Optional<Firewall>> in) {
+    return CacheBuilder.newBuilder().build(in);
+}
```
nacx: [minor] fix indentation

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/functions/Resources.java
```diff
@@ -96,4 +97,9 @@
    @Named("Images:get")
    @GET
    @Fallback(NullOnNotFoundOr404.class) @Nullable Image image(@EndpointParam URI selfLink);
+
+   /**   Returns a firewall by self-link or null if not found*/
```
nacx: [minor] fix leading and trailing spaces in the comment.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/loaders/FirewallLoader.java
```diff
@@ -0,0 +1,53 @@
...
+            throw new ExecutionException(message(key, ex), ex);
+        }
+    }
+
+    public static String message(URI key, Exception ex) {
+        return String.format("could not find firewall %s: %s", key.toString(), ex.getMessage());
+    }
+
+
+}
```
nacx: jclouds uses a 3 space indent. Please format the code accordingly.

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/strategy/CreateNodesWithGroupEncodedIntoNameThenAddToSet.java
```diff
@@ -147,18 +149,27 @@
...
 
       if (options.getNetworks().isEmpty()) {
          networkName = DEFAULT_NETWORK_NAME;
       } else {
          Iterator<String> iterator = options.getNetworks().iterator();
-         networkName = nameFromNetworkString(iterator.next());
+         net = iterator.next();
+         if (net.startsWith("projects")) {
+            isFullURI = true;
+         }
```
nacx: Just `isFullURI = net.startsWith("projects")`

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/strategy/CreateNodesWithGroupEncodedIntoNameThenAddToSet.java
```diff
@@ -147,18 +149,27 @@
...
+         networkName = nameFromNetworkString(net);
          checkArgument(!iterator.hasNext(),
                "Error: Please specify only one network/subnetwork in TemplateOptions when using GCE.");
       }
 
       String region = ZONE == location.getScope() ? location.getParent().getId() : location.getId();
-      Optional<Subnetwork> subnet = subnetworksMap.getUnchecked(fromRegionAndName(region, networkName));
+      Optional<Subnetwork> subnet;
+
+      subnet = isFullURI ? Optional.fromNullable(resources.subnetwork(URI.create(net))) : subnetworksMap.getUnchecked(fromRegionAndName(region, networkName));
```
nacx: Could this logic be abstracted in the subnet cache? Otherwise, we are catching subnets only if the provided ID is not the full URI, which does not seem right?
This way you save an extra call in the adapter. With the current code, you are calling the `resources` API here and calling it again in the adapter.

nacx (commented):
Thanks @tormath1!

providers/google-compute-engine/src/main/java/org/jclouds/googlecomputeengine/compute/strategy/CreateNodesWithGroupEncodedIntoNameThenAddToSet.java
tormath1: Ok. But a shared VPC is identified by his `host project`, `region` and `name` and a classical subnetwork is only identified by his `region` and `name`, the `project` is given by the service account. 
In my opinon, if a user wants to use a shared VPC, it'll have to give the full URI and if he wants to use a classical subnetwork, it'll have to give only the `region` and `name`. In this way, we are not breaking the compatibility. 
This is why, I am checking if the given `net` is like `projects/<project>/regions/<region>/subnetworks/<subnetwork>`.

nacx: Ok. The main issue here is that the cache key is `RegionAndName`. What about creating a different class to be used with these loaders that can cache shared resources, say `RegionAndNameInProject`? In the case of shared resources we can build it from the full URI, and for others we can just get injected the current project (we already have the project bound to the Guice contecxt as `@Named(PROJECT_NAME) String currentProject`) and build it too. This way we can still have backwards compatibility and properly cache all subnets whether they are shared or not, saving unnecessary repeated API calls. WDYT?

gaul:
@nacx @tormath1 does this pull request need anything to move forward?

gaul:

PR #1237, opened 2018-08-29
JCLOUDS-1441: Enables support for ARM regions in China

danielestevez:
- Only one blocker hardcoded was fixed in AzureComputeHttpApiModule.java
- Endpoints for GraphAPI and VaultAPI are also hardcoded but we could deal with this as separate PRs (i couldn't make this work with Guice)
- Also added two new China regions to keep the plugin up to date.

Just with this and using the right credentials and oauth resource and endpoint for the China regions i could run all the livetests (see Jira)


nacx:
Thanks! A couple hints to do this properly:

Is there any way we can determine if the `ContextBuilder` was created for China? (For example by looking at the `endpoint` that has been configured?

If we can determine that, then I'd suggest the following approach to make the whole thing configurable:

1. Create a `@Provides` method in the Http API Module, that gets the endpoint (or whatever that allows us to know if it is China), and returns a boolean saying if we're in China or not. We can also qualify that method with a name. The idea is to be able to inject that `isChina` boolean wherever we need it.
2. Modify the [GraphRBAC](https://github.com/jclouds/jclouds/blob/master/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/GraphRBAC.java) endpoint supplier to get the `isChina` boolean injected and produce the right URI in its `get()` method.
3. Modify the [OAuthResource](https://github.com/jclouds/jclouds/blob/master/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/OAuthResource.java) to accept a second value (`String china()` or similar) so we can use the annotation to configure the normal endpoint and the endpoint for China regions.
4. Modify the [AzureOAuthConfigFactory](https://github.com/jclouds/jclouds/blob/master/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureOAuthConfigFactory.java) to get the `isChina` boolean injected and take it into account to read the right field of the `OAuthResource` annotation and return the right endpoint.

With this we should be good to get the right endpoints automatically configured.

danielestevez:
Yes, a simple `isChina` is easily implementable as we can parse `ContextBuilder.endpoint` 
Thanks for the tip, i'll try that approach and update the PR since i think we can suppose this will be the only exception and Microsoft won't be adding other additional endpoints.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
```diff
@@ -92,7 +97,14 @@ protected final String provideTenant(@Named("oauth.endpoint") final String oauth
...
       }
-      return m.group(1);
+      return m.group(2);
+   }
+
+   @Provides
+   @Singleton
+   @Named(IS_CHINA_ENDPOINT)
+   protected final boolean isChinaEndpoint(@Named("oauth.endpoint") final String oauthEndpoint) {
+      return CHINA_OAUTH_ENDPOINT_PATTERN.matcher(oauthEndpoint).matches();
```
nacx: Add proper unit tests for this method

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
```diff
@@ -54,7 +54,12 @@
 public class AzureComputeHttpApiModule extends HttpApiModule<AzureComputeApi> {
 
    private static final Pattern OAUTH_TENANT_PATTERN = Pattern
-         .compile("https://login.microsoft(?:online)?.com/([^/]+)/oauth2/token");
+         .compile("https://login.(microsoft(?:online)?.com|chinacloudapi.cn)/([^/]+)/oauth2/token");
+
+   private static final Pattern CHINA_OAUTH_ENDPOINT_PATTERN = Pattern
+         .compile("https://login.(chinacloudapi.cn)/([^/]+)/oauth2/token");
```
nacx: nit. You don't need the first grouping

danielestevez:
Squashed and merged to [2.1.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/951b6b44) and [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/a07ab5a9)

PR #1238, opened 2018-09-11
JCLOUDS-1447: URL encode source header on copyObject

davidcurrie:
As described in https://jira.apache.org/jira/browse/JCLOUDS-1447, S3 CopyObject requires the `x-amz-copy-source` header to be URL encoded.

This is not universally true of all headers though (for example the equals in `x-amz-copy-source-range`) therefore introducing a new parameter on `@Headers` to indicate whether URL encoding should take place.

Also updated `testUseBucketWithUpperCaseName` to remove the testing for legacy naming conventions as these are no longer supported by AWS.


apis/s3/src/test/java/org/jclouds/s3/S3ClientLiveTest.java
```diff
@@ -408,6 +408,21 @@ public void testCopyObject() throws Exception {
       }
    }
 
+   String sourceKeyRequiringEncoding = "apples#?:$&'\"<>čॐ";
```
gaul: Should this be a local variable?

demobox: +1 to move into the test, if possible

core/src/main/java/org/jclouds/rest/internal/RestAnnotationProcessor.java
```diff
@@ -763,6 +763,8 @@ private static void addHeader(Multimap<String, String> headers, Headers header,
       for (int i = 0; i < header.keys().length; i++) {
          String value = header.values()[i];
          value = replaceTokens(value, tokenValues);
+         if (header.urlEncode().length > i && header.urlEncode()[i])
```
demobox: [minor] Change to `i < header.urlEncode().length && header.urlEncode()[i]`? The bounds check seems to be written that way around more commonly, which might make it easier to understand that's going on here

nacx:
>can you think of any other providers that need to URL encode headers?

I'm mostly familiar with compute providers, and those don't use HTTP headers as extensively as Blob storage providers, so I'd say we're fine.

@davidcurrie could you please add some unit tests to the `RestAnnotationProcessorTest` class? There are several tests that verify the behavior of the `@Headers` annotation and we'd rather have some tests there that verify the encoding feature.

gaul:
Reworded commit message, shuffled some code between commits, and pushed to master as 5803de0f8ea0fef324b15c4a5a3a801ea7e505f6 and 7ebf12bf3867c7cdead772b8c80675c8a41bf7fa and 2.1.x as 8693687766380901057f1e69070723e7f708bede and f74d1c0976538f343cd192c18b2fccf91eb22b57.  Thank you for your contribution @davidcurrie!  Please use separate pull requests for unrelated commits in the future.

PR #1239, opened 2018-09-11
Update DomainResourceReferences.java

smedavaram75:
Hi, GCB allows the user to create a bucket with Storage Class REGIONAL however, this jClouds API doesn't have the enum for the Storage class as REGIONAL.  That's the reason I have updated this class.


gaul:
This looks good but could you amend the commit message with the prefix JCLOUDS-1419?  This will tag it correctly in Jira.

gaul:
I don't think this is possible so I amended the commit locally and pushed to master as 17cf6fec9e992bf37e58fd78b59042ce37f33304 and 2.1.x as f60e3e4a57fad005156c3a9fd992b9335df1f089.  Thank you for your contribution @smedavaram75!

smedavaram75:
Thanks a lot.  As this is first time I am contributing, what happens next?  How can I get this change?

gaul:
You could wait a few months for the next jclouds release or use snapshot releases in the meantime.  See "Using the daily builds" under Maven at https://jclouds.apache.org/start/install/ .  You probably want to use 2.1.2-SNAPSHOT.

PR #1240, opened 2018-09-12
Fix/azure resource removal

nakomis:
Fixes an issue where Azure ARM resource groups are not deleted when a node is destroyed. This fix resolves the following Jiras

https://issues.apache.org/jira/browse/JCLOUDS-1330
https://issues.apache.org/jira/browse/JCLOUDS-1331


providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
```diff
@@ -65,6 +72,20 @@ public void initializeContext() {
...
+
+   @Test(groups = "live", enabled = true)
+   public void testResourceGroupDeletedOnDestroy() throws Exception {
+      template = buildTemplate(templateBuilder());
+      nodes = newTreeSet(client.createNodesInGroup(group, 1, template));
+      NodeMetadata node = nodes.first();
+      client.destroyNode(node.getId());
+      List<Resource> resources = view.unwrapApi(AzureComputeApi.class).getResourceGroupApi().resources(resourceGroupName);
+      if (!resources.isEmpty()) {
+         Assert.fail("Resource group was not empty, contained " + resources);
```
nacx: nit: static import

providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/strategy/CleanupResourcesTest.java
```diff
@@ -0,0 +1,314 @@
...
+        expect(diskApi.get("myOsDisk")).andReturn(null);
+
+        replay(api, diskApi);
+
+        // Do call
+        cleanupResources(api).cleanupManagedDisks(vm);
+
+        // Verify deleted resource group
+        verify(api, diskApi);
+    }
```
nacx: Now that you're adding this (thanks!) add a method to verify that the virtual networks are deleted too.

Also add a test to verify that the resources that _don't_ have the jclouds tags (availability sets, networks, etc), are not deleted.

nacx (commented):
Thanks, @nakomis! And many thanks for taking care of adding new tests. It's highly appreciated!

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/config/AzurePredicatesModule.java
```diff
@@ -110,6 +122,27 @@ protected VirtualMachineInStatePredicateFactory provideNodeSuspendedPredicate(fi
...
+         final PollPeriod pollPeriod) {
+      long timeout = timeouts.nodeTerminated;
+      return retry(new Predicate<IdReference>() {
+         @Override
+         public boolean apply(final IdReference input) {
+            List<org.jclouds.azurecompute.arm.domain.Resource> attachedResources = api.getResourceGroupApi().resources(input.resourceGroup());
+            Optional<org.jclouds.azurecompute.arm.domain.Resource> resourceInGroup = Iterables.tryFind(attachedResources, new Predicate<org.jclouds.azurecompute.arm.domain.Resource>() {
+               @Override
+               public boolean apply(org.jclouds.azurecompute.arm.domain.Resource resource) {
+                  return resource.id().equalsIgnoreCase(input.id());
```
demobox: Curious about this - in which cases do we expect IDs to have different cases? Is there some documentation we could like to in a comment here to explain why we need a case-insensitive comparison here?

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CleanupResources.java
```diff
@@ -121,7 +131,17 @@ public boolean cleanupVirtualMachineNICs(VirtualMachine virtualMachine) {
...
                logger.debug(">> deleting public ip %s...", publicIpName);
-               deleted &= api.getPublicIPAddressApi(publicIpResourceGroup).delete(publicIpName);
+               try {
+                  boolean ipDeleted = api.getPublicIPAddressApi(publicIpResourceGroup).delete(publicIpName);
+                  if (!ipDeleted) {
+                     logger.warn(">> ip not deleted %s...", ip);
+                  }
+                  deleted &= ipDeleted;
+               } catch (Exception ex) {
+                  logger.warn(">> Error deleting ip %s", ip);
```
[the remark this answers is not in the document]
demobox: Also: log the exception?

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CleanupResources.java
```diff
@@ -196,17 +244,47 @@ public boolean cleanupAvailabilitySetIfOrphaned(VirtualMachine virtualMachine) {
...
+            try {
+               // Virtual networks cannot be deleted if there are any resources attached. It does not seem to be possible
+               // to list devices connected to a virtual network
+               // https://docs.microsoft.com/en-us/azure/virtual-network/manage-virtual-network#delete-a-virtual-network
+               // We also check the tags to ensure that it's jclouds-created
+               api.getVirtualNetworkApi(resourceGroupName).delete(virtualNetwork.name());
+               resourceRemoved.apply(IdReference.create(virtualNetwork.id()));
+            } catch (IllegalArgumentException e) {
+               if (e.getMessage().contains("InUseSubnetCannotBeDeleted")) {
+                  // Can be ignored as virtualNetwork is in use
```
demobox: Curious about this comment - does it mean we _can_ or we _have to_ try not to delete it? Given that the intent here is to delete, as Ignasi points out, should the log message be a `warn` instead?

If so, we can probably remove the comment, as the log message says pretty much the same thing

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CleanupResources.java
```diff
@@ -196,17 +244,47 @@ public boolean cleanupAvailabilitySetIfOrphaned(VirtualMachine virtualMachine) {
...
-      if (api.getResourceGroupApi().resources(group).isEmpty()) {
+      List<org.jclouds.azurecompute.arm.domain.Resource> attachedResources = api.getResourceGroupApi().resources(group);
+
+      if (attachedResources.isEmpty()) {
          logger.debug(">> the resource group %s is empty. Deleting...", group);
-         deleted = resourceDeleted.apply(api.getResourceGroupApi().delete(group));
+         URI uri = api.getResourceGroupApi().delete(group);
+         deleted = uri == null || resourceDeleted.apply(uri);
+      } else {
+         logger.debug(">> the resource group %s contains %s. Not deleting...", group, attachedResources);
```
demobox: Similar comment as above: should this be `logger.warn`?

providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/strategy/CleanupResourcesTest.java
```diff
@@ -0,0 +1,314 @@
...
+import org.testng.annotations.Test;
+
+import com.google.common.base.Predicate;
+import com.google.common.base.Predicates;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
+
+@Test(groups = "unit", testName = "CleanupResourcesTest")
+public class CleanupResourcesTest {
```
demobox: If the case-insensitive resource ID comparison logic is important, do we need a test for that?

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/config/AzurePredicatesModule.java
danielestevez: We should expect different cases.... everywhere :( Azure ARM is case insensitive and our provider should adapt to this. 
There are currently lots of places where this could fail since we use equals or Map keys.
I had this trouble here: https://stackoverflow.com/questions/50068324/azure-arm-api-returns-locations-with-inconsistent-case/50977265#50977265 and it seems to happen to others too https://stackoverflow.com/questions/48561304/resource-group-name-case-insensitive-in-disk-ids
So in short, yes, we should always use equalsIgnoreCase for safety

demobox: > So in short, yes, we should always use equalsIgnoreCase for safety

Thanks for explaining, @danielestevez! In that case, would it make sense to add a test for that as part of this PR?

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CleanupResources.java
```diff
@@ -130,25 +150,50 @@ public boolean cleanupVirtualMachineNICs(VirtualMachine virtualMachine) {
...
+      };
+
+      Predicate<IdReference> filter = retry(diskDeleted, 1200, 5, 15, SECONDS);
+      Set<IdReference> nonDeleted = Sets.filter(deleteDisks, filter);
+
+      if (!nonDeleted.isEmpty()) {
+         logger.warn(">> disks not deleted: %s", Joiner.on(',').join(nonDeleted));
+      }
+
+      Set<IdReference> disksNotInRg = Sets.filter(deleteDisks, resourceRemoved);
```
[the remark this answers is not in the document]
nakomis: Hi @nacx The call to the predicate causes the method to block until the resource has been removed from the resource group (in addition to being deleted), so it is always necessary to call it

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CleanupResources.java
```diff
@@ -171,9 +216,12 @@ public boolean cleanupSecurityGroupIfOrphaned(String resourceGroup, String group
                logger.debug(">> deleting orphaned security group %s from %s...", name, resourceGroup);
                try {
                   deleted = resourceDeleted.apply(sgapi.delete(name));
+                  resourceRemoved.apply(IdReference.create(securityGroup.id()));
```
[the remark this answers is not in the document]
nacx: Then, please add a comment in places where we are using this, to make explicit this is done to deal with eventual consistency issues. Otherwise, we might be tempted to accidentally change this in the future.

nacx:
> Hi, thanks for the feedback. I've addressed a couple of the comments above, and pushed a couple of changes to address the others

Looks like the changes are note pushed?

gaul:
@nakomis does this pull request need anything to move forward?

gaul:

PR #1242, opened 2018-10-16
Mark srouceImage as Nullable to avoid NullPointerException while attaching local-ssd disk in GCE

maismail:


demobox:
@nacx Should we consider adding any kinds of test(s) for this? Also, would a JIRA issue for tracking make sense here?

gaul:
@maismail Could you fix the srouceImage typo in the commit message?

nacx:
Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/f7b83f45) and [2.1.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/71dbce55). Thanks @maismail!

PR #1251, opened 2018-10-24
JCLOUDS-1450: Use S3-style ETags for MPUs.

timuralp:
S3 uses a different ETag for multipart uploads (MPUs) than regular
objects. The ETag consists of the md5 hash of the concatenated ETags of
individual parts followed by the number of parts (separated by "-").

The patch changes the LocalBlobStore's implementation of
CompleteMultipartUpload to set the S3-style ETag before calling
putBlob() and return that ETag to the caller.

To simplify testing, a new protected method with a default NOOP
implementation is added to the BaseBlobIntegrationTest. It allows
providers to further verify MPUs (i.e. ensuring the correct ETag has
been stored alongside the object).

PR #1252, opened 2018-10-25
Filesystem: Fix the MPU ETags to match S3.

timuralp:
Prior commit introduced a bug in the computation of the MPU ETag value,
where it was concatenating strings, rather than operating on the bytes
of the integer value.


timuralp:
rebuild please

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
```diff
@@ -113,19 +121,59 @@ public void testSetBlobAccess() throws Exception {
...
          .toString();
       assertThat(blob.getMetadata().getETag()).isEqualTo(expectedETag);
    }
 
+   @Test(groups = { "integration", "live" })
+   public void testMultipartUploadMultiplePartsKnownETag() throws Exception {
+      BlobStore blobStore = view.getBlobStore();
+      String container = getContainerName();
+      // Pre-computed ETag returned by AWS S3 for the MPU consisting of two 5MB parts filled with 'b'
+      String expectedETag = "\"84462a16f6a60478d50148808aa609c1-2\"";
```
demobox: Just double-checking that the value that S3 returns is also the value that we'd want here, even with the filesystem provider. Assuming it's worth it, have we been able to verify that other providers produce the same value?

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
```diff
@@ -113,19 +121,59 @@ public void testSetBlobAccess() throws Exception {
...
+         assertThat(parts.get(0).partSize()).isEqualTo(part1.partSize());
+         assertThat(parts.get(0).partETag()).isEqualTo(part1.partETag());
+         assertThat(parts.get(1).partNumber()).isEqualTo(part2.partNumber());
+         assertThat(parts.get(1).partSize()).isEqualTo(part2.partSize());
+         assertThat(parts.get(1).partETag()).isEqualTo(part2.partETag());
+
+         blobStore.completeMultipartUpload(mpu, ImmutableList.of(part1, part2));
+
+         BlobMetadata newBlobMetadata = blobStore.blobMetadata(container, name);
+         assertThat(newBlobMetadata.getETag()).isEqualTo(expectedETag);
```
demobox: [minor] Is there anything other than this final assertion (regarding the expected ETag) that we're trying to verify in this test?

timuralp: No, there isn't. I considered validating the content itself, but there are existing tests that do that. I can add the content validation if you see value to it.

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
demobox: Ah, just seeing #1251. A follow-up question to that (also /cc @gaul): is S3-compatibility considered a filesystem provider goal? How (if at all) do other blobstore providers treat this case?

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
demobox: Sorry, I wasn't clear: I think it's fine to have *fewer* validations, so it's clear to the reader immediately what the "main" point of this test is. E.g. skip the assertions on number and size of parts, if that's not what we care about here?

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
timuralp: Good question. Some providers return a similar value, like Swift for static large objects. Others have a different ETag, like Azure, which uses an opaque identifier. The initial proposal I had was to use AWS S3-style ETags because it allows for an easy way to emulate S3 using jclouds (like with the S3Proxy project). That's the motivation behind this PR and the prior one related to it: https://github.com/jclouds/jclouds/pull/1251.

demobox: > The initial proposal I had was to use AWS S3-style ETags because it allows for an easy way to emulate S3 using jcloud

Clear, thanks!

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
timuralp: Gotcha! Will remove the extraneous checks.

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
gaul: S3 has emerged as the de facto standard for object storage.  I think the average user would prefer the local blobstore acts more similarly to S3, especially since one of the main consumers of this is S3Proxy.  :-)

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
```diff
@@ -838,10 +839,10 @@ public String completeMultipartUpload(MultipartUpload mpu, List<MultipartPart> p
             throw propagate(ioe);
          }
          streams.add(is);
-         partHashes.append(blobPart.getMetadata().getETag());
+         md5Hasher.putBytes(new BigInteger(blobPart.getMetadata().getETag(), 16).toByteArray());
```
gaul: This is clever but would `BaseEncoding.base16.decode` address this more naturally?

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
```diff
@@ -113,19 +121,49 @@ public void testSetBlobAccess() throws Exception {
...
+      int partSize = 5 * (int) Math.pow(2, 20);
+      try {
+         String name = "blob-name";
+         BlobBuilder blobBuilder = blobStore.blobBuilder(name);
+         Blob blob = blobBuilder.build();
+         MultipartUpload mpu = blobStore.initiateMultipartUpload(container, blob.getMetadata(), new PutOptions());
+
+         byte[] content = new byte[partSize];
+         Arrays.fill(content, (byte) 'b');
+         Payload payload = Payloads.newByteArrayPayload(content);
```
gaul: Would `TestUtils.randomdByteSource(seed)` work as well?

apis/filesystem/src/test/java/org/jclouds/filesystem/integration/FilesystemBlobIntegrationTest.java
```diff
@@ -113,19 +121,49 @@ public void testSetBlobAccess() throws Exception {
...
       assertThat(blob.getMetadata().getETag()).isEqualTo(expectedETag);
    }
 
+   @Test(groups = { "integration", "live" })
+   public void testMultipartUploadMultiplePartsKnownETag() throws Exception {
+      BlobStore blobStore = view.getBlobStore();
+      String container = getContainerName();
+      // Pre-computed ETag returned by AWS S3 for the MPU consisting of two 5MB parts filled with 'b'
+      String expectedETag = "\"84462a16f6a60478d50148808aa609c1-2\"";
+      int partSize = 5 * (int) Math.pow(2, 20);
```
gaul: Easier to read `5 * 1024 * 1024`.

gaul:
Pushed to master as 896e99df097cf20759f89167fb74962854ab19b5.  Thank you for your contribution @timuralp!

PR #1254, opened 2018-11-01
Removed check for prefix/delimiter in clearContainer

reptillicus:
as discussed in this mail thread: https://lists.apache.org/thread.html/e10af185b92545935223cd070a9eabbda2898ba7265a223512c07f3f@%3Cuser.jclouds.apache.org%3E

Also added some tests in BaseContainerIntegrationTest that test clearContainer with prefix etc.

### DIFF
--- a/blobstore/src/main/java/org/jclouds/blobstore/strategy/internal/DeleteAllKeysInList.java
+++ b/blobstore/src/main/java/org/jclouds/blobstore/strategy/internal/DeleteAllKeysInList.java
@@ -380,9 +380,6 @@ public class DeleteAllKeysInList implements ClearListStrategy, ClearContainerStr
 
    public void execute(final String containerName,
          ListContainerOptions listOptions) {
-      if (listOptions.getDelimiter() != null || listOptions.getPrefix() != null) {
-         throw new IllegalArgumentException("Prefix and delimiter support has not yet been added");
-      }
       final AtomicBoolean deleteFailure = new AtomicBoolean();
       int retries = maxErrors;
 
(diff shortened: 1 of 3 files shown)


gaul:
@timuralp you probably added this check; could you look at this pull request?

gaul:
Fixed nits, reworded commit message, and pushed to master as 22ce5484a412bc06ef62995675c07e7a85f66bdf and 2.1.x as 9ede9d3ddc9ac4aa175bf9c364b83a161bc93294.  Thank you for your contribution @reptillicus!  Also hello to my alma mater.

gaul:
We definitely appreciate the help.  The list of open blobstore issues has grown quite large:

https://issues.apache.org/jira/browse/JCLOUDS-1465?jql=project%20%3D%20JCLOUDS%20AND%20component%20%3D%20jclouds-blobstore%20AND%20status%20%3D%20open

Please jump in if something interests you!

gaul:
Unfortunately filesystem has its own logic so I added a check to disable this in 1ae735bb7d8fa4d512e744f79582477efdfd7f26 and explained the problem in a comment.  @reptillicus could you investigate this?  You can run the single test with 

```
mvn integration-test -am -pl :filesystem -Plive -Dtest=FilesystemContainerIntegrationTest#testClearWithOptions -DfailIfNoTests=false
```

reptillicus:
Sure, I'll poke around this afternoon.

reptillicus:
So is the expected behavior of the FileSystem blob store to also remove any
empty folders if a path is cleared and no files exist in there? As in, same
behavior as S3 type blob store.

gaul:
filesystem should act like other blobstores to the most practical extent possible.  It does this correctly today if you have one blob named `a/b` and remove it.  Both the file `a/b` and the enclosing directory are removed.

PR #1264, opened 2018-12-13
[JCLOUDS-1474] [ARM] Enable Standard SKUs for Load Balancers and PublicIPsg

danielestevez:
* Adds SKU field to PublicIPAddress and LoadBalancer
* Adds LiveTests to create these new type of entities
* Modifies existing calls to have default SKU null, which translates to Basic in Azure deployments. Jclouds will keep creating these Basic SKU IPs when deploying a VM

https://issues.apache.org/jira/browse/JCLOUDS-1474


providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/LoadBalancerSKU.java
```diff
@@ -0,0 +1,44 @@
...
+
+   public static enum LoadBalancerSKUName {
+      Basic, Standard, Unrecognized;
+
+      public static LoadBalancerSKUName fromValue(final String text) {
+         return (LoadBalancerSKUName) GetEnumValue.fromValueOrDefault(text, LoadBalancerSKUName.Unrecognized);
+      }
+   }
+
+   @Nullable
```
nacx: I don't think this is nullable. Can the only field in the object be null?

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/LoadBalancerSKU.java
```diff
@@ -0,0 +1,44 @@
...
+import org.jclouds.azurecompute.arm.util.GetEnumValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class LoadBalancerSKU {
+
+   public static enum LoadBalancerSKUName {
```
nacx: Just call it `Name` and reference it as `LoadBalancerSKU.Name`.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/PublicAddressSKU.java
```diff
@@ -0,0 +1,44 @@
...
+import org.jclouds.azurecompute.arm.util.GetEnumValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class PublicAddressSKU {
+
+   public static enum PublicIPAddressSkuName {
```
nacx: Same naming pattern as above.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/PublicAddressSKU.java
```diff
@@ -0,0 +1,44 @@
...
+      Basic, Standard, Unrecognized;
+
+      public static PublicIPAddressSkuName fromValue(final String text) {
+         return (PublicIPAddressSkuName) GetEnumValue.fromValueOrDefault(text, PublicIPAddressSkuName.Unrecognized);
+      }
+   }
+
+   @Nullable
```
nacx: Probably this is not nullable.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/LoadBalancerApi.java
```diff
@@ -68,7 +69,7 @@
    @MapBinder(BindToJsonPayload.class)
    LoadBalancer createOrUpdate(@PathParam("loadbalancername") String lbName,
          @PayloadParam("location") String location, @Nullable @PayloadParam("tags") Map<String, String> tags,
-         @PayloadParam("properties") LoadBalancerProperties properties);
+         @PayloadParam("properties") LoadBalancerProperties properties, @Nullable @PayloadParam("sku") LoadBalancerSKU sku);
```
nacx: Leave the properties parameter at the end. That is more aligned with the style of all other APIs and methods.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/PublicIPAddressApi.java
```diff
@@ -66,9 +67,9 @@
    @MapBinder(BindToJsonPayload.class)
    @PUT
    PublicIPAddress createOrUpdate(@PathParam("publicipaddressname") String publicipaddressname,
-                                                 @PayloadParam("location") String location,
-                                                 @Nullable @PayloadParam("tags") Map<String, String> tags,
-                                                 @PayloadParam("properties") PublicIPAddressProperties properties);
+         @PayloadParam("location") String location, @Nullable @PayloadParam("tags") Map<String, String> tags,
+         @PayloadParam("properties") PublicIPAddressProperties properties,
+         @Nullable @PayloadParam("sku") PublicAddressSKU sku);
```
nacx: Same comment, keep the properties object at the end.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/LoadBalancerSKU.java
```diff
@@ -0,0 +1,44 @@
+/*
```
nacx: You probably can embed this type in the LoadBalancer type. It only makes sense inside the LB and naming is more meaningful as you just do `LoadBalancer.SKU`. Up to you.

danielestevez: Same concern as before keeping the original MS names, but embedding it seems a good idea.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/PublicIPAddressApi.java
danielestevez: sure

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/LoadBalancerSKU.java
danielestevez: You're right, the whole `LoadBalancerSKU` can be null and will default to `Basic` but this field is mandatory. Will change

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/LoadBalancerSKU.java
danielestevez: I like `Name` better but i was trying to respect the original MS names for this entity since we are mapping their domain objects. Does it make sense?
https://docs.microsoft.com/en-us/rest/api/load-balancer/loadbalancers/get#loadbalancerskuname

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/PublicAddressSKU.java
danielestevez: same concern!

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/PublicAddressSKU.java
danielestevez: Same as before. Will change

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/LoadBalancerApi.java
danielestevez: Fine!

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/LoadBalancerSKU.java
danielestevez: Besides the 'respect original MS domain names', keeping the name makes it easier to find the entity documentation.
But as i said, we could keep our own simpler and more readable field names. Just not sure what's better

danielestevez:
Thanks for the review @nacx !
Finally i added all SKU entities as inner classes so it's easier to reference them in code.
Also both SKUs have a `SKUName` field similar to what we do at `AvailabilitySet`

I also noticed there is a generic `SKU` entity that should be under Vault and a `VirtualMachineScaleSKU` with the same problem as in this PR's original commit

danielestevez:
rebuild please

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/loadbalancer/LoadBalancer.java
```diff
@@ -33,11 +34,13 @@
...
    @Nullable public abstract LoadBalancerProperties properties();
-   @Nullable public abstract LoadBalancerSKU sku();
+
+   @Nullable
+   public abstract SKU sku();
 
    @SerializedNames({ "id", "name", "location", "etag", "tags", "properties", "sku"})
    public static LoadBalancer create(String id, final String name, final String location, final String etag,
-         final Map<String, String> tags, final LoadBalancerProperties properties, final LoadBalancerSKU sku) {
+         final Map<String, String> tags, final LoadBalancerProperties properties, final SKU sku) {
```
nacx: Keep properties as the last parameter, for consistency.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/publicipaddress/PublicIPAddress.java
```diff
@@ -34,11 +35,13 @@
...
    public abstract PublicIPAddressProperties properties();
-   @Nullable  public abstract PublicAddressSKU sku();
+
+   @Nullable
+   public abstract SKU sku();
 
    @SerializedNames({ "name", "id", "etag", "location", "tags", "properties", "sku" })
    public static PublicIPAddress create(String name, String id, String etag, String location, Map<String, String> tags,
-         PublicIPAddressProperties properties, PublicAddressSKU sku) {
+         PublicIPAddressProperties properties, SKU sku) {
```
nacx: Keep properties at the end.

danielestevez:
Merged to [2.1.x](http://git-wip-us.apache.org/repos/asf/jclouds/commit/3d308560) and [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/d621edd7)

PR #1266, opened 2019-01-04
JCLOUDS-1465: Added test for large multipart uploads

reptillicus:
This is going to fail for now. 

https://issues.apache.org/jira/browse/JCLOUDS-1465?jql=project%20%3D%20JCLOUDS%20AND%20component%20%3D%20jclouds-blobstore%20AND%20status%20%3D%20open

### DIFF
--- a/apis/s3/src/test/java/org/jclouds/s3/blobstore/integration/S3BlobIntegrationLiveTest.java
+++ b/apis/s3/src/test/java/org/jclouds/s3/blobstore/integration/S3BlobIntegrationLiveTest.java
@@ -74,4 +74,9 @@ public class S3BlobIntegrationLiveTest extends BaseBlobIntegrationTest {
          throw new SkipException("S3 does not allow setting Glacier storage class on putBlob", hre);
       }
    }
+
+   @Override
+   public void testPutHugeObject() throws Exception {
+      super.testPutHugeObject();
+   }
 }
--- a/blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
+++ b/blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
@@ -35,6 +35,7 @@ import static org.testng.Assert.fail;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.RandomAccessFile;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
 import java.util.Date;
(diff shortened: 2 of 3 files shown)


apis/s3/src/test/java/org/jclouds/s3/blobstore/integration/S3BlobIntegrationLiveTest.java
```diff
@@ -74,4 +74,9 @@ public void testPutBlobTierArchiveMultipart() throws Exception {
          throw new SkipException("S3 does not allow setting Glacier storage class on putBlob", hre);
       }
    }
+
+   @Override
+   public void testPutHugeObject() throws Exception {
```
gaul: Does this method have a purpose?  Subclasses will inherit this functionality without overriding it.

blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
```diff
@@ -685,6 +686,32 @@ public void testPutMultipartInputStream() throws Exception {
...
+      //Create a 25GB file, 
+      long length = 25L * 1024L  * 1024L * 1024L;
+      BlobStore blobStore = view.getBlobStore();
+      String containerName = getContainerName();
+      try {
+         File tempFile = File.createTempFile("prefix", "suffix");
+         tempFile.deleteOnExit();
+         RandomAccessFile raf = new RandomAccessFile(tempFile, "rw");
+         raf.setLength(length);
+         ByteSource byteSource = Files.asByteSource(tempFile);
```
gaul: Prefer `TestUtils.randomByteSource`.

blobstore/src/test/java/org/jclouds/blobstore/integration/internal/BaseBlobIntegrationTest.java
```diff
@@ -685,6 +686,32 @@ public void testPutMultipartInputStream() throws Exception {
       testPut(payload, null, new ByteSourcePayload(byteSource), length, new PutOptions().multipart(true));
    }
 
+   @Test(groups = { "integration", "live" })
+   public void testPutHugeObject() throws Exception {
```
gaul: We cannot add this to the general integration test group since it will explode run times.  It looks like we have a "performance" group for this; could you use that tag instead?

gaul (changes requested):

reptillicus:
How did this pass??? Locally, this fails every time for me.

reptillicus:
ahhh, ok. I'm at a loss on how to debug this one, putting in breakpoints with a debugger it passes. If anyone has any suggestions let me know :) 

Also, when do the integration tests get run?

gaul:
Sorry I do not understand the status of the referenced issue or this pull request.  I commented in the issue but this symptom could be a socket error.  To run this single test against AWS one would run:

```
mvn integration-test -pl :aws-s3 -Plive \
        -Dtest.aws-s3.identity="${JCLOUDS_IDENTITY}" \
        -Dtest.aws-s3.credential="${JCLOUDS_CREDENTIAL}"
        -Dtest=AWSS3BlobIntegrationLiveTest#testPutHugeObject
```

Developers manually run integration tests, including before releases.  We lack an automated mechanism to do this on a per-pull request basis.

reptillicus:
This is mainly a repeatable test case that fails (for me against AWS and a local Minio instance). I'm not sure where to put this (it is a bug and blocker for some people) but the comments in the JIRA issue mentioned making a test case for it.

gaul:
I have done some testing with this and large uploads succeed for me.  I am still hacking around on a fix for S3Proxy's large object emulation middleware.

gaul:

PR #1267, opened 2019-01-15
Adding support for AWS S3 get_objects V2 api call.

jumenzel:
- Adding support for AWS S3 get_objects V2 api call.


apis/s3/src/main/java/org/jclouds/s3/options/ListBucketOptions.java
```diff
@@ -37,6 +37,15 @@
 public class ListBucketOptions extends BaseHttpRequestOptions implements Cloneable {
    public static final ListBucketOptions NONE = new ListBucketOptions();
 
+   public ListBucketOptions withListType(String listType) {
```
gaul: Should this be `int` to match other getters and setters?

apis/s3/src/main/java/org/jclouds/s3/options/ListBucketOptions.java
```diff
@@ -37,6 +37,15 @@
...
+      queryParameters.put("list-type", checkNotNull(listType, "listType"));
+      return this;
+   }
+
+   public String getListType() {
```
gaul: Similarly, should this be `Integer` (matching `getMaxResults`)?

apis/s3/src/test/java/org/jclouds/s3/blobstore/functions/BucketToContainerListOptions.java
```diff
@@ -28,13 +28,24 @@
    public ListContainerOptions apply(ListBucketOptions[] optionsList) {
       ListContainerOptions options = new ListContainerOptions();
       if (optionsList.length != 0) {
+         int listType = 1;
+         if (optionsList[0].getListType() != null && optionsList[0].getListType().equals("2")) {
```
gaul: Prefer `Strings.nullToEmpty(optionsList[0].getListType()).equals("2")`.

apis/s3/src/main/java/org/jclouds/s3/options/ListBucketOptions.java
jumenzel: It's a bit strange, the only value that's allowed is 2. That's how list objects v2 is triggered. 
With v1 you don't set list-type at all. So there's no list type v1. 
I can just make it an int and only allow value 2.

apis/s3/src/main/java/org/jclouds/s3/options/ListBucketOptions.java
jumenzel: Again here the value returned would be either 2 or null.

jumenzel:
The difference that I found useful is the use of the ContinuationToken and NextContinuationToken for listing files in blocks. I think v2 also has a option to get some metadata of each object listed.

gaul:
@jumenzel does this pull request need anything to move forward?

gaul:
> I also noticed that without any of my changes building master fails:
> `[INFO] jclouds filesystem core ............................ FAILURE [ 19.118 s]`
> The 2.1.x branch builds just fine.

`mvn test -pl :filesystem` succeeds for me on Fedora 9 with Java 8.  Which environment do you build on?

jumenzel:
Just realized it's the the integration test that is failing for filesystem.

blobstore/src/main/java/org/jclouds/blobstore/options/ListContainerOptions.java
```diff
@@ -44,18 +44,20 @@
...
    private boolean recursive;
    private boolean detailed;
+   private int listType = 1;
 
    public ListContainerOptions() {
    }
 
    ListContainerOptions(Integer maxKeys, String marker, String dir, boolean recursive,
-            boolean detailed, String prefix, String delimiter) {
+            boolean detailed, String prefix, String delimiter, int listType) {
```
gaul: Sorry I missed this earlier, but why does the portable abstraction need to know the S3 list type?  This looks like a layering violation.

gaul:
> Just realized it's the integration test that is failing for filesystem.

Both of these succeed for me on Fedora 29:

```
mvn install -pl :filesystem
mvn integration-test -pl :filesystem
```

Probably better to file an issue with your symptoms since it is not related to this pull request.

gaul:
@jumenzel do you not plan to continue this pull request?  This feature is useful to return the owner information during list objects.

jumenzel:
I'll make another attempt later this week.

PR #1268, opened 2019-01-25
JCLOUDS-1371: JCLOUDS-1488: list optimize prefix

gaul:
Previously `getBlobKeysInsideContainer` returned all keys and filtered
in `LocalBlobStore`.  Now `getBlobKeysInsideContainer` filters via prefix
which can dramatically decrease the number of keys returned,
especially for the filesystem provider.  Further optimizations are
possible for delimiter.

### DIFF
--- a/apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
+++ b/apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
@@ -342,7 +342,7 @@ public class FilesystemStorageStrategyImpl implements LocalStorageStrategy {
     * @throws IOException
     */
    @Override
-   public Iterable<String> getBlobKeysInsideContainer(String container) throws IOException {
+   public Iterable<String> getBlobKeysInsideContainer(String container, String prefix) throws IOException {
       filesystemContainerNameValidator.validate(container);
       // check if container exists
       // TODO maybe an error is more appropriate
@@ -353,7 +353,7 @@ public class FilesystemStorageStrategyImpl implements LocalStorageStrategy {
 
       File containerFile = openFolder(container);
       final int containerPathLength = containerFile.getAbsolutePath().length() + 1;
-      populateBlobKeysInContainer(containerFile, blobNames, new Function<String, String>() {
+      populateBlobKeysInContainer(containerFile, blobNames, prefix, new Function<String, String>() {
          @Override
          public String apply(String string) {
             return denormalize(string.substring(containerPathLength));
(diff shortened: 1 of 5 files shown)


gaul:
@timuralp could you review this?

blobstore/src/main/java/org/jclouds/blobstore/TransientStorageStrategy.java
```diff
@@ -148,8 +149,13 @@ public boolean blobExists(final String containerName, final String blobName) {
...
 
    @Override
-   public Iterable<String> getBlobKeysInsideContainer(final String containerName) {
-      return containerToBlobs.get(containerName).keySet();
+   public Iterable<String> getBlobKeysInsideContainer(final String containerName, String prefix) {
+      ConcurrentSkipListMap<String, Blob> blobs = containerToBlobs.get(containerName);
+      if (prefix == null) {
+         return blobs.keySet();
+      }
+      String lastPrefix = prefix + (char) 65535;  // TODO: better sentinel?
```
gaul: This is hacky but I don't think is has practical drawbacks?

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -964,17 +964,27 @@ private File openFolder(String folderName) throws IOException {
...
-            blobNames.add( function.apply(child.getAbsolutePath()) );
+            if (prefix != null && !fullPath.startsWith(prefix)) {
+               continue;
+            }
+            blobNames.add(fullPath);
          } else if (child.isDirectory()) {
-            blobNames.add(function.apply(child.getAbsolutePath()) + File.separator); // TODO: undo if failures
-            populateBlobKeysInContainer(child, blobNames, function);
+            // Consider a prefix /a/b/c but we have only descended to path /a.
+            // We need to match the path against the prefix to continue
```
timuralp: I was confused by the additional check here, but after removing it and running the tests, it makes sense: the directories will not include the trailing "/", so if the prefix is "dir/", then it wouldn't match the directory "dir" (excluding it from the list and not listing its contents). I think this comment helps, but maybe it could more explicitly call out this specific issues. As far as I can tell, this only affects prefix listings that end with the file separator.

gaul: Addressed suggestion via a more narrow filter.

blobstore/src/main/java/org/jclouds/blobstore/TransientStorageStrategy.java
demobox: [minor] Perhaps create a constant `SENTINEL` or so just to make it clear that that's what the last character is supposed to be?

That might be confusing, especially if the TODO is removed in future?

PR #1269, opened 2019-01-29
Lazily open InputStream during complete MPU

gaul:
Previously the filesystem provider could exhaust file descriptors by
eagerly opening up to 10,000 parts.  This partially undoes
JCLOUDS-1367.


gaul:
@timuralp could you review this?

gaul:
> Is there anything about this change that would merit a test? I'm assuming "does not open tons of file descriptors" is not something that's easy/feasible to test for..?

Indeed this commit was ineffective due to JCLOUDS-1367.  I added a test and fixed the other issue.

PR #1271, opened 2019-02-02
Remove JavaScript from footer

gaul:
Addresses an error while building Javadoc.

### DIFF
--- a/project/pom.xml
+++ b/project/pom.xml
@@ -1120,19 +1120,6 @@
             <links>
               <link>http://download.oracle.com/javase/6/docs/api/</link>
             </links>
-            <footer><![CDATA[
-            <!-- Google Analytics -->
-            <script type='text/javascript'>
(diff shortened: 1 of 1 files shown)


gaul:
Blocks 2.1.2-rc1.

project/pom.xml
```diff
@@ -1120,19 +1120,6 @@
             <links>
               <link>http://download.oracle.com/javase/6/docs/api/</link>
             </links>
-            <footer><![CDATA[
-            <!-- Google Analytics -->
```
demobox: Out of curiosity: do we know whether there was ever a GA account associated with this?

nacx: If we have it we never used it. I think Everett added this?

gaul: Looks like Adrian in 0e07b7236fb594146bb8c3cd485004716f2897c0.

PR #1276, opened 2019-07-18
Ignore null StorageMetadata returned from blob store

enchobelezirev:
This is needed because of this issue: https://issues.apache.org/jira/browse/JCLOUDS-1504
The null check will ignore the BlobMetadata which is not found in the blob store and will not be added to the result from the query.

JIRA issue: https://issues.apache.org/jira/browse/JCLOUDS-1504

### DIFF
--- a/core/src/main/java/org/jclouds/concurrent/FutureIterables.java
+++ b/core/src/main/java/org/jclouds/concurrent/FutureIterables.java
@@ -80,7 +80,9 @@ public class FutureIterables {
 
          for (F from : fromIterable) {
             ListenableFuture<? extends T> to = function.apply(from);
-            responses.put(from, to);
+            if (to != null) {
+                responses.put(from, to);
+            }
          }
          try {
             exceptions = awaitCompletion(responses, exec, maxTime, logger, logPrefix);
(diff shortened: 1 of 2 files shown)


enchobelezirev:
I agree it is not the best solution ever but for your worries, if the metadata for some blobs could not be taken for some reason(failure, exception, etc.) should there be an exception risen and the processing to stop? For me it is not quite ok to return null if something goes wrong :)

enchobelezirev:
Moreover, I have added logs in the JIRA issue. In the logs, you could check that when some blob metadata is not found(the response code is 404 Not Found), it is being included in the list of blobs(which now includes null objects) returned to the user. For me that is not OK, so that I decided to put a null check just for the retrieval of the blobMetadata which is returning the nulls. 

If I understand the behaviour of the s3 client correctly, when something is not found in the blob store, it returns null objects. I am saying this, because all the files which are related to the null metadatas, are being scheduled for deletion in the blob store.

enchobelezirev:
Something more, could you tell me where I can find the exact implementation of the aws-s3 client which implements the S3Client interface? I want to check whether the issue is in the client.

enchobelezirev:
Hello @gaul,

Any updates here? The issue is hurting our code base.
Let's continue the discussion!

gaul:
Sorry this approach will never merge.  As stated in my previous comment, we need to do a root cause analysis to determine where null S3 blobs come from instead of changing the behavior for all possible callers across all possible providers.  I will share more in [JCLOUDS-1504](https://issues.apache.org/jira/browse/JCLOUDS-1504).

PR #1278, opened 2019-08-08
JCLOUDS-1509: read AWS response data with the UTF-8 charset explicitly

roded:
AWS response data is encoded in UTF-8. Creating a String from said data
using the JVM's default charset results in incorrect encoding on
environments in which the JVM's default charset is not UTF-8.

https://issues.apache.org/jira/browse/JCLOUDS-1509


roded:
Ideally, the response should be parsed as an XML whose encoding is derived from its header. I'm not sure if that's out of scope though, since it seems that currently this code is AWS specific and UTF-8 can be expected.

gaul:
Pushed to master as a248697c0668f8f1c442b04e3833d876b125dc2c and 2.1.x as 95c4443022f9bfaa87952d5ac59befe0b89d755e.  Thank you for your contribution @roded!  Please use the apache/jclouds repository for future pull requests.

roded:
Thanks @gual. Will do.

PR #1279, opened 2019-09-16
Add azure disk SKU

pimuzzo:
Fix https://issues.apache.org/jira/browse/JCLOUDS-1514


pimuzzo:
opened under: https://github.com/apache/jclouds/pull/45

PR #1280, opened 2019-11-27
JCLOUDS-1529: Do not export org.jclouds.json.gson.internal because it causes a

lemmy:
use constraint violation with OSGi.

https://issues.apache.org/jira/browse/JCLOUDS-1529


demobox:
Closing this and looking forward to picking this up on apache/jclouds instead (see also #1281)

PR #1281, opened 2019-11-27
JCLOUDS-1529: Do not export org.jclouds.json.gson.internal because it causes a

lemmy:
use constraint violation with OSGi.

https://issues.apache.org/jira/browse/JCLOUDS-1529

Now for master branch.


nacx:
Thanks!
Mind opening the PR against https://github.com/apache/jclouds instead? This repo is deprecated and we're only keeping it in sync to help people, but will eventually die. All CI builds and tests are not running here, only in the Apache one, so I'd appreciate if you could do that.

Also, just opening against master is fine. I'll cherry-pick the changes to the bugfix branch.

demobox:
Closing this and looking forward to picking this up on apache/jclouds instead

PR #1282, opened 2020-01-23
JCLOUDS-1538: fix typo in rfc1123SimpleDateFormat

ispringer:


nacx:
Thanks! Mind opening it against the apache repo? https://github.com/apache/jclouds

This is a deprecated read-only mirror as we moved to the Apache org. There we run the proper CI and PR tests, etc.

ispringer:
Done.

PR #1283, opened 2020-02-14
JCLOUDS-1533 - Fix upload with SAS token when blob name contains slash

majaseremet:


providers/azureblob/src/main/java/org/jclouds/azure/storage/filters/SharedKeyLiteAuthentication.java
```diff
@@ -281,4 +284,19 @@ void appendUriPath(HttpRequest request, StringBuilder toSign) {
       }
    }
 
+   private String join(String[] parts, String delimiter) {
```
nacx: Can you use the Guava `Joiner` instead?

majaseremet: Sure, sorry, didn't notice guava is already there.

majaseremet:
> Thanks for the fix!
> This repository is deprecated, though. Could you update your fork and open the PR against https://github.com/apache/jclouds ? The CI checks, etc are migrated there too.
> Thanks!

Thanks, I will close this and I opened: https://github.com/apache/jclouds/pull/61.

PR #1284, opened 2020-03-01
add comment to notify user that it is not secure

YYTVicky:


demobox:
@YYTVicky Could you kindly move this to the apache/jclouds repo, as this one is no longer active.

Also, could you add a few more details describing what the security vulnerability is?

